PR #5, opened 2021-09-16
Fix ambiguous log use statement

FrozenDroid:
When using this crate with the latest nightly, I was getting an error about the ambiguous `log` uses. It didn't know whether to refer to the crate's `log` module, or the `log` dependency.
This PR fixes that.


ivmarkov:
Thanks!

PR #11, opened 2021-10-26
feat: http server error handler

ForsakenHarmony:
Necessary for the captive portal

### DIFF
--- a/src/httpd.rs
+++ b/src/httpd.rs
@@ -13,7 +13,7 @@ use embedded_svc::httpd::*;
 
 use esp_idf_sys::c_types::*;
 use esp_idf_sys::esp;
-use esp_idf_sys::esp_nofail;
+use esp_idf_sys::*;
 
 use crate::private::cstr::*;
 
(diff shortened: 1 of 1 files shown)


src/httpd.rs
```diff
@@ -173,20 +173,108 @@ impl Default for Configuration {
...
+            HttpdError::UriTooLong => httpd_err_code_t_HTTPD_414_URI_TOO_LONG,
+            HttpdError::ReqHdrFieldsTooLarge => httpd_err_code_t_HTTPD_431_REQ_HDR_FIELDS_TOO_LARGE,
+        }
+    }
+}
+
+pub struct ErrorHandler {
+    error: HttpdError,
+    // handler: Box<dyn Fn(Request, HttpdError) -> Result<Response>>,
+    handler: unsafe extern "C" fn(req: *mut httpd_req_t, error: httpd_err_code_t) -> esp_err_t,
```
ivmarkov, line 236: The error handler is unsafe. Is there a good reason for that?

src/httpd.rs
```diff
@@ -173,20 +173,108 @@ impl Default for Configuration {
...
+            HttpdError::ReqTimeout => httpd_err_code_t_HTTPD_408_REQ_TIMEOUT,
+            HttpdError::LengthRequired => httpd_err_code_t_HTTPD_411_LENGTH_REQUIRED,
+            HttpdError::UriTooLong => httpd_err_code_t_HTTPD_414_URI_TOO_LONG,
+            HttpdError::ReqHdrFieldsTooLarge => httpd_err_code_t_HTTPD_431_REQ_HDR_FIELDS_TOO_LARGE,
+        }
+    }
+}
+
+pub struct ErrorHandler {
+    error: HttpdError,
```
ivmarkov, line 234: How am I supposed to use this `pub` struct outside of `esp-idf-svc` given that its members are private?

src/httpd.rs
```diff
@@ -273,6 +367,17 @@ impl Server {
         Ok(())
     }
 
+    fn register_error_handler(&mut self, handler: ErrorHandler) -> Result<()> {
```
ivmarkov, line 370: Is de-registration of these necessary, on server stop?

ivmarkov (changes requested):

src/httpd.rs
ForsakenHarmony: probably, I'll check

src/httpd.rs
ForsakenHarmony: Semi good reason, the error handler doesn't get the `user_ctx` like the normal handler does, so this has to be a unique function that gets called

> When invoked, the validity of `uri`, `method`, `content_len` and `user_ctx` fields of the httpd_req_t parameter is not guaranteed as the HTTP request may be partially received/parsed.

It should be possible to work around it with a global Mutex or something else

src/httpd.rs
ForsakenHarmony: Should most likely not be `pub`

ivmarkov:
Closing as the PR is too old and is not mergeable. If support for the custom error handling is still desired, it should be implemented with a new PR.

PR #13, opened 2021-10-26
feat: waiter and some small wrappers

ForsakenHarmony:
The Waiter with a condvar can probably be used in the wifi module, saw some comments about it there

Didn't know where else to put the smaller changes

### DIFF
--- a/src/misc.rs
+++ b/src/misc.rs
@@ -1,0 +1,11 @@
+use esp_idf_sys::*;
+
+pub fn get_default_efuse_mac() -> Result<[u8; 6], EspError> {
+    let mut mac = [0; 6];
+    unsafe { esp!(esp_efuse_mac_get_default(mac.as_mut_ptr()))? }
+    Ok(mac)
+}
+
+pub fn restart() {
+    unsafe { esp_restart() };
+}
--- a/src/private/wait.rs
+++ b/src/private/wait.rs
@@ -1,0 +1,97 @@
+use core::time::Duration;
+#[cfg(feature = "std")]
+use std::sync::{Condvar, Mutex};
+
+use embedded_svc::mutex::Mutex as _;
+#[cfg(not(feature = "std"))]
+use esp_idf_sys::*;
+
+#[cfg(not(feature = "std"))]
+use super::time::micros_since_boot;
+
+pub struct Waiter {
+    #[cfg(feature = "std")]
+    cvar: Condvar,
+    #[cfg(feature = "std")]
+    running: Mutex<bool>,
+    #[cfg(not(feature = "std"))]
+    running: EspMutex<bool>,
+}
+
+impl Waiter {
+    pub fn new() -> Self {
+        Waiter {
+            #[cfg(feature = "std")]
+            cvar: Condvar::new(),
+            #[cfg(feature = "std")]
+            running: Mutex::new(false),
+            #[cfg(not(feature = "std"))]
+            running: EspMutex::new(false),
+        }
+    }
+
+    pub fn start(&self) {
+        self.running.with_lock(|running| *running = true);
+    }
+
+    #[cfg(feature = "std")]
+    pub fn wait(&self) {
+        if !self.running.with_lock(|running| *running) {
+            return;
+        }
+
+        let _running = self
+            .cvar
+            .wait_while(self.running.lock().unwrap(), |running| *running)
+            .unwrap();
+    }
+
+    #[cfg(not(feature = "std"))]
+    pub fn wait(&self) {
+        while self.running.with_lock(|running| *running) {
+            unsafe { vTaskDelay(500) };
+        }
+    }
+
+    /// return = !timeout (= success)
+    #[cfg(feature = "std")]
+    pub fn wait_timeout(&self, dur: Duration) -> bool {
+        if !self.running.with_lock(|running| *running) {
+            return true;
+        }
+
+        let (_running, res) = self
+            .cvar
+            .wait_timeout_while(self.running.lock().unwrap(), dur, |running| *running)
+            .unwrap();
+
+        return !res.timed_out();
+    }
+
+    /// return = !timeout (= success)
+    #[cfg(not(feature = "std"))]
+    pub fn wait_timeout(&self, dur: Duration) {
+        let now = micros_since_boot();
+        let end = now + dur.as_micros();
+
+        while self.running.with_lock(|running| *running) {
+            if micros_since_boot() > end {
+                return false;
+            }
+            unsafe { vTaskDelay(500) };
+        }
+
+        return true;
+    }
+
+    #[cfg(feature = "std")]
+    pub fn notify(&self) {
+        *self.running.lock().unwrap() = false;
+        self.cvar.notify_all();
+    }
+
+    #[cfg(not(feature = "std"))]
+    pub fn notify(&self) {
+        self.running.with_lock(|running| *running = false);
+    }
+}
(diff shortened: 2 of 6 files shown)


src/misc.rs
```diff
@@ -0,0 +1,11 @@
+use esp_idf_sys::*;
+
+pub fn get_default_efuse_mac() -> Result<[u8; 6], EspError> {
```
ivmarkov, line 3: I don't feel very comfortable wrapping this function, unless there are multiple - as in many - many use-cases where we need it.
If you are just going to use it in  e.g. the smtp implementation and a couple of others - then I rather say - we pass on it.

src/time.rs
```diff
@@ -0,0 +1,7 @@
+use core::convert::TryInto;
```
ivmarkov, line 1: Ditto. Let's be careful when extending the API surface with all sorts of utilities, until we have good use cases for those. And ideally - traits in `embedded-svc` or `embedded-hal`.

src/misc.rs
```diff
@@ -0,0 +1,11 @@
...
+    let mut mac = [0; 6];
+    unsafe { esp!(esp_efuse_mac_get_default(mac.as_mut_ptr()))? }
+    Ok(mac)
+}
+
+pub fn restart() {
```
ivmarkov, line 9: I think this is overall useful and can stay. Ideally, we should have a trait for it in `embedded-svc`, but this can come later (at the expense of API breakage).

ivmarkov (changes requested):

src/misc.rs
ForsakenHarmony: This, like the nvs api is mostly to have safe wrappers around esp idf functions, so I don't have to have unsafe code in my project

I suppose if this is mostly meant as an implementation of `embedded-svc` there could be another crate just for safe wrappers (which this crate could make use of)

ivmarkov: > This, like the nvs api is mostly to have safe wrappers around esp idf functions, so I don't have to have unsafe code in my project
> 
> I suppose if this is mostly meant as an implementation of `embedded-svc` there could be another crate just for safe wrappers (which this crate could make use of)

No it is not necessarily just an implementation of `embedded-svc`. Take a look at `EspNetif`. This is currently not having a corresponding `embedded-svc` trait, as I'm not sure that the notion of a network interface can be easily exposed everywhere (as in on all platforms) as a top level notion.

But `esp-idf-svc` IS about exposing *self-contained* pieces of functionality, rather than individual functions on "as needed" basis, because the moment you expose an individual function "because I just need it for this one use case", you have to maintain and evolve it, which is not easy and results in breaking changes if you don't think upfront how this is going to evolve. If I have to give you example precisely about the `get_default_efuse_mac` function you've exposed: this is just one API call from a whole set of APIs: `esp_mac.h`. So, how do we think about the whole set of APIs which are out there in `esp_mac.h`?
* Are we expecting to expose all of these over time? If yes, would all of these be in `misc`, or do we need to dedicate a separate namespace (a.k.a. a Rust module) for them, rather than keeping them in `misc`? if so, how about we dedicate a module for all of these, from the get-go?
* if we don't expect that this whole API would be necessary, and we'll only need `get_default_efuse_mac` - for whatever reasons - at one place, perhaps it is not worth it exposing it as a public, safe API in the first place, as it won't be having that many usages anyway?

ivmarkov (changes requested):

src/private/wait.rs
```diff
@@ -0,0 +1,97 @@
...
+    #[cfg(not(feature = "std"))]
+    pub fn wait(&self) {
+        while self.running.with_lock(|running| *running) {
+            unsafe { vTaskDelay(500) };
+        }
+    }
+
+    /// return = !timeout (= success)
+    #[cfg(feature = "std")]
+    pub fn wait_timeout(&self, dur: Duration) -> bool {
```
ivmarkov, line 58: Ditto.

src/private/wait.rs
```diff
@@ -0,0 +1,97 @@
...
+            .cvar
+            .wait_timeout_while(self.running.lock().unwrap(), dur, |running| *running)
+            .unwrap();
+
+        return !res.timed_out();
+    }
+
+    /// return = !timeout (= success)
+    #[cfg(not(feature = "std"))]
+    pub fn wait_timeout(&self, dur: Duration) {
```
ivmarkov, line 73: Ditto.

src/private/wait.rs
```diff
@@ -0,0 +1,97 @@
...
+                return false;
+            }
+            unsafe { vTaskDelay(500) };
+        }
+
+        return true;
+    }
+
+    #[cfg(feature = "std")]
+    pub fn notify(&self) {
```
ivmarkov, line 88: Needs to take a state `T` that is pushed to the mutex before notifying. We also need to probably rename this to `set` or something.

src/private/wait.rs
```diff
@@ -0,0 +1,97 @@
...
+        *self.running.lock().unwrap() = false;
+        self.cvar.notify_all();
+    }
+
+    #[cfg(not(feature = "std"))]
+    pub fn notify(&self) {
```
ivmarkov, line 94: Ditto.

src/private/wait.rs
```diff
@@ -0,0 +1,97 @@
...
+use esp_idf_sys::*;
+
+#[cfg(not(feature = "std"))]
+use super::time::micros_since_boot;
+
+pub struct Waiter {
+    #[cfg(feature = "std")]
+    cvar: Condvar,
+    #[cfg(feature = "std")]
+    running: Mutex<bool>,
```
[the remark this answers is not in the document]
ivmarkov: If you are still willing to drive e.g. the `sntp` implementation which needs (?) a "waiter", here's one that (ideally) should work for sntp & ping, but also for esp, wifi:
https://github.com/esp-rs/esp-idf-svc/compare/waitable

Not really tested but at least compiles and passed fmt and clippy checks. And used in all of ping, eth and wifi. Hopefully that implementation makes clearer some of the changes I'm asking for here.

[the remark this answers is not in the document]
ForsakenHarmony: > If you are still willing to drive e.g. the sntp implementation which needs (?) a "waiter", here's one that (ideally) should work for sntp & ping, but also for eth & wifi:

Oh, you wrote your own now?
Should I close this then?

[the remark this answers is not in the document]
ivmarkov: Up to you. You can also take bits and pieces of mine and update yours with these. The reason I took the effort to actually write something is because you might feel that I'm nitpicking on everything which is definitely not the impression I want to leave - so I thought - I can at least take the effort to justify -in code - including how it will be used in eth and wifi code - why the waiter needs to be generified, you know. :)

src/misc.rs
ForsakenHarmony: Hmm, yes I understand your concerns

The reason I exposed it was to generate a unique Wi-Fi name (which seems to be a common use case)
But exposing it like this is probably too low level

ivmarkov: What do you mean by `unique Wi-Fi name`? A MAC (as in OSI layer 2 ethernet MAC)?
Huh. I was expecting that this is somehow handled by the ESP WiFi driver itself. Now, if it is not, we should fix the `EspWifi` code itself to derive a useful MAC or something. Ditto for `EspEth`.

But, if you don't need this MAC in the captive DNS portal, or in the SNTP implementation (and I'm curious why would you need a layer 2 thing in  layer 3 protocols anyway), then I say let's open a new issue describing the problem, and then fix it in the `EspWifi` and `EspEth` driver wrappers themselves.

ivmarkov:
https://github.com/esp-rs/esp-idf-svc/commit/49d66eda2c6331023e979f7aeab6453e1dfd6953

PR #14, opened 2021-10-26
feat: sntp 

ForsakenHarmony:
Will get the time from a NTP server

```rs
let sntp = EspSntp::new(SntpConf::default())?;
sntp.wait_for_sync(Duration::from_secs(30));
```

After you can use the usual time APIs and get the real-world time (i.e. `std::time::SystemTime::now()`)

Depends on #13

### DIFF
--- a/src/sntp.rs
+++ b/src/sntp.rs
@@ -1,0 +1,191 @@
+use core::{time::Duration};
+
+use anyhow::*;
+use log::*;
+use mutex_trait::Mutex;
+
+use esp_idf_sys::*;
+
+use crate::private::{cstr::CString, wait::Waiter};
+
+const SNTP_SERVER_NUM: usize = SNTP_MAX_SERVERS as usize;
+
+pub enum OperatingMode {
+    Poll,
+    ListenOnly,
+}
+
+impl From<u8_t> for OperatingMode {
+    fn from(from: u8_t) -> Self {
+        match from as u32 {
+            SNTP_OPMODE_POLL => OperatingMode::Poll,
+            SNTP_OPMODE_LISTENONLY => OperatingMode::ListenOnly,
+            _ => unreachable!(),
+        }
+    }
+}
+
+impl From<OperatingMode> for u8_t {
+    fn from(from: OperatingMode) -> Self {
+        match from {
+            OperatingMode::Poll => SNTP_OPMODE_POLL as u8_t,
+            OperatingMode::ListenOnly => SNTP_OPMODE_LISTENONLY as u8_t,
+        }
+    }
+}
+
+pub enum SyncMode {
+    Smooth,
+    Immediate,
+}
+
+impl From<sntp_sync_mode_t> for SyncMode {
+    #[allow(non_upper_case_globals)]
+    fn from(from: sntp_sync_mode_t) -> Self {
+        match from {
+            sntp_sync_mode_t_SNTP_SYNC_MODE_SMOOTH => SyncMode::Smooth,
+            sntp_sync_mode_t_SNTP_SYNC_MODE_IMMED => SyncMode::Immediate,
+            _ => unreachable!(),
+        }
+    }
+}
+
+impl From<SyncMode> for sntp_sync_mode_t {
+    fn from(from: SyncMode) -> Self {
+        match from {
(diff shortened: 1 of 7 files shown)


src/sntp.rs
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        let sntp = EspSntp {
+            _sntp_servers: c_servers,
+        };
+
+        sntp_set_time_sync_notification_cb(Some(Self::sync_cb));
+        sntp_init();
+
+        let waiter = WAITER.insert(Waiter::new());
```
ivmarkov, line 146: I'm not sure what is the point of waiting here? To get at least one sync from the ntp server(s)? I think if the user wants to do this, she should be explicitly calling `wait_for_sync` then? What do you think?

src/sntp.rs
```diff
@@ -0,0 +1,191 @@
...
+        waiter.start();
+
+        let c_tz = CString::new("TZ")?;
+        let c_tz_val = CString::new("UTC")?;
```
ivmarkov, line 150: Do we really want to hard-code the timezone as `UTC`? I might want to do `EST`? Also, what do timezones have to do with sntp (which - I hope - is just working with epoch millis (with or without leap seconds, but definitely not concerned with the local timezone))?

ivmarkov (changes requested):

src/sntp.rs
ForsakenHarmony: No, probably don't want to hard-code this, I went off examples from espressif, I'll check again if it's necessary

src/sntp.rs
ForsakenHarmony: We're not waiting here, we're just initializing the Waiter to make sure it's there before the `sync_cb` can be called

src/sntp.rs
```diff
@@ -0,0 +1,191 @@
+use core::{time::Duration};
+
+use anyhow::*;
```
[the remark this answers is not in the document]
ForsakenHarmony: I probably just copied it from one of the things you did, sntp was the first thing I worked on

I'll replace it with `EspError` if possible

src/sntp.rs
```diff
@@ -0,0 +1,191 @@
...
+                    let sntp = Self::init(conf)?;
+
+                    *taken = true;
+                    Ok(sntp)
+                }
+            })
+        }
+    }
+
+    unsafe fn init(conf: SntpConf) -> Result<Self> {
```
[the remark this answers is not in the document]
ForsakenHarmony: will do

src/sntp.rs
```diff
@@ -0,0 +1,191 @@
...
+static mut TAKEN: EspMutex<bool> = EspMutex::new(false);
+static mut WAITER: Option<Waiter> = None;
+
+pub struct EspSntp {
+    // Needs to be kept around because the C bindings only have a pointer.
+    _sntp_servers: [CString; SNTP_SERVER_NUM],
+}
+
+impl EspSntp {
+    pub fn new(conf: SntpConf) -> Result<Self> {
```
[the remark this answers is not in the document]
ForsakenHarmony: sure

src/sntp.rs
ivmarkov: That's true. Sorry for the noise.

ivmarkov: That's true. Sorry for the noise.

ivmarkov:
Since I did not know when (or even if) you would be resuming work on this PR, I've taken it and - with a few changes - merged it: https://github.com/esp-rs/esp-idf-svc/commit/d20b9bd479e82b4c5fe1af6354542da4a06bf656

The changes in details:
- The suggested changes as per my feedback
- Additionally:
  - Make it work in no_std
  - Remove the waiter (even though we have generic `Waitable` now!). I just was not sure what is the benefit of that. If somebody really wants to wait until the first sync with the NTP server had happened, we can always restore/implement a similar functionality

Thanks for your effort!

ForsakenHarmony:
> Remove the waiter (even though we have generic `Waitable` now!). I just was not sure what is the benefit of that. If somebody really wants to wait until the first sync with the NTP server had happened, we can always restore/implement a similar functionality

Why would you not want a way to wait for it?
For me, the purpose was to get a timestamp I can send to influxdb

PR #15, opened 2021-10-26
feat: basic rtos task api

ForsakenHarmony:
"threads" without std

I'm not entirely sure if just stopping them is ok, the [`vTaskDelte` docs](https://www.freertos.org/a00126.html) mention having to free task memory prior to stopping the task

depends on #13

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,6 +44,8 @@ pub mod nvs_storage;
 pub mod ota;
 pub mod ping;
 pub mod sysloop;
+#[cfg(feature = "alloc")]
+pub mod task;
 pub mod time;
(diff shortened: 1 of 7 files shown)


ivmarkov:
See my comments in https://github.com/esp-rs/esp-idf-svc/pull/20
If we abstract away the spawning of a thread with a trait, then we would not need this.

ivmarkov:
Closing. Main argument is as follows: using `no_std` on top of ESP IDF remains to be a solution looking for a problem, really.
Yet - while we still do that for the moment, exposing another thread API (FreeRTOS) in a Rust-safe way (as opposed to just using the unsafe `esp-idf-sys` bindings), seems to be going too far. Why would we do that?

PR #16, opened 2021-10-26
feat: double reset detector

ForsakenHarmony:
Detects if the esp was reset 2 (or more) times during the given duration after booting.

Usage:

```rs
use esp_idf_svc::reset::detect_double_reset;

detect_double_reset(nvs.clone(), Duration::from_secs(3))?;
```

Depends on #15 (and in turn #13)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,6 +43,8 @@ pub mod nvs_storage;
 ))]
 pub mod ota;
 pub mod ping;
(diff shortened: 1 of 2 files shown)


ivmarkov:
Closing because the Pr is too old and in the meantime the crate changed a lot - most of the APIs this PR introduced already exist in a different shape elsewhere.

Also not sure double-reset detection belongs to the library as opposed to user code.

PR #17, opened 2021-10-26
feat: wifi power-save mode

ForsakenHarmony:


src/wifi.rs
```diff
@@ -828,7 +851,7 @@ impl Drop for EspWifi {
     fn drop(&mut self) {
         unsafe {
             TAKEN.lock(|taken| {
-                self.clear_all().unwrap();
+                // self.clear_all().unwrap();
```
ivmarkov, line 854: Forgotten debugging code?

src/wifi.rs
```diff
@@ -301,6 +320,10 @@ impl EspWifi {
         Ok(wifi)
     }
 
+    pub fn set_power_save_mode(&mut self, mode: PowerSaveMode) -> Result<(), EspError> {
```
ivmarkov, line 323: Just for symmetry, perhaps implement also `get_power_safe_mode`?

ivmarkov (changes requested):

src/wifi.rs
ForsakenHarmony: most likely, I think I ran into a crash related to this, but I don't remember for sure

PR #18, opened 2021-10-26
fix: switching wifi mode

ForsakenHarmony:
The ESP crashed when I was trying to switch from ap to client mode before

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -545,17 +545,22 @@ impl EspWifi {
 
         self.shared.with_lock(|shared| shared.operating = false);
 
-        esp!(unsafe { esp_wifi_disconnect() }).or_else(|err| {
-            if err.code() == esp_idf_sys::ESP_ERR_WIFI_NOT_STARTED as esp_err_t {
-                Ok(())
-            } else {
-                Err(err)
-            }
-        })?;
-        info!("Disconnect requested");
+        let status = self.get_status();
+        if status.0 != ClientStatus::Stopped {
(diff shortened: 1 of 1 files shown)


src/wifi.rs
```diff
@@ -545,17 +545,22 @@ impl EspWifi {
...
 
-        esp!(unsafe { esp_wifi_disconnect() }).or_else(|err| {
-            if err.code() == esp_idf_sys::ESP_ERR_WIFI_NOT_STARTED as esp_err_t {
-                Ok(())
-            } else {
-                Err(err)
-            }
-        })?;
-        info!("Disconnect requested");
+        let status = self.get_status();
```
ivmarkov, line 548: This protection is not necessary, as the code below is protected already?

src/wifi.rs
```diff
@@ -545,17 +545,22 @@ impl EspWifi {
...
+                } else {
+                    Err(err)
+                }
+            })?;
+            info!("Disconnect requested");
+        }
 
-        esp!(unsafe { esp_wifi_stop() })?;
-        info!("Stop requested");
+        if status.1 != ApStatus::Stopped {
```
ivmarkov, line 560: Perhaps just do the same protection that is implemented in the `Eth` driver? (Basically, analyzing the error code returned from `esp_wifi_stop`.)

ivmarkov (changes requested):

src/wifi.rs
ForsakenHarmony: I'll try to check again

PR #19, opened 2021-10-26
feat: raw nvs apis

ForsakenHarmony:
Was kinda something I did because I didn't notice you had already written some wrappers 🙃

Also curious about the micro optimization you did with storing data in an `u64`, did that give you a major performance benefit in some scenario?

### DIFF
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -1,9 +1,11 @@
 extern crate alloc;
 
-use ::log::*;
+use core::ops::Deref;
 
+use ::log::*;
 use mutex_trait::*;
 
+use esp_idf_sys::c_types::*;
 use esp_idf_sys::*;
 
 use crate::private::cstr::*;
@@ -13,10 +15,7 @@ static mut NONDEFAULT_LOCKED: EspMutex<alloc::collections::BTreeSet<CString>> =
     EspMutex::new(alloc::collections::BTreeSet::new());
 
 #[derive(Debug)]
-struct PrivateData;
-
-#[derive(Debug)]
-pub struct EspDefaultNvs(PrivateData);
+pub struct EspDefaultNvs(EspNvs);
 
 impl EspDefaultNvs {
     pub fn new() -> Result<Self, EspError> {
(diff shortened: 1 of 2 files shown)


ivmarkov:
To be honest, I haven't gotten to the point of testing my own wrappers, so they should've been marked with "experimental" and possibly hidden behind a feature flag.

With that said, I'm not sure what to do here. I don't think we really need two sets of wrappers. What would be the usefulness of that? Why can't we have just make sure that my wrappers work, and extend as necessary?

Note also that my wrappers try to implement traits defined in `embedded-svc`. The idea - again - is to code against the traits rather than against the implementation and thus create content which is usable also outside of the context of ESP-IDF. Whenever that's possible of course.

ForsakenHarmony:
> With that said, I'm not sure what to do here. I don't think we really need two sets of wrappers. What would be the usefulness of that? Why can't we just make sure that my wrappers work, and extend as necessary?

This is just a safe rust version of the esp nvs apis without another layer on top, maybe there should be another place for that if not in `esp-idf-svc`

> Note also that my wrappers try to implement traits defined in `embedded-svc`. The idea - again - is to code against the traits rather than against the implementation and thus create content which is usable also outside of the context of ESP-IDF. Whenever that's possible of course.

Yes, `embedded-svc` seems like a great idea, thought I feel like experimenting with APIs first before putting them in there would be a good idea as well, but I guess you're doing both at the same time

ivmarkov:
> > With that said, I'm not sure what to do here. I don't think we really need two sets of wrappers. What would be the usefulness of that? Why can't we just make sure that my wrappers work, and extend as necessary?
> 
> This is just a safe rust version of the esp nvs apis without another layer on top, maybe there should be another place for that if not in `esp-idf-svc`
> 
> > Note also that my wrappers try to implement traits defined in `embedded-svc`. The idea - again - is to code against the traits rather than against the implementation and thus create content which is usable also outside of the context of ESP-IDF. Whenever that's possible of course.
> 
> Yes, `embedded-svc` seems like a great idea, thought I feel like experimenting with APIs first before putting them in there would be a good idea as well, but I guess you're doing both at the same time

Yes, I'm trying to do both, as risky as it is when you have just one use case at hand (esp32), as you might start modeling your use case (esp32) in your APIs and thus get a leaky API abstraction. With that said, I still believe having leaky APIs is more useful than having no APIs at all, as it makes you think hard what is the gist of functionality you really would like to expose to Rust folks - that they really can't get otherwise - rather than just blindly exposing everything there is in the ESP-IDF.

But my point is a different one: now that I took the effort of writing some wrappers, and then exposing them in rust, how is exposing another set of wrappers - next to the original ones - making `esp-idf-svc` a better thing to use? Doesn't it bring confusion? How are we going to explain this to users?

ForsakenHarmony:
> how is exposing another set of wrappers - next to the original ones - making `esp-idf-svc` a better thing to use? Doesn't it bring confusion? How are we going to explain this to users?

That's why I said "maybe there should be another place for that if not in `esp-idf-svc`"
A place for more raw (but safe rust) wrappers

ivmarkov:
> > how is exposing another set of wrappers - next to the original ones - making `esp-idf-svc` a better thing to use? Doesn't it bring confusion? How are we going to explain this to users?
> 
> That's why I said "maybe there should be another place for that if not in `esp-idf-svc`" A place for more raw (but safe) wrappers

If there is indeed such a pressing need to have 1:1 wrappers for the C-level APIs, extending/merging my code with yours, where the resulting thing implements the embedded-svc trait *and* has extra APIs not covered by the trait is still better than two separate implementations sitting next to each other.

ivmarkov:
> > > how is exposing another set of wrappers - next to the original ones - making `esp-idf-svc` a better thing to use? Doesn't it bring confusion? How are we going to explain this to users?
> > 
> > 
> > That's why I said "maybe there should be another place for that if not in `esp-idf-svc`" A place for more raw (but safe) wrappers
> 
> If there is indeed such a pressing need to have 1:1 wrappers for the C-level APIs, extending/merging my code with yours, where the resulting thing implements the embedded-svc trait _and_ has extra APIs not covered by the trait is still better than two separate implementations sitting next to each other.

I don't see this PR going anywhere. As per above ^^^ folding these ESP-specific APIs in `nvs_storage.rs` as public methods would've been a way, but this was not followed on. Closing.

PR #20, opened 2021-10-26
feat: captive portal dns (+lwip apis)

ForsakenHarmony:
Implemented the captive portal dns from the esp idf example

To use it:

1. Start wifi in ap mode
2. Start captive portal DNS
3. Start http server with 404 handler (#11), which redirects to `/` (important!) 

The captive portal should show up when you log into wifi

[Example gist with working code](https://gist.github.com/ForsakenHarmony/9206d4e94862ddd5e00957708e3c8498)

Depends on #15 (and in turn #13)

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -1,0 +1,305 @@
+use core::mem;
+
+use anyhow::*;
+use log::*;
+
+use esp_idf_sys::c_types::*;
+use esp_idf_sys::*;
+
+use crate::private::cstr::{CStr, CString};
+use crate::task::{TaskConfig, TaskHandle};
+use crate::lwip;
+
+/// 0.0.0.0
+const IPADDR_ANY: u32 = 0x00000000;
+
+const DNS_PORT: c_ushort = 53;
+const DNS_MAX_LEN: usize = 256;
+
+const OPCODE_MASK: u16 = 0x7800;
+const QR_FLAG: u16 = 1 << 7;
+const QD_TYPE_A: u16 = 0x0001;
+const ANS_TTL_SEC: u32 = 300;
+
+pub struct CaptivePortalDns {
+    task_handle: Option<TaskHandle>,
+}
+
+impl CaptivePortalDns {
+    pub fn new() -> Self {
+        CaptivePortalDns { task_handle: None }
+    }
+
+    pub fn start(&mut self) -> Result<()> {
+        if self.task_handle.is_some() {
+            bail!("dns server is already running");
+        }
+
+        let handle = TaskConfig::default()
+            .priority(5)
+            .spawn("dns_server", dns_server_task)?;
+
+        self.task_handle = Some(handle);
+        Ok(())
+    }
+
+    pub fn stop(&mut self) -> Result<()> {
+        if let Some(handle) = self.task_handle.take() {
+            handle.stop();
+            Ok(())
+        } else {
+            Err(anyhow!("dns task already stopped or was never started"))
+        }
+    }
+}
+
+impl Drop for CaptivePortalDns {
+    fn drop(&mut self) {
+        if self.task_handle.is_some() {
+            self.stop().unwrap();
+        }
+    }
+}
+
+/// DNS Header Packet
+#[repr(C, packed)]
+struct DnsHeader {
+    id: u16,
+    flags: u16,
+    qd_count: u16,
+    an_count: u16,
+    ns_count: u16,
+    ar_count: u16,
+}
+
+/// DNS Question Packet
+#[repr(C)]
+struct DnsQuestion {
+    typ: u16,
+    class: u16,
+}
+
+/// DNS Answer Packet
+#[repr(C, packed)]
+struct DnsAnswer {
+    name_ptr: u16,
+    typ: u16,
+    class: u16,
+    ttl: u32,
+    addr_len: u16,
+    ip_addr: u32,
+}
+
+fn parse_dns_name(raw_name: *mut u8, parsed_name: &mut [u8]) -> *mut u8 {
+    let mut label = raw_name;
+    let parsed_name_max_len = parsed_name.len();
+    let mut name_itr = parsed_name.iter_mut();
+    let mut name_len: usize = 0;
+
+    loop {
+        let sub_name_len = unsafe { *label as c_int };
+        // (len + 1) since we are adding  a '.'
+        name_len += (sub_name_len + 1) as usize;
+        if name_len > parsed_name_max_len {
+            return core::ptr::null_mut();
+        }
+
+        // Copy the sub name that follows the the label
+        for i in 0..sub_name_len {
+            let ptr = name_itr.next().unwrap();
+            *ptr = unsafe { *label.offset((i + 1) as isize) };
+        }
+        *name_itr.next().unwrap() = '.' as u8;
+        label = unsafe { label.offset((sub_name_len + 1) as isize) };
+
+        if unsafe { *label == 0 } {
+            break;
+        }
+    }
+
+    // Terminate the final string, replacing the last '.'
+    parsed_name[name_len - 1] = '\0' as u8;
+    // Return pointer to first char after the name
+    return unsafe { label.offset(1) };
+}
+
+fn parse_dns_request(
+    req: &mut [u8],
+    req_len: usize,
+    dns_reply: &mut [u8],
+    dns_reply_max_len: usize,
+) -> Option<usize> {
+    if req_len > dns_reply_max_len {
+        return None;
+    }
+
+    // Prepare the reply
+    dns_reply.fill(0);
+    (&mut dns_reply[0..req_len]).copy_from_slice(&req[0..req_len]);
+
+    let header_len = mem::size_of::<DnsHeader>();
+    let (header_bytes, rest) = dns_reply.split_at_mut(header_len);
+
+    // Endianess of NW packet different from chip
+    let header = unsafe {
+        header_bytes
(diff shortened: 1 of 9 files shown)


src/dns.rs
```diff
@@ -0,0 +1,305 @@
+use core::mem;
```
ivmarkov, line 1: I must say, I *LOVE* the idea of having a Captive Portal!
What I don't love that much:
* Overall, the code feels very low level (as if we are coding in C, via Rust) - I'll comment more below
* It is part of `esp-idf-svc` but I'm not sure what it is doing here? This is a generic DNS server, which can - just as well - be coded against Rust's STD APIs (correct me if I'm wrong!) - or better yet - against the [embedded-nal ](https://github.com/rust-embedded-community/embedded-nal/blob/master/src/stack/udp.rs#L50) so that it does have a `no_std` story as well. (`embedded-nal` traits do have STD implementations too, and if these are lacking, we can implement / patch / upstream our own changes.). So I'm very much in favor of moving this into `embedded-svc` (which depends on `no-std-net` already anyway) - possibly behind a feature toggle.
* As for what would be the use-case of that? Well, I might want to use a captive dns portal on the Raspberry Pi too?

src/dns.rs
```diff
@@ -0,0 +1,305 @@
...
+const QD_TYPE_A: u16 = 0x0001;
+const ANS_TTL_SEC: u32 = 300;
+
+pub struct CaptivePortalDns {
+    task_handle: Option<TaskHandle>,
+}
+
+impl CaptivePortalDns {
+    pub fn new() -> Self {
+        CaptivePortalDns { task_handle: None }
```
ivmarkov, line 30: I guess the desire to stay `no_std` compatible is what  brought you into exposing bits and pieces for FreeRTOS? How about we take a simple trait instead, that would allow us to spawn a thread? We can even provide an implementation of the trait when Rust STD is enabled, and provide additional constructor which does not take the trait, for STD mode.

src/dns.rs
```diff
@@ -0,0 +1,305 @@
...
+struct DnsAnswer {
+    name_ptr: u16,
+    typ: u16,
+    class: u16,
+    ttl: u32,
+    addr_len: u16,
+    ip_addr: u32,
+}
+
+fn parse_dns_name(raw_name: *mut u8, parsed_name: &mut [u8]) -> *mut u8 {
```
ivmarkov, line 93: This and the next function are littered with `unsafe` statements. Would it be possible to stay in Rust safe-land for parsing this stuff? Only if possible?

src/dns.rs
```diff
@@ -0,0 +1,305 @@
...
+        if qd_type == QD_TYPE_A {
+            let answer = unsafe { cur_ans_ptr.cast::<DnsAnswer>().as_mut().unwrap() };
+
+            let ptr_offset = unsafe { cur_qd_ptr.offset_from(dns_reply.as_ptr()) };
+            answer.name_ptr = htons((0xC000 | ptr_offset) as u16);
+            answer.typ = htons(qd_type);
+            answer.class = htons(qd_class);
+            answer.ttl = htonl(ANS_TTL_SEC);
+
+            let mut ip_info = esp_netif_ip_info_t::default();
```
ivmarkov, line 212: In an implementation which is independent of ESP-IDF, we would want the IP address to be abstracted away, via a trait, or passed at construction/start time.

ivmarkov (changes requested):

src/dns.rs
ForsakenHarmony: The code is indeed C converted to Rust, but not quite at the point where I'm super happy with the "Rustyness" of it yet

Sure, it would be nice to have it generic.
But my use case wasn't quite as generic, so I implemented it specifically for the esp right now.

`embedded-nal` seems nice, hadn't heard of it before.

I feel like someone else must have already implemented a captive portal dns in rust that's more generic, though. (not that I checked and not that I'm opposed to make this more generic)

src/dns.rs
ForsakenHarmony: > I guess the desire to stay `no_std` compatible is what brought you into exposing bits and pieces for FreeRTOS?
Yes

> How about we take a simple trait instead, that would allow us to spawn a thread? We can even provide an implementation of the trait when Rust STD is enabled, and provide additional constructor which does not take the trait, for STD mode.

As in, implement it with `std::thread` when we're using `std` and otherwise add an impl with FreeRTOS?

ivmarkov: As in implement it with `std::thread` when we are using `std` and *don't provide an implementation at all* for `no_std`. With `no_std`, user would be on their own to provide an implementation in whatever way they feel like.

It boils down to the following: *we have to keep the `esp-idf-svc` surface small, clean and easy to explain by exposing Rust-safe, self-contained pieces of functionality, - ideally - behind generic traits*. The emphasis being on safe and self-contained. We cannot just go there and expose random bits and pieces of all the raw level unsafe APIs that are out there in the ESP-IDF. For example, in your "basic freertos task api" what you have exposed from FreeRTOS is just a way to start a task (and that is always pinned to a core, if I'm not mistaken). Is that API generic enough so that folks can use it without calling other, unsafe Freertos APIs? Well, no.  For one, the queueing and synchronization primitives of freertos are not exposed as safe APIs, because you did not need them for your particular case. But very likely people *will* need these to do something meaningful with freertos. So, how are we evolving your Freertos code? Is it going to - at some point in time - reach a feature parity with the unsafe freertos APIs? 
* If yes, then we are basically creating a whole Freertos wrapper, and shouldn't we then instead work towards making [some of the Freertos Rust wrappers interop-ing with `esp-idf-svc`](https://github.com/lobaro/FreeRTOS-rust) interop-ing with `esp-idf-sys`?
 * If no, aren't we in a better position by saying to folks that there are NO freertos safe APIs, and they should just use the unsafe bindings which are out there in `esp-idf-sys`?

ivmarkov:
OK so the real trickiness of all this code is in `parse_dns_request`. The `socket` stuff is just a 3-liner that receives DNS requests and replies in a loop. Everybody can write these 3 lines.

Now, the nice thing is - everything in `parse_dns_request` is ESP-IDF-agnostic (that is, aside from it always taking and serving the ESP-IDF Wifi gateway IP).

I still don't like it that there are so many unsafe statements in the above function, but my little mind cannot figure out for now how to go from byte buffers to Rust structures and back with less unsafe, so the unsafes will stay intact (for now).

My plan is the following:
- I'll take your `parse_dns_request` almost verbatim and move it into `embedded-svc`, as a `pub` static function in a `dns::CaptivePortalDns` structure. It will be generalized just a bit, so that it can "ask" - via a closure - what IP to return for a particular domain name, but other than that the signature (and the code) would stay intact.
- The socket 3-liner I'll implement only when the code is compiled against STD. In `no_std`, users would have to write the socket 3-liner themselves, and then call the magic `parse_dns_request` to process the incoming UDP packet.

Any glaring holes in this plan that you might spot?

ForsakenHarmony:
> I still don't like it that there are so many unsafe statements in the above function, but my little mind cannot figure out for now how to go from byte buffers to Rust structures and back with less unsafe, so the unsafes will stay intact (for now).

Should be possible to use an iterator I think, there were more pointers in there I converted

Just for reference, this is the example I converted https://github.com/espressif/esp-idf/blob/4a011f318377ab717b9dc459019c1055602c93ff/examples/protocols/http_server/captive_portal/main/dns_server.c

> * I'll take your `parse_dns_request` almost verbatim and move it into `embedded-svc`, as a `pub` static function in a `dns::CaptivePortalDns` structure. It will be generalized just a bit, so that it can "ask" - via a closure - what IP to return for a particular domain name, but other than that the signature (and the code) would stay intact.
Could also pass in an IP when you start the dns server, it's probably unlikely to change

> * The socket 3-liner I'll implement only when the code is compiled against STD. In `no_std`, users would have to write the socket 3-liner themselves, and then call the magic `parse_dns_request` to process the incoming UDP packet.

Would be great to have an impl with `embedded-nal` I suppose, but that should do for now

ivmarkov:
> > I still don't like it that there are so many unsafe statements in the above function, but my little mind cannot figure out for now how to go from byte buffers to Rust structures and back with less unsafe, so the unsafes will stay intact (for now).
> 
> Should be possible to use an iterator I think, there were more pointers in there I converted
> 
> Just for reference, this is the example I converted https://github.com/espressif/esp-idf/blob/4a011f318377ab717b9dc459019c1055602c93ff/examples/protocols/http_server/captive_portal/main/dns_server.c
> 
> > * I'll take your `parse_dns_request` almost verbatim and move it into `embedded-svc`, as a `pub` static function in a `dns::CaptivePortalDns` structure. It will be generalized just a bit, so that it can "ask" - via a closure - what IP to return for a particular domain name, but other than that the signature (and the code) would stay intact.
> 
> Could also pass in an IP when you start the dns server, it's probably unlikely to change
> 
> > * The socket 3-liner I'll implement only when the code is compiled against STD. In `no_std`, users would have to write the socket 3-liner themselves, and then call the magic `parse_dns_request` to process the incoming UDP packet.
> 
> Would be great to have an impl with `embedded-nal` I suppose, but that should do for now

I was halfway there in removing the unsafes over the weekend when I found [this](https://github.com/NLnetLabs/domain). To be expected of course that such a thing exists in the first place. I'm switching to it instead.

ivmarkov:
Closing in favor of https://github.com/esp-rs/embedded-svc/commit/759b9962664d8ca96af7e9327a5596b56bede0ce

WIP and not tested yet, but I hope to get it working over the weekend.

PR #23, opened 2021-11-08
Making put_raw and get_raw work

jakobdalsgaard:
Solving some issues with pointers for the in/out buffers to esp-idf.
Also settles solely on 'blobs' and uses none of the other datatypes
of esp idf nvs storage.

I did some debugging on how to make nvs storage work on the ESP32 -- and this is what I came up with; most importantly it seemed that the data buffers were not correctly transferred to esp idf in the nvs_get_* and nvs_set_* functions. Furthermore it seems like the esp idf nvs storage functions do not allow to mix datatypes, i.e. doing an 'nvs_get_u64' on a piece of data that is stored as a blob does not given 'incorrect length' as error, but rather 'not found' -- thus some logic was not working.

This implementation settles on blobs and uses nothing else; while blob support is sufficient and very desirable in that it forces developers to consider the encoding of all the data they wish to store, there might be some performance and scalability gains of supporting more datatypes. Ultimately some design decision must be taken.

### DIFF
--- a/src/nvs_storage.rs
+++ b/src/nvs_storage.rs
@@ -77,13 +77,29 @@ impl Storage for EspNvsStorage {
     fn contains(&self, key: impl AsRef<str>) -> Result<bool, Self::Error> {
         let c_key = CString::new(key.as_ref()).unwrap();
 
-        let dummy: u_int64_t = 0;
+        let mut dummy: u_int64_t = 0;
 
-        match unsafe { nvs_get_u64(self.1, c_key.as_ptr(), dummy as *mut _) } as u32 {
-            ESP_ERR_NVS_NOT_FOUND => Ok(false),
-            ESP_ERR_NVS_INVALID_LENGTH => Ok(true),
+        // check if key is present for u64 datatype
+        match unsafe { nvs_get_u64(self.1, c_key.as_ptr(), &mut dummy as *mut _) } as u32 {
+            ESP_ERR_NVS_NOT_FOUND => {
+                // now check if key is present for blob datatype
+                let mut len: size_t = 0;
+                match unsafe {
+                    nvs_get_blob(self.1, c_key.as_ptr(), ptr::null_mut(), &mut len as *mut _)
+                } as u32
+                {
+                    // not found as u64, nor as blob, this key has not been found
+                    ESP_ERR_NVS_NOT_FOUND => Ok(false),
+                    result => {
+                        // bail on any kind of error, return true if no error
+                        esp!(result)?;
+                        Ok(true)
+                    }
+                }
+            }
             result => {
                 esp!(result)?;
+                // if we get here, the value was found as a u64
                 Ok(true)
             }
         }
@@ -92,6 +108,7 @@ impl Storage for EspNvsStorage {
     fn remove(&mut self, key: impl AsRef<str>) -> Result<bool, Self::Error> {
         let c_key = CString::new(key.as_ref()).unwrap();
 
+        // nvs_erase_key is not scoped by datatype
         let result = unsafe { nvs_erase_key(self.1, c_key.as_ptr()) };
 
         if result == ESP_ERR_NVS_NOT_FOUND as i32 {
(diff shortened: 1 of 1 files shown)


ivmarkov:
Thank you for your contribution!

> Solving some issues with pointers for the in/out buffers to esp-idf. 
Can you point me to the code-lines in your commit where this is solved? Can't see that...

Also settles solely on 'blobs' and uses none of the other datatypes of esp idf nvs storage.
> 
> I did some debugging on how to make nvs storage work on the ESP32 -- and this is what I came up with; most importantly it seemed that the data buffers were not correctly transferred to esp idf in the nvs_get_* and nvs_set_* functions. Furthermore it seems like the esp idf nvs storage functions do not allow to mix datatypes, i.e. doing an 'nvs_get_u64' on a piece of data that is stored as a blob does not given 'incorrect length' as error, but rather 'not found' -- thus some logic was not working.
> 
> This implementation settles on blobs and uses nothing else; while blob support is sufficient and very desirable in that it forces developers to consider the encoding of all the data they wish to store, there might be some performance and scalability gains of supporting more datatypes. Ultimately some design decision must be taken.

Indeed. I'm a bit worried that storing everything as blobs might be a too high price to pay. We might still be able to implement my `u64` optimization though, even if - exactly as you say - the ESP-IDF API behaves differently from expectations in that it gives unique namespace for each datatype. By the way, implementing optimizations for smaller data types is not worth it, as everything with size < u64 is expanded to 8 bytes anyway.

The other question is to what extent we would like to interop with C code that is saving stuff in the NVS storage. Im that case, we might have to implement all get/set methods of the NVS api, including these for strings. But that might be a second stage.

jakobdalsgaard:
> Can you point me to the code-lines in your commit where this is solved? Can't see that...

Hmm... it appears that this part was really only in my working notes; here they are:

I found that this construct:

```
        let mut value_ptr: *mut u64 = &mut 0;
        match unsafe { nvs_get_u64(self.1, c_key.as_ptr(), value_ptr as *mut _) } as u32 {
```

Works, whereas the construct found in nvs_storage.rs before was:

```
        let mut value: u_int64_t = 0;
        match unsafe { nvs_get_u64(self.1, c_key.as_ptr(), value as *mut _) } as u32 {
```

Which broke with 'StoreProhibited'

Datatype namespace wise all the points you make are valid :-) I'm happy to have blobs for now 🥳

ivmarkov:
> > Can you point me to the code-lines in your commit where this is solved? Can't see that...
> 
> Hmm... it appears that this part was really only in my working notes; here they are:
> 
> I found that this construct:
> 
> ```
>         let mut value_ptr: *mut u64 = &mut 0;
>         match unsafe { nvs_get_u64(self.1, c_key.as_ptr(), value_ptr as *mut _) } as u32 {
> ```
> 
> Works, whereas the construct found in nvs_storage.rs before was:
> 
> ```
>         let mut value: u_int64_t = 0;
>         match unsafe { nvs_get_u64(self.1, c_key.as_ptr(), value as *mut _) } as u32 {
>```

Indeed and stupid me. Strange how Rust even allowed the above to compile?! It should be:
> ```
>         let mut value: u_int64_t = 0;
>         match unsafe { nvs_get_u64(self.1, c_key.as_ptr(), &mut value as *mut _) } as u32 {
>```

> Datatype namespace wise all the points you make are valid :-) I'm happy to have blobs for now 🥳

I might try to restore the u64 code if you don't mind.
The other missing thing is: you have not touched the `contains` and `remove` methods?

jakobdalsgaard:
I certainly do not mind restoring the u64 code, I can have a go at the contains and remove code :-)

and yes, I'm a bit baffled by what Rust allows in unsafe bindings ...

jakobdalsgaard:
With this implementation:

```
    fn contains(&self, key: impl AsRef<str>) -> Result<bool, Self::Error> {
        let c_key = CString::new(key.as_ref()).unwrap();

        let mut dummy: u_int64_t = 0;

        match unsafe { nvs_get_u64(self.1, c_key.as_ptr(), &mut dummy as *mut _) } as u32 {
            ESP_ERR_NVS_NOT_FOUND => Ok(false),
            ESP_ERR_NVS_INVALID_LENGTH => Ok(true),
            result => {
                esp!(result)?;
                Ok(true)
            }
        }
    }
```

it works... but it's limited to `u64` -- and thus out of sync with the rest of the implementation that works on blobs -- I can make this one work on blobs too - but if you're rewriting the stuff anyway, you might want to have a version that embodies your ideas :-)

jakobdalsgaard:
I'll have go at it tomorrow then :-)

jakobdalsgaard:
I believe the `remove` one is the odd one -- as it does not take datatype at all -- it must operate across all datatypes. Less work for me.

ivmarkov:
> I'll have go at it tomorrow then :-)

Seems tomorrow never came? :-)

jakobdalsgaard:
:-) I'll be posting something today ... I'm testing out some cornercases

jakobdalsgaard:
So ... @ivmarkov -- got something now ... have a look at this code. I've been trying to break stuff for a while, but this solution seems solid :-)

This example: https://gist.github.com/jakobdalsgaard/5105f36fbc6885cc9e8b69fcc368c8ed -- completes fine.

The encoding scheme done on u64 is fine - but I guess it's specific to this implementation, making integration with the existing toolsuite for NVS storage a bit tricky. To feed the application, however, you can always create the NVS data with the toolsuite by just using blobs -- they'll get read by this implementation even if they're less than 7 bytes.

Enjoy.

ivmarkov:
> So ... @ivmarkov -- got something now ... have a look at this code. I've been trying to break stuff for a while, but this solution seems solid :-)

Thanks!

> The encoding scheme done on u64 is fine - but I guess it's specific to this implementation, making integration with the existing toolsuite for NVS storage a bit tricky. To feed the application, however, you can always create the NVS data with the toolsuite by just using blobs -- they'll get read by this implementation even if they're less than 7 bytes.

Not sure I understand what you mean by `toolsuite`? If you mean that this implementation cannot really interop with the C/C++ API in ESP-IDF (as in e.g. writing a string from ESP-IDF side wont be read correctly on the Rust side using this implementation) - that's indeed correct.

However I was not really sure what is the use case for that interop **in the first place**? If you know of any use cases where we would need to e.g. read data that was written by C/C++ or the other way around - please let me know. If there are indeed such use cases, perhaps we can merge bits and pieces from [this pull request](https://github.com/esp-rs/esp-idf-svc/pull/19/files) into `EspNvsStorage` as a public API, but without a corresponding trait in `embedded-svc`? This way we would have a feature parity and interop with the C/C++ side in case somebody really needs this interop. But then again: use case?

jakobdalsgaard:
I've been using the python scripts for making binary NVS files to upload from CSV files - it's weird toolsuite but it kinda works; imho, for debugging purposes having these tools is nice. It's not that it's extremely hard to encode the optimised u64 - it's just an extra step (and it's extra logic in the nvs_storage.rs file). After all, it is possible to put in a 3 byte blob with the esp python tools and have it read by this implementation; if you then store it from the ESP itself, it'll get autoconverted to u64. No big deal.

As for having full access to the datatypes in NVS. It's probably something to be looked at, not very important though. As I see it there is a slim use case where someone has a C/C++ implementation and they OTA update to a Rust based implementation. In that case you'd want to read what's on the NVS partition. For any other type of updates, you most likely either have no existing NVS data or you have the possibility to read them all out and put in again.

I wish the ESP IDF implementation would have had a get_blob that would access any kind of bytes... and that all the dataypes would've been implemented on top of that. That way we could just implement the get_blob method and magically be able to access any datatype stored.

But reality is different.

ivmarkov:
OK and thanks again.

If you are not interested, then I might myself be looking into merging @ForsakenHarmony 's "C/C++-compatible" bindings into `EspNvsStorage` these days. You know, "just in case". But as you say, not urgent.

jakobdalsgaard:
👍 

I'd also be happy to give it a shot - but I won't have time to do so for the next few weeks unfortunately.

PR #25, opened 2021-11-13
An error occurred when sdkconfig set CONFIG_LOG_DEFAULT_LEVEL_NONE=y 

planet0104:
function or associated item not found in `esp_idf_sys::log`


ivmarkov:
Thanks!

PR #27, opened 2021-11-20
added an option to set hostname in STA mode

UnTraDe:
This pull request depends on: https://github.com/esp-rs/embedded-svc/pull/3

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -111,7 +111,7 @@ impl From<&Newtype<wifi_sta_config_t>> for ClientConfiguration {
                 None
             },
             ip_conf: None, // This must be set at a later stage
-			hostname: None
+            hostname: None,
         }
     }
 }
@@ -353,11 +353,11 @@ impl EspWifi {
 
         esp!(unsafe { esp_wifi_set_config(wifi_interface_t_WIFI_IF_STA, &mut wifi_config) })?;
 
-		let hostname = if let Some(hn) = &conf.hostname {
-			Some(hn.as_str())
-		} else {
-			None
-		};
+        let hostname = if let Some(hn) = &conf.hostname {
+            Some(hn.as_str())
+        } else {
+            None
+        };
 
         self.set_client_ip_conf(&conf.ip_conf, hostname)?;
 
@@ -396,7 +396,7 @@ impl EspWifi {
     fn set_client_ip_conf(
         &mut self,
         conf: &Option<ipv4::ClientConfiguration>,
-		hostname: Option<&str>
+        hostname: Option<&str>,
     ) -> Result<(), EspError> {
         Self::netif_unbind(self.sta_netif.as_mut())?;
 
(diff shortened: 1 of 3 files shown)


src/wifi.rs
```diff
@@ -352,7 +353,13 @@ impl EspWifi {
 
         esp!(unsafe { esp_wifi_set_config(wifi_interface_t_WIFI_IF_STA, &mut wifi_config) })?;
 
-        self.set_client_ip_conf(&conf.ip_conf)?;
+        let hostname = if let Some(hn) = &conf.hostname {
```
ivmarkov, line 356: Thank you for your contribution and for fixing the formatting error.

If you don't mind I would request a few extra changes. See below for why and how.

1) The notion of "hostname" is not really a Wifi STA mode specific thing. It is a DHCP-client specific thing, which - by the way - should work just fine for the Eth driver as well. The `wifi::ClientConfiguration` WiFi struct is a network L2 thing (where you've put the hostname), while the `ipv4::ClientConfiguration` is a network L3 thing (IP) where the hostname actually belongs.

2) Moreover, since the `hostname` is something related to DHCP *only*, I suggest you model it as follows (this is in `embedded-svc/ipv4.rs`:
```
#[derive(Clone, Debug, PartialEq)]
#[cfg_attr(feature = "use_serde", derive(Serialize, Deserialize))]
pub struct DHCPClientSettings {
    pub hostname: Option<String>,
}

impl Default for DHCPClientSettings {
    fn default() -> Self {
        Self {
            hostname: None,
        }
    }
}

#[derive(Clone, Debug, PartialEq)]
#[cfg_attr(feature = "use_serde", derive(Serialize, Deserialize))]
pub enum ClientConfiguration {
    DHCP(DHCPClientSettings), // This is new, just like the DHCPClientSettings struct. It used to be just "DHCP,"
    Fixed(ClientSettings),
}
```

src/wifi.rs
```diff
@@ -403,6 +411,15 @@ impl EspWifi {
             esp!(unsafe { esp_netif_attach_wifi_station(netif.1) })?;
             esp!(unsafe { esp_wifi_set_default_wifi_sta_handlers() })?;
 
+            if let Some(hn) = hostname {
+                if let Ok(hostname) = CString::new(hn) {
+                    esp!(unsafe { esp_netif_set_hostname(netif.1, hostname.as_ptr()) })?;
```
ivmarkov, line 416: Minor nit, but I suggest you surface two new public members in the ESP-IDF Netif wrapper, `EspNetif`:
* `get_hostname(&self)`
* `set_hostname(&self...)`

... and then call `set_hostname` here.

You can model `get_hostname` by roughly copy-pasting the code from the existing `get_name` function. For `set_name` I think you've figured it out already how it should look (roughly).

ivmarkov (changes requested):

ivmarkov:
Note also that Clippy fails to compile your code (it is not clippy warnings, but the code really somehow does not compile?) so you have to fix that as well. However - given that the changes I suggested move the hostname to a different structure altogether, I suggest you don't bother and instead perhaps look into the suggested changes instead.

ivmarkov:
> Note also that Clippy fails to compile your code (it is not clippy warnings, but the code really somehow does not compile?) so you have to fix that as well. However - given that the changes I suggested move the hostname to a different structure altogether, I suggest you don't bother and instead perhaps look into the suggested changes instead.

^^^ Disregard. it fails to compile because of course it needs the corresponding changes in embedded-svc. :)

src/wifi.rs
UnTraDe: Can't it be set for `Fixed` as well?

src/wifi.rs
UnTraDe: Yeah that looks like a better fit. But then shouldn't I call `set_hostname` at `EspNetif::new`?

ivmarkov (approved):
Great work, thank you!

PR #29, opened 2021-11-23
HTTP Client: Use per-client event handlers

bugadani:
This PR indents to replace the global shared BTreeMap with private event handler pointers, thus saving a lock and a lookup.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -176,7 +176,7 @@ pub struct EspHttpRequest<'a> {
 }
 
 impl<'a> EspHttpRequest<'a> {
-    fn register_handler(&self, handler: &mut impl Fn(&esp_http_client_event_t) + 'static) {
+    fn register_handler(&self, handler: &mut (impl Fn(&esp_http_client_event_t) + 'static)) {
         *self.event_handler = Some(unsafe {
             // SAFETY: reference to pointer cast always results in a non-null pointer.
             NonNull::new_unchecked(handler as *mut _)
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -51,10 +51,6 @@ impl From<HttpMethod> for Newtype<(esp_http_client_method_t, ())> {
     }
 }
 
-static mut EVENT_HANDLERS: EspMutex<
-    BTreeMap<esp_http_client_handle_t, *const dyn Fn(&esp_http_client_event_t)>,
-> = EspMutex::new(BTreeMap::new());
-
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 #[cfg_attr(feature = "std", derive(Hash))]
 pub enum FollowRedirectsPolicy {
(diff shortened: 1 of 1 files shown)


ivmarkov:
Heh. I was looking for this `user_data` backdoor and couldn't believe my eyes that it is not available for the ESP-IDF HTTP client. Apparently, I was not looking carefully enough - hence the btree + mutex monstrocity :)

 Anyway - I have one comment (look at the review) and then we can merge that. Code looks very decent!

src/http/client.rs
```diff
@@ -75,9 +71,12 @@ pub struct EspHttpClientConfiguration {
     pub follow_redirects_policy: FollowRedirectsPolicy,
 }
 
+trait EventHandler = Fn(&esp_http_client_event_t) + 'static;
```
ivmarkov, line 74: Is this a [trait alias](https://doc.rust-lang.org/beta/unstable-book/language-features/trait-alias.html)?
Was this stabilized recently and I just missed that (wondering how it even works without `#![feature(trait_alias)]`)? If not stabilized, I wonder if we want it (I already have tons of problems with GATs + HRTBs) so I don't want to make the whole thing even less stable in terms of compiler crashes.

ivmarkov (changes requested):

src/http/client.rs
bugadani: Oh shoot, it is, and I didn't intend to introduce any unstable features (I don't think it has been stabilized yet?!). I'll remove the alias ASAP.

bugadani: I suspect this code is excluded by some feature flag that is not used in the check build?

ivmarkov: Yes. you need a `--features experimental` of course. This is BTW why the CI did not catch it. I was lazy and did not implement (yet) a run with that one ^^^; neither a run with no_std. We need both actually, to catch all errors.

ivmarkov: By the way, I DID introduce one unstable feature of course: GATs. But my hope is it is really close to being stabilized, and the ESP-IDF story is anyway a nightly-only thing (because `cargo -Z build-std` is for now a nightly-only thing) but I'm not keen on introducing more unstable features.

bugadani:
Fair warning: I don't have the setup to build this locally, so even without using the `trait_alias` feature, the PR might not compile as is.

ivmarkov:
> Fair warning: I don't have the setup to build this locally, so even without using the `trait_alias` feature, the PR might not compile as is.

OK then let me actually test it tmr before merging :)

src/http/client.rs
```diff
@@ -105,23 +104,27 @@ impl EspHttpClient {
...
-
-        let handler =
-            unsafe { EVENT_HANDLERS.lock(|handlers| handlers.get(&event.client).copied()) };
+        match unsafe { event.as_mut() } {
+            Some(event) => {
+                let event_handler = unsafe {
+                    (event.user_data as *const Option<NonNull<dyn Fn(&esp_http_client_event_t)>>).as_ref()
+                };
+                if let Some(opt_handler) = event_handler {
+                    let handler = unsafe { opt_handler.unwrap().as_mut() };
```
bugadani, line 119: If I can believe my code (which uses a similar, but not identical http client implementation), this unwrap might actually panic.

ivmarkov: I do not understand this statement. Isn't this *your* code?

bugadani: The change in the PR is mine, but it was originally made for a different http client, not the one in this repository. What I mean by my code, means that other client and the application that uses it. Not important, really.

ivmarkov:
I've implemented something along the lines of your code here: https://github.com/esp-rs/esp-idf-svc/commit/9c81cf63577949495906a23cfb880b6ff3b9b4e9

PR #30, opened 2021-11-24
fix sending data in HttpClient

fkohlgrueber:
I noticed that sending data using the http client was broken. The root cause is the error handling in `do_write`. `esp_http_client_write` returns the number of bytes written, but the `esp!` macro converts every non-zero return code into an error.

In `do_read`, the implementation is correct already. I used the error handling from `do_read` in `do_write` now, which should be correct. I already tried the changes and they seem to work fine.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -295,9 +295,13 @@ impl<'a> Write for EspHttpRequest<'a> {
     type Error = EspError;
 
     fn do_write(&mut self, buf: &[u8]) -> Result<usize, Self::Error> {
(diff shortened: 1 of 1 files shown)


MabezDev (approved):
LGTM

ivmarkov:
Pls do a `cargo+nightly fmt

fkohlgrueber:
done

ivmarkov:
Hmm, still not OK. Perhaps you can look at the CI cargo fmt output and manually format the code as requested? Sorry - I need the formatting fixed, because otherwise the CI will stay broken. :)

ivmarkov:
Thanks!

PR #33, opened 2021-12-12
added initial support for SmartConfig

UnTraDe:
This adds initial support for SmartConfig. (https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_smartconfig.html)

SmartConfig is a protocol that allows sending SSID connection info (password, bssid, etc) to an unconnected ESP32 via another device connected to the same network.

Currently only SmartConfig V1 is supported, I plan to add support for V2 soon.

This pull request does depend on adding `#include "esp_smartconfig.h"` to the `bindings.h` file in the `esp-idf-sys` crate.

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -181,6 +181,20 @@ impl From<Newtype<&wifi_ap_record_t>> for AccessPointInfo {
     }
 }
 
+#[derive(Clone)]
+pub struct SmartConfigResult {
+    pub ssid: String,
+    pub password: Option<String>,
+    pub bssid: Option<[u8; 6]>,
+}
+
+enum SmartConfigStatus {
+    Stopped,
+    Requested,
+    Running,
+    Done(SmartConfigResult),
+}
+
 static mut TAKEN: EspMutex<bool> = EspMutex::new(false);
 
 struct Shared {
@@ -189,6 +203,7 @@ struct Shared {
 
     status: Status,
     operating: bool,
+    smart_config_status: SmartConfigStatus,
 
     sta_netif: Option<*mut esp_netif_t>,
     ap_netif: Option<*mut esp_netif_t>,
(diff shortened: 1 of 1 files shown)


ivmarkov:
I hope you do understand how insecure SmartConfig is. The fact that Texas Instruments invented it (if that's even true; there seems to exist prior art) does not make it secure. For the gory details, read [this](https://www.shielder.it/blog/2020/04/notsosmartconfig-broadcasting-wifi-credentials-over-the-air/). The fact that they hard-coded an AES key in the esptouch v2 app does not make it much more secure, as anyone can decompile and get the key.

I think protocols like wifi Easy Connect hold a much better promise, in that they'll likely be standardized by the CHIP specs and will soon land as a standard builtin feature on every smartphone.

Sure, they need a secondary communication channel like a BT connection or at least a QR code, but are secure by design. And the QR code trick is pretty neat and easy.

ivmarkov:
On a second thought, the protocol can be made secure if the esptouch app requires you to enter an aes key which is specific for the concrete iot device. The usability concerns can be mitigated by having the aes key of the device being printed and sticked on the device itself. Esptouch can then use the smartphone camera to read the qr code of the aes key before transmission.


Hmm, that might very well be what wifi easy connect does when using a qr code
 Need to read about it...

ivmarkov:
Ok so easy connect seems to use public action frames instead of udp packet length to send the (encrypted) ssid and pass to the iot device, but otherwise the approach is similar. They key to security is to use an aes key (or better yet - a public pki key) dictated by the iot device itself, which changes the security profile of the whole thing.

ivmarkov:
Wifi easy connect seems to be part of Android 10 already, strange that I cannot find it on my Samsung ui... https://source.android.com/devices/tech/connect/wifi-easy-connect

PR #37, opened 2022-01-18
[http/client] expose basic tls config

kcking:
exposing these tls options enabled me to get https _with_ server verification working like so

```rs
    let mut client = EspHttpClient::new(&EspHttpClientConfiguration {
        use_global_ca_store: true,

        crt_bundle_attach: Some(esp_idf_sys::esp_crt_bundle_attach),

        ..Default::default()
    })?;
```

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -10,6 +10,7 @@ use embedded_svc::http::client::*;
 use embedded_svc::http::*;
 use embedded_svc::io::{Read, Write};
 
+use esp_idf_sys::c_types::c_void;
 use esp_idf_sys::*;
 
 use crate::private::common::Newtype;
@@ -61,6 +62,10 @@ impl Default for FollowRedirectsPolicy {
 pub struct EspHttpClientConfiguration {
     pub buffer_size: Option<usize>,
     pub follow_redirects_policy: FollowRedirectsPolicy,
+
+    pub use_global_ca_store: bool,
+    #[cfg(not(esp_idf_version = "4.3"))]
+    pub crt_bundle_attach: Option<unsafe extern "C" fn(conf: *mut c_void) -> esp_err_t>,
 }
 
 #[allow(clippy::type_complexity)]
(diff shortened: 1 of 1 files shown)


ivmarkov:
The  `crt_bundle_attach` API you are using was introduced in the not-yet-released ESP-IDF V4.4+, while the minimum supported ESP-IDF version (and the official latest release) is still V4.3.X.

As you can check, it is not available here (V4.3.X): https://github.com/espressif/esp-idf/blob/release/v4.3/components/esp_http_client/include/esp_http_client.h#L108

**No prob with that**, but you need to conditionalize this code, as well as the `crt_bundle_attach` new member of struct `Configuration` to only compile with ESP-IDF versions later than 4.3.X.

The easiest way to do this to use `#[cfg(not(esp_idf_version = "4.3"))]` above your field, and also above the assignment statement further down. That would work (even though Rust does not support `<` and `>`), because versions earlier than 4.3.X are not supported as a compilation target by these crates anyway.

kcking:
Good catch :) just made that change

Also I should note that this PR https://github.com/esp-rs/esp-idf-sys/pull/57 has the added import in the `sys` crate. Looks like that PR suffers from the same version compatibility issue, I'll fix it as well.

src/http/client.rs
```diff
@@ -4,12 +4,13 @@ use alloc::collections::BTreeMap;
 use alloc::string::String;
 use alloc::string::ToString;
 
-use ::log::*;
+use log::*;
```
ivmarkov, line 7: Why?

ivmarkov (changes requested):
Only one superminor question for a changed import statement. The rest is perfect.

src/http/client.rs
kcking: Sorry, my auto cargo fmt in vscode must have changed that. Just fixed it!

ivmarkov:
Thanks!

PR #40, opened 2022-01-20
HTTP header keys are case-insensitive

### LINKED ISSUE #39: HTTP headers wrongly treated as case-sensitive by client
EspHttpResponse defines headers as `BTreeMap<String, String>`: https://github.com/esp-rs/esp-idf-svc/blob/a39c1199b1468d6b581d457cdc752bdf5a38ed1d/src/http/client.rs#L314-L317

If a server sends `Content-Type: something`, then `resp.content_type()` returns `None` because [embedded_svc asks for `content-type`](https://github.com/esp-rs/embedded-svc/blob/47c21851fb47141629057ceb2f834c4139eb666e/src/http.rs#L59-L61) and `EspHttpResponse` does a case-sensitive search: https://github.com/esp-rs/esp-idf-svc/blob/a39c1199b1468d6b581d457cdc752bdf5a38ed1d/src/http/client.rs#L337-L339

I *think* the fix is replacing the `String` key with something like [`Uncased`](https://docs.rs/uncased/0.9.6/uncased/struct.Uncased.html)

AlyoshaVasilieva:
Use the `uncased` crate when HTTP client is enabled to ensure this.

Fixes #39 in my brief test of my project, tested on an ESP32-C3.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ default = ["std"]
 std = ["alloc", "anyhow/std", "log/std", "esp-idf-sys/std", "esp-idf-hal/std", "embedded-svc/std"]
 alloc = ["cstr_core/alloc", "anyhow", "embedded-svc/alloc"]
 
-experimental = ["embedded-svc/experimental", "esp-idf-hal/experimental"]
+experimental = ["embedded-svc/experimental", "esp-idf-hal/experimental", "uncased"]
 
 [dependencies]
 enumset = { version = "1", default-features = false }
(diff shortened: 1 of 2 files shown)


ivmarkov:
This is great! I'll merge it even though CI fails, as it fails due to a [completely different reason](https://github.com/rust-lang/rust/pull/93152).

With that said, the experimental HTTPD server needs a similar fix [here](https://github.com/esp-rs/esp-idf-svc/blob/master/src/http/server.rs#L618). Are you up for applying the same fix there as well?

PR #41, opened 2022-01-21
Also use uncased for experimental HTTP server header names

AlyoshaVasilieva:
I left the TODO note about lowercasing because I haven't benchmarked any of this.

Tested briefly on an ESP32-C3.


ivmarkov:
Thanks!

PR #46, opened 2022-02-01
Push docs to github pages on publish

MabezDev:


ivmarkov:
Ideally also for embedded-svc

MabezDev:
Embedded SVC already has docs hosted on docs.rs because it doesn't depend on any idf crates: https://docs.rs/embedded-svc/0.16.5/embedded_svc/

PR #49, opened 2022-02-10
fix for mqtt keepalive

anichno:
Previous code would set keepalive interval to 1 second if a keepalive value was provided (or just not explicitly set to None).

One issue I've noticed with Mosquitto, if you explicitly set keep_alive_interval in MqttClientConfiguration to None, the esp32 will fail to connect. I'm not sure of the cause of this issue (I'm new to MQTT).


ivmarkov:
You are completely right - thanks for the fix! I'll merge it together with a bunch of other changes, as I'm about to release a new version.

PR #51, opened 2022-02-20
Get MAC address from EspNetinfo

galenguyer:


src/netif.rs
```diff
@@ -356,6 +356,16 @@ impl EspNetif {
         Cow::Owned(from_cstr(&netif_name).into_owned())
     }
 
+    pub fn get_mac(&self) -> Result<[u8; 6], EspError> {
```
ivmarkov, line 359: You need to fix formatting. Also, instead of matching on the result, please use the `?` Rust operator, which is more idiomatic.

ivmarkov (changes requested):

ivmarkov:
Thanks a lot!

PR #53, opened 2022-02-21
MQTT username/password

simonchatts:
Thanks for all your work on esp-rs! This trivial PR adds MQTT username / password support.

With this change, the [Ably Realtime](https://ably.com/pub-sub-messaging) cloud service seems to work well, using the [recommended configuration](https://ably.com/documentation/mqtt):

 - `mqtts://mqtt.ably.io:8883`
 - username and password from the API key
 - `MqttProtocolVersion::V3_1_1`
 - keepalive interval between 15 and 60 seconds


ivmarkov:
Thanks a lot!

PR #58, opened 2022-03-14
add esp-now support: safe api for EspNow and callback handling

bernii:
this PR adds support for [ESP-NOW](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_now.html) to the project.

I had to use a hack visible in `new()` that turns off the radio sleep `esp!(unsafe { esp_wifi_set_ps(0) })?;` - otherwise data stops being sent after a few seconds (this seems to be a known [bug](https://github.com/espressif/esp-idf/issues/7496)).

There is a related PR for `esp-idf-sys` and an issue this relates to: https://github.com/esp-rs/esp-idf-sys/pull/78 and https://github.com/esp-rs/esp-idf-sys/issues/70

### DIFF
--- a/src/espnow.rs
+++ b/src/espnow.rs
@@ -3,8 +3,10 @@ use log::info;
 use esp_idf_hal::mutex::Mutex;
 use esp_idf_sys::*;
 
-static RECV_CALLBACK: Mutex<Option<Box<dyn FnMut(&[u8], &[u8]) + Send>>> = Mutex::new(None);
-static SEND_CALLBACK: Mutex<Option<Box<dyn FnMut(&[u8], SendStatus) + Send>>> = Mutex::new(None);
+type Singleton<T> = Mutex<Option<Box<T>>>;
+
+static RECV_CALLBACK: Singleton<dyn FnMut(&[u8], &[u8]) + Send> = Mutex::new(None);
+static SEND_CALLBACK: Singleton<dyn FnMut(&[u8], SendStatus) + Send> = Mutex::new(None);
 
 pub static BROADCAST: [u8; 6] = [0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF];
 static TAKEN: Mutex<bool> = Mutex::new(false);
@@ -141,7 +143,7 @@ impl EspNowClient {
         Ok(())
     }
 
-    extern "C" fn send_callback(mac_addr: *const u8, status: esp_now_send_status_t) -> () {
+    extern "C" fn send_callback(mac_addr: *const u8, status: esp_now_send_status_t) {
         let c_mac = unsafe { std::slice::from_raw_parts(mac_addr, 6usize) };
 
         if let Some(ref mut callback) = *SEND_CALLBACK.lock() {
@@ -155,7 +157,7 @@ impl EspNowClient {
         mac_addr: *const u8,
         data: *const u8,
         data_len: c_types::c_int,
-    ) -> () {
+    ) {
         let c_mac = unsafe { std::slice::from_raw_parts(mac_addr, 6usize) };
         let c_data = unsafe { std::slice::from_raw_parts(data, data_len as usize) };
 
@@ -171,12 +173,12 @@ impl Drop for EspNowClient {
     fn drop(&mut self) {
         esp!(unsafe { esp_now_deinit() }).unwrap();
         let send_cb = &mut *SEND_CALLBACK.lock();
-        if let Some(_) = send_cb {
+        if send_cb.is_some() {
             *send_cb = None;
         }
 
         let recv_cb = &mut *RECV_CALLBACK.lock();
-        if let Some(_) = recv_cb {
+        if recv_cb.is_some() {
             *recv_cb = None;
         }
     }
--- a/src/espnow.rs
+++ b/src/espnow.rs
@@ -53,15 +53,6 @@ impl EspNowClient {
         Ok(Self(PrivateData))
     }
 
-    pub fn new_with_callback(
-        callback: impl for<'b, 'c> FnMut(&'b [u8], &'c [u8]) + 'static + Send,
-    ) -> Result<Self, EspError> {
-        let client = Self::new()?;
-
-        client.register_recv_cb(callback)?;
-        Ok(client)
-    }
-
     pub fn send(&self, peer_addr: [u8; 6], data: &[u8]) -> Result<(), EspError> {
         esp!(unsafe {
             esp_idf_sys::esp_now_send(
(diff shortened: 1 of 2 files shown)


ivmarkov:
This PR compiles fine even without the changes introduced with https://github.com/esp-rs/esp-idf-sys/pull/78
Moreover, I have some concerns with the changeset in `esp-idf-sys`. See there.

bernii:
I just tried compiling a sample project locally with `esp-idf-sys v0.30.6` instead of https://github.com/esp-rs/esp-idf-sys/pull/78
and def running to a failure:
```
$cargo build
...
   Compiling embedded-svc v0.17.3
error[E0432]: unresolved import `esp_idf_svc::espnow`
  --> src/main.rs:16:18
   |
16 | use esp_idf_svc::espnow::*;
   |                  ^^^^^^ could not find `espnow` in `esp_idf_svc`
```

I'm guessing that `svc` CI does not fail as cfg `esp_idf_comp_esp_espnow_enabled` is not enabled?

src/espnow.rs
```diff
@@ -0,0 +1,183 @@
...
+pub type PeerInfo = esp_now_peer_info_t;
+
+#[derive(Debug)]
+struct PrivateData;
+
+#[derive(Debug)]
+pub struct EspNowClient(PrivateData);
+
+impl EspNowClient {
+    pub fn new() -> Result<Self, EspError> {
```
ivmarkov, line 36: Since the ESP NOW client seems to be a singleton, you need to make sure that we are following the singleton pattern at least at runtime. (We don't have a way to statically enforce that (yet) in `esp-idf-svc`. We do - in `esp-idf-hal` with its "peripherals" thing.)

In other words, the first `EspNowClient::new` should succeed, and the second should fail. Please check [here](https://github.com/esp-rs/esp-idf-svc/blob/master/src/wifi.rs#L251) how this is enforced.

We also probably need to enforce that the `EspNowClient` is only created *after* the Wifi connection is initiated, by somehow passing a reference to the `EspWifi` instance into `EspNowClient`, however we can introduce this with a subsequent commit.

src/espnow.rs
```diff
@@ -0,0 +1,183 @@
+use log::info;
+
+use esp_idf_hal::mutex::Mutex;
+use esp_idf_sys::*;
+
+static RECV_CALLBACK: Mutex<Option<Box<dyn FnMut(&[u8], &[u8]) + Send>>> = Mutex::new(None);
```
ivmarkov, line 6: I think it would be better if this state is part of the `EspNowClient` structure. For one, that would allow you to get rid of the mutexes in case you decide to supply the callbacks during construction time, which I think is the simplest and cleanest approach.

src/espnow.rs
```diff
@@ -0,0 +1,183 @@
...
+static SEND_CALLBACK: Mutex<Option<Box<dyn FnMut(&[u8], SendStatus) + Send>>> = Mutex::new(None);
```
ivmarkov, line 7: Ditto.

src/espnow.rs
```diff
@@ -0,0 +1,183 @@
...
+        // to send any esp-now data after a few messages
+        // esp-idf bug report: https://github.com/espressif/esp-idf/issues/7496
+        esp!(unsafe { esp_wifi_set_ps(0) })?;
+
+        info!("Initializing ESP NOW");
+        esp!(unsafe { esp_now_init() })?;
+        Ok(Self(PrivateData))
+    }
+
+    pub fn new_with_callback(
```
ivmarkov, line 47: Is operation without a receiver callback even feasible? If not, perhaps we just need one constructor (this one)?
Also, given that this is a constructor for the receiver callback, it should be named perhaps `new_with_recv_callback`. 

Or - if you decide to also always provide a sender callback, perhaps: `new_with_callbacks`?

src/espnow.rs
```diff
@@ -0,0 +1,183 @@
...
+        Ok(())
+    }
+
+    pub fn get_verson() -> Result<u32, EspError> {
+        let mut version: u32 = 0;
+        esp!(unsafe { esp_now_get_version(&mut version as *mut u32) })?;
+        Ok(version)
+    }
+
+    pub fn register_recv_cb(
```
ivmarkov, line 114: I say, we either keep this API, or the `new_with_callback` one. Why would we need both? I'm inclined to a single `new_with_callbacks` API that just supplies both sender and receiver callbacks at construction time. This also allows us to no longer use mutexes to protect these.

ivmarkov (changes requested):
I think we can probably simplify this a bit by just enforcing that the callbacks are supplied by the user at construction time. Please let me know what you think, and whether we would be losing flexibility that way, and if yes, where.

src/espnow.rs
bernii: I think using espnow without callbacks still makes sense - someone might be just interested in broadcasting the data and not necessarily processing any incoming messages. Will update 👍

src/espnow.rs
bernii: yeah, making callbacks storage a part of `EspNowClient` struct is something I was aiming for originally but seems it's not possible in `rust` since I do not have a way to pass `context` (self) with the function signature (https://github.com/esp-rs/esp-idf-svc/pull/58/files#diff-fab8c30d89ae294965502f35d1d766710c2a9970398b2be902f0872948300ee9R154) that `esp_now_register_recv_cb` expects. AFAICT there's no user_ptr (`*mut c_void`) that I could leverage in order to avoid the global state variable. Let me know if you think there's a technique I could use to workaround this - I was not able to come up with a solution :(

src/espnow.rs
bernii: removed it since I think it's quite common to not register any callbacks

src/espnow.rs
bernii: Added a mutex to force singleton 👍

bernii:
@ivmarkov made some changes - pls let me know if you think there's a better technique around callbacks as I don't see a way to avoid globals there (since I can't pass the context given how the C function signature looks like)

ivmarkov:
> @ivmarkov made some changes - pls let me know if you think there's a better technique around callbacks as I don't see a way to avoid globals there (since I can't pass the context given how the C function signature looks like)

I agree. If you can fix the formatting errors with `cargo fmt`, I think that would be the final state of  the PR.

ivmarkov:
> sounds good! done ✅

Unfortunately now clippy fails. Can you look into that as well?

bernii:
should be good now 👍

ivmarkov:
> should be good now 👍

Seems you keep forgetting `cargo fmt`? Oh well...

bernii:
not sure why that one failed, `cargo fmt` is not complaining on my machine
```
esp-idf-svc berni$ cargo fmt
esp-idf-svc berni$
```

PR #59, opened 2022-03-25
Fix enums from `esp-idf-sys`.

reitermarkus:


ivmarkov:
See https://github.com/esp-rs/esp-idf-sys/pull/76#issuecomment-1725558061

PR #60, opened 2022-03-31
Make Clock configurable

elwerene:
~~Needs marker trait: https://github.com/esp-rs/esp-idf-hal/pull/56~~

### DIFF
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -89,7 +89,7 @@ impl RmiiClockConfig {
     fn eth_mac_clock_config(&self) -> eth_mac_clock_config_t {
         let rmii = match self {
             Self::Input(_) => eth_mac_clock_config_t__bindgen_ty_2 {
-                clock_mode: emac_rmii_clock_mode_t_EMAC_CLK_DEFAULT,
+                clock_mode: emac_rmii_clock_mode_t_EMAC_CLK_EXT_IN,
                 clock_gpio: emac_rmii_clock_gpio_t_EMAC_CLK_IN_GPIO,
             },
             Self::OutputGpio0(_) => eth_mac_clock_config_t__bindgen_ty_2 {
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -44,12 +44,7 @@ pub use nonblocking::*;
 
 #[cfg(all(esp32, esp_idf_eth_use_esp32_emac))]
 // TODO: #[derive(Debug)]
-pub struct RmiiEthPeripherals<
-    MDC,
-    MDIO,
-    CLK: gpio::RmiiClkPin = gpio::Gpio0<gpio::Unknown>,
-    RST: gpio::OutputPin = gpio::Gpio10<gpio::Unknown>,
-> {
+pub struct RmiiEthPeripherals<MDC, MDIO, RST: gpio::OutputPin = gpio::Gpio10<gpio::Unknown>> {
     pub rmii_rdx0: gpio::Gpio25<gpio::Unknown>,
     pub rmii_rdx1: gpio::Gpio26<gpio::Unknown>,
     pub rmii_crs_dv: gpio::Gpio27<gpio::Unknown>,
@@ -58,7 +53,7 @@ pub struct RmiiEthPeripherals<
     pub rmii_tx_en: gpio::Gpio21<gpio::Unknown>,
     pub rmii_txd0: gpio::Gpio19<gpio::Unknown>,
     pub rmii_mdio: MDIO,
-    pub rmii_ref_clk: CLK,
+    pub rmii_ref_clk_config: RmiiClockConfig,
     pub rst: Option<RST>,
 }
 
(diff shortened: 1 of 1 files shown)


src/eth.rs
```diff
@@ -3,7 +3,7 @@ use core::fmt::Debug;
 use core::ptr;
 use core::time::Duration;
 
-use ::log::*;
```
ivmarkov, line 6: Why?

src/eth.rs
```diff
@@ -232,6 +238,17 @@ where
 
         Ok((mac, phy))
     }
+
+    fn eth_mac_rmii_config() -> eth_mac_config_t {
+        let mut config = Self::eth_mac_default_config();
+
+        #[cfg(any(esp_idf_version = "4.4", esp_idf_version_major = "5"))]
```
ivmarkov, line 245: 1. Switch that to `#[cfg(not(esp_idf_version = "4.3"))]`. This way when version 6 comes out in 10 years, we are covered. :)
2. What does that mean for V4.3.2 (the minimum supported ESP-IDF version by us)? No clock configuration necessary? Also, what does that mean for `esp-idf-hal`? Is the `clock_config_t` even available on ESP-IDF V4.3.2?

ivmarkov (changes requested):

src/eth.rs
elwerene: I have rust format active on save and it changed it. I'll change it back.

src/eth.rs
elwerene: I have no idea :)

I copied that guard from https://github.com/esp-rs/esp-idf-svc/blob/master/src/eth.rs#L843

ivmarkov: Could it be that on V4.3.2 the only pin that is usable is gpio0?

elwerene: no, it seems to be the same in 4.3.2 as in 4.4: https://docs.espressif.com/projects/esp-idf/en/v4.3.2/esp32/api-reference/kconfig.html#config-eth-rmii-clk-mode

elwerene: So I really don't know why clock_config is not set on 4.3.2? Maybe it had a different format?

elwerene: So in 4.3 it was probably set up differently:

* 4.3 config: https://github.com/espressif/esp-idf/blob/release/v4.3/components/esp_eth/include/esp_eth_mac.h
* 4.4 config: https://github.com/espressif/esp-idf/blob/release/v4.4/components/esp_eth/include/esp_eth_mac.h

elwerene: I changed it to your suggestion

ivmarkov:
Hold on, hold on. Everything depends actually on the configuration options set via `sdkconfig.defaults.esp32`, so it makes the implementation very, very different (but still doable):
* First we need to expose `CONFIG_ETH_USE_ESP32_EMAC` and `CONFIG_ETH_RMII_CLK_MODE` and `CONFIG_ETH_RMII_CLK_OUT_GPIO` in `esp-idf-sys`. Since these are not boolean configs, they are not exposed by default, as non-bool configs need to be explicitly exposed (as there are a ton of them).
* Once we have those:
  * if `#[cfg(esp_idf_eth_use_esp32_emac = "ETH_PHY_INTERFACE_MII")]`, all things clock related should apparently be hidden (not compiled) and the Eth imnpl should **not** take any clk pin
 * if  `#[cfg(esp_idf_eth_use_esp32_emac = "ETH_PHY_INTERFACE_RMII")]`, then we have again multiple options:
   * if  `#[cfg(esp_idf_eth_rmii_clk_mode = "ETH_RMII_CLK_INPUT")]` => the impl **should** take a non-optional Gpio0 instance
   * if  `#[cfg(esp_idf_eth_rmii_clk_mode = "ETH_RMII_CLK_OUTPUT")]` => Either Gpio16 or Gpio17 is the Clk pin and the impl should take either gpio16 or gpio17 instance. Which pin should be taken by the impl  depends on which of the following two holds true: `#[cfg(esp_idf_eth_rmii_clk_out_gpio = "16")]` or `#[cfg(esp_idf_eth_rmii_clk_out_gpio = "17")]`.

I think you should get rid of your Clk marker trait actually, remove all changes to `esp-idf-hal` and then everything from above should be conditional compilation in `esp-idf-svc` only.

ivmarkov:
[This is the regex](https://github.com/esp-rs/esp-idf-sys/blob/master/build/build.rs#L55) in `esp-idf-sys` that needs to be extended to also include `CONFIG_ETH_USE_ESP32_EMAC`, `CONFIG_ETH_RMII_CLK_MODE` and `CONFIG_ETH_RMII_CLK_OUT_GPIO`.

elwerene:
What to do about https://github.com/esp-rs/esp-idf-svc/blob/master/src/eth.rs#L843 ? Here a variant is set, should we set the variant which matches the configuration option?

ivmarkov:
I'm reading the ESP-IDF source.... it is a bit... convoluted. So actually, **all** of the above compile-time configurations only matter if the user has programmatically set `EMAC_CLK_DEFAULT` for the clock. Only then whatever was specified in `sdkconfig.defaults` matters. However, you can - at run-time - instead of using `EMAC_CLK_DEFAULT` - select either `EMAC_CLK_OUT` (then you should provide either the Gpio16 or the Gpio17 instance or even the Gpio0 instance) or `EMAC_CLK_EXT_IN` (then you should provide the Gpio0 instance).

Which brings the question what shall we do? This is the first time I see such a thing with ESP-IDF - something can be configured compile time, but it only matters if the user selected "default configuration", or else user can override everything at run time.

Maybe this is a plan:
* For now, ditch these ESP-IDF `CONFIG_` things completely, don't change `esp-idf-sys` and require the user to provide configuration **at runtime**. Basically, how you had started, except that I'm no longer sure we need to touch `esp-idf-hal` at all? * (By the way, even though there is code for MII in ESP-IDF, the docu says ESP32 supports only RMII, so we should always unconditionally set and use RMII.)
* So we need an enum, perhaps, which is roughly modeling the ESP-IDF's `emac_rmii_clock_gpio_t` but also owns the pin:
```
pub enum RmiiClockConfiguration {
    In(Gpio0),
    ApplOutGpio0(Gpio0), // This according to ESP-IDF is for "testing" only
    ApplOutGpio16(Gpio16),
    ApplOutInvertedGpio17(Gpio17),
}
```
* Alternatively, you can continue with your original `Clk` trait, but this is touching two crates, so not sure this is so much better, in retrospective.
... and then you need to match on that enum and build the ESP-IDF clock structure depending on its values (and keep the passed pin owned).

All of that again only in `esp-idf-svc`.
... and regardless of the ESP-IDF version.

P.S. Sorry for the constant back-and-forth, but this stuff is more sophisticated than how it looked to me at first. I am starting to wonder if I ever got the board sent to me by Espressif which had Eth RMII port working actually... don't remember anymore.

elwerene:
I failed implementing it in sys anyways as I don't know how to set the configuration options so they are passed into transient dependencies. I tried setting an env in `.cargo/config` - did not work.

ivmarkov:
Let me look at it over the weekend. You might be right in your implementation that pins 16 and 17 in output mode are only supported on 4.4+ ([contrary to what the ESP-IDF documentation for V4.3.2 says](https://docs.espressif.com/projects/esp-idf/en/v4.3.2/esp32/api-reference/kconfig.html#config-eth-rmii-clk-mode), as I can't find any reference to these configuration parameters in the ESP-IDF V4.3.2 source code specifically.

src/eth.rs
```diff
@@ -72,6 +72,44 @@ pub enum RmiiEthChipset {
...
+    #[cfg(not(esp_idf_version = "4.3"))]
+    OutputInvertedGpio17(gpio::Gpio17<gpio::Unknown>),
+}
+
+#[cfg(all(esp32, esp_idf_eth_use_esp32_emac, not(esp_idf_version = "4.3")))]
+impl RmiiClockConfig {
+    fn eth_mac_clock_config(&self) -> eth_mac_clock_config_t {
+        let rmii = match self {
+            Self::Input(_) => eth_mac_clock_config_t__bindgen_ty_2 {
+                clock_mode: emac_rmii_clock_mode_t_EMAC_CLK_DEFAULT,
```
ivmarkov, line 92: I think this should **not** be `emac_rmii_clock_mode_t_EMAC_CLK_DEFAULT`, but `emac_rmii_clock_mode_t_EMAC_CLK_EXT_IN`. With `emac_rmii_clock_mode_t_EMAC_CLK_DEFAULT`, the configuration settings specified in `sdkconfig.defaults` will be used instead, and we do not want this, as agreed earlier.

Check [here](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_eth.html) for more info.

ivmarkov (changes requested):
I think we are there! If I'm not mistaken, you need to change just one constant (see the only feedback in this review), and then I'll merge.

src/eth.rs
elwerene: Good catch! Thanks :)

ivmarkov:
Many thanks!

PR #64, opened 2022-04-13
Fix ambiguous `log` import

gyscos:
Seems to be a leftover from https://github.com/esp-rs/esp-idf-svc/pull/5.

Otherwise it may conflict with the one imported from esp_idf_sys.

```
error[E0659]: `log` is ambiguous
 --> /esp-idf-svc-0.39.1/src/sntp.rs:4:5
  |
4 | use log::*;
  |     ^^^ ambiguous name
  |
  = note: ambiguous because of multiple potential import sources
  = note: `log` could refer to a crate passed with `--extern`
  = help: use `::log` to refer to this crate unambiguously
note: `log` could also refer to the struct imported here
 --> /esp-idf-svc-0.39.1/src/sntp.rs:8:5
  |
8 | use esp_idf_sys::*;
  |     ^^^^^^^^^^^^^^
  = help: use `self::log` to refer to this struct unambiguously

For more information about this error, try `rustc --explain E0659`.
```

### DIFF
--- a/src/espnow.rs
+++ b/src/espnow.rs
@@ -1,4 +1,4 @@
-use log::info;
+use ::log::info;
 
 use esp_idf_hal::mutex::Mutex;
 use esp_idf_sys::*;
--- a/src/sntp.rs
+++ b/src/sntp.rs
@@ -1,7 +1,7 @@
 extern crate alloc;
 use alloc::string::String;
 
-use log::*;
+use ::log::*;
 
 use esp_idf_hal::mutex;
 


gyscos:
Indeed, in my build, `esp_idf_sys` defines `log` in its `bindings.rs`.

For reference, my `sdkconfig.defaults` is:
```
# Rust often needs a bit of an extra main task stack size compared to C (the default is 3K)
CONFIG_ESP_MAIN_TASK_STACK_SIZE=7000

# Use this to set FreeRTOS kernel tick frequency to 1000 Hz (100 Hz by default).
# This allows to use 1 ms granuality for thread sleeps (10 ms by default).
#CONFIG_FREERTOS_HZ=1000

# Workaround for https://github.com/espressif/esp-idf/issues/7631
CONFIG_MBEDTLS_CERTIFICATE_BUNDLE=n
CONFIG_MBEDTLS_CERTIFICATE_BUNDLE_DEFAULT_FULL=n

CONFIG_BT_ENABLED=y
CONFIG_BT_NIMBLE_ENABLED=y
```

gyscos:
An alternative solution would be to _not_ `use esp_idf_sys::*;`. Given that its content can change depending on build configurations, it can indeed be risky to glob-import everything.

PR #66, opened 2022-04-24
Fix espnow.rs to build under no-std

Talljoe:
The recent addition of ESPNow had references to `std` and `Box` that weren't hidden behind cfg flags. This PR changes `std` to `core` and also add the `alloc` and network checks to `lib.rs`.


ivmarkov:
Thanks!

PR #70, opened 2022-05-01
Expose MAC address to wifi sta/ap

nim65s:
Hi,

I would need to identify a Wifi device by one of its MAC addresses, and couldn't find another way to do so. Did I miss something ?

Otherwise, this work for me.

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -336,6 +336,20 @@ impl EspWifi {
         Ok(wifi)
     }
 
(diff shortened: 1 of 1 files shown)


ivmarkov:
The following functions which are public and already available should allow you to do what you need to, without extending the API surface:
* `with_client_netif`
* `with_client_netif_mut`
* `with_router_netif`
* `with_router_netif_mut`

PR #72, opened 2022-05-04
Add support for buffer_size_tx config option in HTTP Client

callmephilip:
This adds support for `buffer_size_tx` config option in HTTP Client as discussed [here](https://github.com/ivmarkov/rust-esp32-std-demo/issues/79#issuecomment-1116457365)


MabezDev:
LGTM! Thanks :)

PR #75, opened 2022-05-06
Empty SSID for client config means not to connect

usbalbin:
This change makes it possible to set Mixed or Client mode without specifying a Wifi to connect to (as in passing Default::default()). Correct me if I am wrong but I believe this change would later make it possible to make a version of Wifi::scan() that does not need to restart the wifi.

```rust
let mut wifi = esp_wifi::EspWifi::new(...);

wifi.set_configuration(&wifi::Configuration::Client(
    Default::default()
));
```

or

```rust
let mut wifi = esp_wifi::EspWifi::new(...);
let my_ap_config = ...;

wifi.set_configuration(&wifi::Configuration::Mixed(
    Default::default(),
    my_ap_config,
));
```

Tested and seems to work on an esp32-wroom-32D
```
Chip type:         ESP32 (revision 1)
Crystal frequency: 40MHz
Flash size:        4MB
Features:          WiFi, BT, Dual Core, Coding Scheme None
```

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -401,8 +401,7 @@ impl EspWifi {
             Self::netif_unbind(shared.sta_netif.as_mut())?;
             shared.client_ip_conf = None;
             shared.sta_netif = None;
-        }
-        else {
+        } else {
             self.set_client_ip_conf(&conf.ip_conf)?;
         }
 
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -384,13 +384,27 @@ impl EspWifi {
     fn set_client_conf(&mut self, conf: &ClientConfiguration) -> Result<(), EspError> {
         info!("Setting STA configuration: {:?}", conf);
 
-        let mut wifi_config = wifi_config_t {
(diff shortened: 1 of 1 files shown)


src/wifi.rs
```diff
@@ -537,7 +551,7 @@ impl EspWifi {
             let mut shared = self.waitable.state.lock();
 
             shared.status = status.clone();
-            shared.operating = status.is_operating();
+            shared.operating = status.0.is_operating();
```
usbalbin, line 554: Do you have some better suggestion for handling this? From what I can tell we only care about `operating` in `Station`-parts of the code?

ivmarkov:
The new Wifi API in the `master` branch now has explicit `start`/`stop`/`connect` and `disconnect` methods. This means, we don't have to "hint" the Wifi driver anymore whether to connect/disconnect on `set_configuration`. User is explicitly in control.

The above changes were primarily (but not only) driven from the fact that protocols like `EspNow` do not even have the notion of "connecting".

PR #76, opened 2022-05-06
Allow scanning with optional restart

usbalbin:
Wifi::scan() will currently restart the Wifi clearing the config.
EspWifi::scan_with_optional_restart(do_restart: bool) will not do that when passing `false` as argument


ivmarkov:
This should now be fixed.

PR #77, opened 2022-05-07
Fix crash with zero-length MQTT messages

lamperez:
MQTT messages with null (zero length) payload crash the ESP32 with a guru meditation error when accessing the data. The solution just bypasses this access when `event_data_length` is zero, returning a zero-length Cow.


ivmarkov:
Thanks!

PR #79, opened 2022-05-19
Add `stats()` function for nvs_storage

psiphi75:
This maps to the `nvs_get_stats()` function from the ESP IDF.

### DIFF
--- a/src/nvs_storage.rs
+++ b/src/nvs_storage.rs
@@ -18,6 +18,8 @@ enum EspNvsRef {
     Nvs(Arc<EspNvs>),
 }
 
+pub type NvsStats = nvs_stats_t;
+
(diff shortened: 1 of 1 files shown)


ivmarkov:
Shall we instead expose the `nvs_handle_t` member, so that you can call any unsafe APIs on it, if you wish to?

psiphi75:
> Shall we instead expose the nvs_handle_t member, so that you can call any unsafe APIs on it, if you wish to?

That's a good question.  And you are probably the one to answer that.  I haven't looked how the rest of the codebase, but I'd think following the same pattern as the rest of the code makes sense.

ivmarkov:
I'm closing this because the `next` branch now has a `RawHandle` trait, which is implemented everywhere where it makes sense, including `EspNetif`. `next` to be released soon.

PR #82, opened 2022-05-25
Add Mdns

otakuto:
This adds support for mDNS.
Best regards for your review.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -42,6 +42,8 @@ pub mod http;
 #[cfg(all(feature = "std", esp_idf_comp_esp_http_server_enabled))]
 pub mod httpd;
 pub mod log;
+#[cfg(feature = "alloc")]
+pub mod mdns;
 #[cfg(all(
     feature = "alloc",
     esp_idf_comp_mqtt_enabled,
--- a/src/mdns.rs
+++ b/src/mdns.rs
@@ -1,0 +1,462 @@
+use core::time::Duration;
+
+extern crate alloc;
+use alloc::string::String;
+use alloc::vec::Vec;
+
+use embedded_svc::ipv4::Ipv4Addr;
+use std::net::{IpAddr, Ipv6Addr};
+
+use esp_idf_sys::*;
+
+use crate::private::cstr::{CStr, CString};
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Interface {
+    STA,
+    AP,
+    ETH,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Protocol {
+    V4,
+    V6,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Type {
+    A = MDNS_TYPE_A as _,
+    AAAA = MDNS_TYPE_AAAA as _,
+    ANY = MDNS_TYPE_ANY as _,
+    NSEC = MDNS_TYPE_NSEC as _,
+    OPT = MDNS_TYPE_OPT as _,
+    PTR = MDNS_TYPE_PTR as _,
+    SRV = MDNS_TYPE_SRV as _,
+    TXT = MDNS_TYPE_TXT as _,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct QueryResult {
+    pub instance_name: Option<String>,
+    pub hostname: Option<String>,
+    pub port: u16,
+    pub txt: Vec<(String, String)>,
+    pub addr: Vec<IpAddr>,
+    pub interface: Interface,
+    pub ip_protocol: Protocol,
+}
+
+impl From<mdns_result_t> for QueryResult {
+    #[allow(non_upper_case_globals)]
+    fn from(result: mdns_result_t) -> Self {
+        let instance_name = unsafe { result.instance_name.as_ref() }
+            .map(|p| unsafe { CStr::from_ptr(p) }.to_str().unwrap().to_string());
+        let hostname = unsafe { result.hostname.as_ref() }
+            .map(|p| unsafe { CStr::from_ptr(p) }.to_str().unwrap().to_string());
+        let port = result.port;
+
+        let mut txt = Vec::with_capacity(result.txt_count as usize);
+        for i in 0..result.txt_count as _ {
+            let p = unsafe { result.txt.offset(i) };
+            let key = unsafe { CStr::from_ptr((*p).key) }
+                .to_str()
+                .unwrap()
+                .to_string();
+            let value = unsafe { (*p).value.as_ref() }
+                .map_or(Default::default(), |p| {
+                    unsafe { CStr::from_ptr(p) }.to_str().unwrap()
+                })
+                .to_string();
+            txt.push((key, value));
+        }
+
+        let mut addr = Vec::new();
+        let mut p = result.addr;
+        while !p.is_null() {
+            let a = unsafe { (*p).addr };
+            let a = match a.type_ as _ {
+                ESP_IPADDR_TYPE_V4 => IpAddr::V4(from_esp_ip4_addr_t(unsafe { &a.u_addr.ip4 })),
+                ESP_IPADDR_TYPE_V6 => IpAddr::V6(from_esp_ip6_addr_t(unsafe { &a.u_addr.ip6 })),
+                _ => unreachable!(),
+            };
+            addr.push(a);
+
+            p = unsafe { (*p).next };
+        }
+
+        let interface = match result.tcpip_if {
+            mdns_if_internal_MDNS_IF_STA => Interface::STA,
+            mdns_if_internal_MDNS_IF_AP => Interface::AP,
+            mdns_if_internal_MDNS_IF_ETH => Interface::ETH,
+            _ => unreachable!(),
+        };
+
+        let ip_protocol = match result.ip_protocol {
+            mdns_ip_protocol_t_MDNS_IP_PROTOCOL_V4 => Protocol::V4,
+            mdns_ip_protocol_t_MDNS_IP_PROTOCOL_V6 => Protocol::V6,
+            _ => unreachable!(),
+        };
+
+        QueryResult {
+            instance_name,
+            hostname,
+            port,
+            txt,
+            addr,
+            interface,
+            ip_protocol,
+        }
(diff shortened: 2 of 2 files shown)


MabezDev:
@otakuto, Did you mean to close this?

src/lib.rs
```diff
@@ -42,6 +42,7 @@ pub mod httpd;
 #[cfg(feature = "alloc")]
 // TODO: Ideally should not need "alloc" (also for performance reasons)
 pub mod log;
+pub mod mdns;
```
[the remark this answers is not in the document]
otakuto: Thank you for your review.
Fixed it.

ivmarkov:
I'll go through this over the weekend.

reinismu:
@ivmarkov Sorry for the poke :) Have you managed to take a look at this?
Soon will start programming esp32 chip and would like to use Rust. Mdns is something I will need

ivmarkov:
> @ivmarkov Sorry for the poke :) Have you managed to take a look at this? Soon will start programming esp32 chip and would like to use Rust. Mdns is something I will need

I'm sorry - it completely escaped my head! Will follow up asap.

src/mdns.rs
```diff
@@ -0,0 +1,444 @@
...
+            mdns_if_internal_MDNS_IF_STA => Interface::STA,
+            mdns_if_internal_MDNS_IF_AP => Interface::AP,
+            mdns_if_internal_MDNS_IF_ETH => Interface::ETH,
+            _ => panic!(),
+        };
+
+        let ip_protocol = match result.ip_protocol {
+            mdns_ip_protocol_t_MDNS_IP_PROTOCOL_V4 => Protocol::V4,
+            mdns_ip_protocol_t_MDNS_IP_PROTOCOL_V6 => Protocol::V6,
+            _ => panic!(),
```
MabezDev, line 97:
```suggestion
            _ => unreachable!(),
```

MabezDev (commented):
Would you mind rebasing on master? It should rebase cleanly! I tried running this locally and ran into breaking changes in esp-idf-svc since this was submitted. Then `cargo-patch` failed me see https://github.com/itmettkeDE/cargo-patch/issues/30

src/mdns.rs
```diff
@@ -0,0 +1,444 @@
+use esp_idf_sys::*;
+
+extern crate alloc;
+use alloc::vec::Vec;
+
+use std::ffi::CStr;
```
ivmarkov, line 6: Please use the module of esp_idf_svc itself, which exposes either STD or non-std ffi support

src/mdns.rs
```diff
@@ -0,0 +1,444 @@
...
+use std::ffi::CString;
```
ivmarkov, line 7: Ditto.

src/mdns.rs
```diff
@@ -0,0 +1,444 @@
...
+
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
```
ivmarkov, line 9: This needs to use the types exposed from the `ipv4` module instead. This is a module of `embedded_svc`.

src/mdns.rs
```diff
@@ -0,0 +1,444 @@
...
+
+use std::time::Duration;
```
ivmarkov, line 11: `core::time::Duration`.

src/mdns.rs
```diff
@@ -0,0 +1,444 @@
...
+    ANY = MDNS_TYPE_ANY as _,
+    NSEC = MDNS_TYPE_NSEC as _,
+    OPT = MDNS_TYPE_OPT as _,
+    PTR = MDNS_TYPE_PTR as _,
+    SRV = MDNS_TYPE_SRV as _,
+    TXT = MDNS_TYPE_TXT as _,
+}
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct QueryResult {
```
ivmarkov, line 39: Please use the String type from the `alloc` module, not its std:: reexport. Otherwise this code will not compile under no_std. Or better yet: use `heapless::String` if these strings have upper bounds in terms of max length
 This way you won't depend for them on neither `alloc` nor `std` which would allow to move these types or a generalized version of those to `embedded-svc`, where no-std and no-alloc is a requirement.

ivmarkov (changes requested):

src/mdns.rs
```diff
@@ -0,0 +1,444 @@
...
+
+pub struct Mdns {}
+
+impl Drop for Mdns {
+    fn drop(&mut self) {
+        unsafe { mdns_free() };
+    }
+}
+
+impl Mdns {
```
ivmarkov, line 150: Let's name this `EspMdns`, as the other structs are named.

src/mdns.rs
```diff
@@ -0,0 +1,444 @@
...
+
+    pub fn query(
+        &mut self,
+        name: Option<&str>,
+        service_type: Option<&str>,
+        proto: Option<&str>,
+        mdns_type: Type,
+        timeout: Duration,
+        max_results: usize,
+    ) -> Result<Vec<QueryResult>, EspError> {
```
ivmarkov, line 326: Given that this method takes `max_result: usize`, let's instead of returning a `Vec<QueryResult>` take a `&mut [QueryResult]` where the method should fill it in. And the method should return `Result<usize, EspError>`, where the `usize` is the number of elements in the slice which were filled in.

src/mdns.rs
```diff
@@ -0,0 +1,444 @@
...
+        Ok(from_esp_ip6_addr_t(&addr))
+    }
+
+    pub fn query_txt(
+        &mut self,
+        instance_name: impl AsRef<str>,
+        service_type: impl AsRef<str>,
+        proto: impl AsRef<str>,
+        timeout: Duration,
+    ) -> Result<Vec<QueryResult>, EspError> {
```
ivmarkov, line 380: Can we pass a `&mut []` slice here as well?

src/mdns.rs
```diff
@@ -0,0 +1,444 @@
...
+                service_type.as_ptr(),
+                proto.as_ptr(),
+                timeout.as_millis() as _,
+                &mut result,
+            )
+        })?;
+        Ok(into_query_results(unsafe { Box::from_raw(result) }))
+    }
+
+    pub fn query_ptr(
```
ivmarkov, line 422: Ditto?

ivmarkov (changes requested):

src/mdns.rs
otakuto: We fixed above.
And rebased on master.

src/mdns.rs
otakuto: We Fixed it.

src/mdns.rs
otakuto: Fixed.

src/mdns.rs
otakuto: Fixed.

reinismu (commented):
Take my comments with a grain of salt as I'm still very green at embedded space.

Another thing I noticed that wifi module does a lot of logging that helps. Maybe could add few logs here as well?

Code that I used
```
pub fn init_mdns(
    _netif_stack: Arc<EspNetifStack>,
    _sys_loop_stack: Arc<EspSysLoopStack>,
    _default_nvs: Arc<EspDefaultNvs>,
) -> Result<Box<EspMdns>> {
    let mut mdns = Box::new(mdns::EspMdns::new()?);

    mdns.hostname_set("ESP32-S3")?;
    mdns.service_add(Some("ESP32-S3-Instance"),"_http", "_tcp", 80, &[])?;

    Ok(mdns)
}
```

src/mdns.rs
```diff
@@ -0,0 +1,447 @@
...
+pub struct EspMdns {}
+
+impl Drop for EspMdns {
+    fn drop(&mut self) {
+        unsafe { mdns_free() };
+    }
+}
+
+impl EspMdns {
+    pub fn new() -> Result<Self, EspError> {
```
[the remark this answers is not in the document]
ivmarkov: Thats a good idea, but has to be checked if all three are necessary, or only a subset of these.

src/mdns.rs
otakuto: Can we use heapless::Vec?

src/mdns.rs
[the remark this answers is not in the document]
reinismu: From their examples it seems that need all 3 https://github.com/espressif/esp-idf/blob/cb1abdf33e359c5bdb42a71bc13198e1e045c93f/examples/protocols/http_server/restful_server/main/esp_rest_main.c#L131
There was another one as well

src/mdns.rs
otakuto: Which IpAddr and Ipv6Addrshould We use?
Does embedded_svc plan to export that?

src/mdns.rs
ivmarkov: Sure, though i kind of don't see the benefit? If you go for heapless Vec, `max_result` should become a const generic which defines the size of the heapless vec.

src/mdns.rs
ivmarkov: You are right, embedded-svc does not yet expose these. Keep them as they are, ill fix post-commit

otakuto: We fixed it.

src/mdns.rs
otakuto: We used alloc module.

src/mdns.rs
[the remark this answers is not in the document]
otakuto: The Esp* interface has been changed.
What should I do?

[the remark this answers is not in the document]
ivmarkov: Leave it without taking any of these as the original design was.

src/mdns.rs
otakuto: that's true.

otakuto: We fixed it.

otakuto:
I fixed everything for now.
Review please.

ivmarkov:
Clippy in CI fails. Can you fix the clippy lints?

ivmarkov:
Actually a lot of the clippy lints are caused by existing code. Let me fix these.

ivmarkov:
Thanks! I've replaced your `new()` with a mutex-protected singleton `take()` constructor to guard against double-initialization of mdns. I've also re-formatted the whitespacing in the code a bit, and I'll fix a clippy error.

PR #92, opened 2022-07-11
log: implementation with no allocations

tcbennun:
As noted in TODO, use esp_log_level_get to decide whether to log, then use println!() to directly write message, including no-alloc write of record.args().

### DIFF
--- a/src/log.rs
+++ b/src/log.rs
@@ -117,10 +117,10 @@ impl EspLogger {
 
     fn should_log(record: &Record) -> bool {
         let level = Newtype::<esp_log_level_t>::from(record.level()).0;
-        let min_level = unsafe {
+        let max_level = unsafe {
             esp_log_level_get(b"rust-logging\0" as *const u8 as *const _) // TODO: use record target?
         };
-        level <= min_level
+        level <= max_level
     }
 }
 
--- a/src/log.rs
+++ b/src/log.rs
@@ -1,9 +1,26 @@
+use core::fmt::Write;
+
 use ::log::{Level, LevelFilter, Metadata, Record};
 
 use esp_idf_sys::*;
 
 use crate::private::common::*;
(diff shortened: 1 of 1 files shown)


ivmarkov:
The one problem I just realized is that println! drags in a hard dependency in STD. We have to think a bit how to do that in a no-std compatible way.

ivmarkov:
We probably need to implement the `core::fmt::Write` trait for the stdout file descriptor of ESP-IDF using unsafe code and from there we can use the `write!` macro instead.

tcbennun:
First pass without handling errors:

```
use core::fmt::Write;

struct EspStdout;

impl core::fmt::Write for EspStdout {
    fn write_str(&mut self, s: &str) -> core::fmt::Result {
        let stdout = unsafe { __getreent().as_mut() }.unwrap()._stdout;
        let slice = s.as_bytes();
        unsafe {
            fwrite(
                slice.as_ptr() as *const _,
                1,
                slice.len() as u32,
                stdout
            );
        }
        Ok(())
    }
}
```

then in `log()`,

```
                write!(EspStdout,
                    "\x1b[0;{}m{} ({}) {}: {}\x1b[0m\n",
                    color,
                    Self::get_marker(record.metadata().level()),
                    unsafe { esp_log_timestamp() },
                    record.metadata().target(),
                    record.args(),
                )
                .unwrap();
```

Tested with `rust-esp32-std-demo`, working fine.

Do you know another way to grab `stdout`? `getreent` is inexpensive but a bit messy.

ivmarkov:
I think that's OK actually!

tcbennun:
I've only tested on ESP32-C3 - if anyone can test on other platforms that would be good (just to be sure!)

ivmarkov:
@tcbennun Can you fix the `fmt` error so that I can merge?

PR #93, opened 2022-07-14
add with_handle

elwerene:
Add possibility to access the ethernet handle after initializing the driver. I needed this because I had to set a phy register.

### DIFF
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -613,7 +613,7 @@ impl<P> EspEth<P> {
         Ok(eth)
     }
 
-    pub fn with_handle<F, T>(&self, f: F) -> T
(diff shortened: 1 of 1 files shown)


src/eth.rs
```diff
@@ -613,6 +613,13 @@ impl<P> EspEth<P> {
         Ok(eth)
     }
 
+    pub fn with_handle<F, T>(&self, f: F) -> T
```
MabezDev, line 616: Semantically this should be `&mut self`, not `&self` if it can be used to modify the ethmac

MabezDev (changes requested):

MabezDev (approved):
LGTM, thanks!

PR #107, opened 2022-07-22
#105 Add option to set a mac for an interface and use the default if none is set

### LINKED ISSUE #105: Question: How to obtain Mac Address of Wifi Adapter?
This is probably an easy one:

I'm trying to write a custom WifiMgr which handles wifi reconnects and creating an Wifi AP if no WiFI is available.
To avoid SSID duplicates and keep them identifyable, I want to use the last 3 bytes of the mac as hex in the SSID.

**What I want to do:**

Get the default Mac Address of the Wifi Interface before starting a connection/AP.

**What I've tried:**

I've tried some different variations of this already:
```rust
let netif = EspNetif::new(
    Arc::new(EspNetifStack::new()?),
    &InterfaceConfiguration {
        interface_stack: InterfaceStack::Sta,
        ..Default::default()
    },
)?;
println!("Netif: {:?}", netif.get_mac());
```

**What I get:**

But this always prints the mac as only zeros.

**I also tried this:**

```rust
let mut wifi = EspWifi::new(
  Arc::new(EspNetifStack::new()?),
  Arc::new(EspSysLoopStack::new()?),
  Arc::new(EspDefaultNvs::new()?)
)?;

let status = wifi.get_status();
let wifi_config = wifi.get_configuration();
println!("Status: {status:?}, wifi config: {wifi_config:?}");
```

**Used Hardware:**

[M5StampC3](https://docs.m5stack.com/en/core/stamp_c3) - ESP32-C3 based board

I'm probably close, but the docs don't really help me out and I feel a little lost.

Snapstromegon:
This PR sets the default mac for any initialized mac (if it can be done reasonably) and allows to set a custom max during creation (or at a later point in time).

That way the `get_mac` method doesn't require an unsafe call to [esp_netif_set_mac](https://esp-rs.github.io/esp-idf-svc/esp_idf_sys/fn.esp_netif_set_mac.html) to initially set the mac address.

This is a breaking change, as it modifies the members of the InterfaceConfiguration struct.

closes #105

### DIFF
--- a/src/netif.rs
+++ b/src/netif.rs
@@ -403,7 +403,7 @@ impl EspNetif {
     }
 
     pub fn set_mac(&mut self, mac: &[u8; 6]) -> Result<(), EspError> {
-        esp!(unsafe { esp_netif_set_mac(self.1, mac.clone().as_mut_ptr() as *mut _) })?;
+        esp!(unsafe { esp_netif_set_mac(self.1, mac.as_ptr() as *mut _) })?;
         Ok(())
     }
 
--- a/src/netif.rs
+++ b/src/netif.rs
@@ -60,6 +60,7 @@ pub struct InterfaceConfiguration {
     pub route_priority: u32,
     pub ip_configuration: InterfaceIpConfiguration,
     pub interface_stack: InterfaceStack,
+    pub custom_mac: Option<[u8; 6]>,
 }
 
 impl Default for InterfaceConfiguration {
@@ -76,6 +77,7 @@ impl InterfaceConfiguration {
             route_priority: 60,
             ip_configuration: InterfaceIpConfiguration::Client(Default::default()),
             interface_stack: InterfaceStack::Eth,
+            custom_mac: None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/netif.rs
```diff
@@ -369,6 +402,11 @@ impl EspNetif {
         Ok(mac)
     }
 
+    pub fn set_mac(&mut self, mac: &[u8; 6]) -> Result<(), EspError> {
+        esp!(unsafe { esp_netif_set_mac(self.1, mac.clone().as_mut_ptr() as *mut _) })?;
```
ivmarkov, line 406: Nit: I would assume `clone()` workarounds the fact that `esp_netif_set_mac` expects a non-const C pointer to the MAC.

I don't think the above is on purpose. In my experience, the ESP-IDF APIs don't always follow strictly C const semantics, and sometimes take non-const pointers for stuff which is not modified by these.

Or to cut the long story short `mac.as_ptr() as *const _ as *mut _` would also do, though this is such a micro-optimization that who cares after all. :)

ivmarkov (changes requested):
Changes look OK to me, I just have a minor nit you might want to review.

src/netif.rs
Snapstromegon: Thanks for the comment, I will adapt that.
The clone was mainly, because the mac is a "normal" reference and therefore `.as_mut_ptr()` was not available and it was just a quick fix for that. I will fix that.

PR #109, opened 2022-07-23
Publish method for filtering ip_events from outside

### LINKED ISSUE #106: Feature Request / RFC: Allow subscribing to wifi and ip event seperately
At the moment the callback for [EspWifi::subscribe](https://esp-rs.github.io/esp-idf-svc/esp_idf_svc/wifi/struct.EspWifi.html#method.subscribe) does not pass the event into the callback, because the Subscription is to a combination of WifiEvent and IpEvent.

For some applications it would be useful to get the event instead of just an info that something happened.

To achieve this, I see two options, which are possible to implement together and **I'd be willing to create a PR for this**.

**Option 1: Two `subscribe_*` methods**

This options contains adding two new methods to the [esp_idf_svc::wifi::EspWifi](https://esp-rs.github.io/esp-idf-svc/esp_idf_svc/wifi/struct.EspWifi.html#) struct: 

1. `subscribe_wifi_events(&mut self, callback: impl for<'a> FnMut(&'a WifiEvent) + Send + 'static) -> Result<Self::Subscription, Self::Error>)`
2. `subscribe_ip_events(&mut self, callback: impl for<'a> FnMut(&'a IpEvent) + Send + 'static) -> Result<Self::Subscription, Self::Error>)`

This would allow users to clearly define which events they want to listen on and also provide a way to actually receive the triggering events.

This approach can be implemented without making a breaking change.

**Option 2: Changing the `subscribe` signature**

This option would change the current method signature in a way, that allows returning the events. This could be achieved like this:

1. Change the callback param to `(Option<&WifiEvent>, Option<&IpEvent>)` and fill the corresponding
2. Add a new type `enum EspWifiEvent { Wifi(&WifiEvent), Ip(&IpEvent) }` (simplified) and using that as the callback argument

This would be a breaking change to the API.

**Currently possible options:**

ATM you could of course subscribe to the corresponding event bus yourself and reimplement the logic inside of esp-idf-svc, but I think this is neither ergonomic, nor a good pattern.

Snapstromegon:
Fixes #106 and allows third parties to test wether or not an IpEvent is related to an EspWifi instance.
This allows using the eventloop to get all wifi events and to filter for relevant events.

### DIFF
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -936,6 +936,14 @@ impl<P> EspEth<P> {
             ..Default::default()
         }
     }
+
+    /// Filter wether or not an IpEvent is related to this [`EspEth`] instance.
+    ///
+    /// As an example this can be used to check when the Ip changed.
+    pub fn is_ip_event_for_self(&self, event: &IpEvent) -> bool {
+        let shared = self.waitable.state.lock();
+        shared.is_our_ip_event(event)
+    }
 }
 
 impl<P> Eth for EspEth<P> {
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -782,6 +782,36 @@ impl EspWifi {
             ClientStatus::Started(ClientConnectionStatus::Disconnected)
         })
     }
+
+    /// Filter wether or not an IpEvent is related to this [`EspWifi`] instance.
+    ///
+    /// As an example this can be used to check when the Wifi Ip changed.
+    /// ```rust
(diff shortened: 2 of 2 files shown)


ivmarkov:
Sorry for the delay here, will try to look at it over the weekend.

Snapstromegon:
No rush, thanks a lot for your work!

MabezDev (approved):
Looks good to me.

PR #112, opened 2022-08-04
Tag the new release on Publish action

N3xed:


ivmarkov:
@N3xed - I assume you'll do something similar for `embedded-svc`, `esp-idf-hal` and `esp-idf-template`? Would be greatly appreciated!

N3xed:
> @N3xed - I assume you'll do something similar for `embedded-svc`, `esp-idf-hal` and `esp-idf-template`? Would be greatly appreciated!

Yup, done. I've also pushed tags that I've scraped from the successful `Publish` runs.

PR #124, opened 2022-08-30
Add mutual TLS for EspHttpClient

AnthonyGrondin:
## Description
Add `client_cert_pem` and `client_key_pem` attributes to `EspHttpClient` for mutual TLS, as defined in https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/esp_http_client.html#_CPPv424esp_http_client_config_t

This is currently a PoC. The feature still has a bug to resolve, described below.

## Example
You can include certificates with the `include_str!()` macro
```rust
const client_cert: &str = include_str!("certs/client_cert.pem");
const client_key: &str = include_str!("certs/client_key.pem");
```
and specify them in `EspHttpClientConfiguration`
```rust
let mut client = EspHttpClient::new(&EspHttpClientConfiguration {
        crt_bundle_attach: Some(esp_idf_sys::esp_crt_bundle_attach),
        client_cert_pem: Some(client_cert),
        client_key_pem: Some(client_key),
        ..Default::default()
    })?;
```

## ~~Issue~~ (Fixed)

~~There's currently an issue with the underlying implementation~~, 
https://github.com/esp-rs/esp-idf-svc/blob/73b3e53e1f7fd87b4abcab68089509afbb5c7166/src/http/client.rs#L114-L131
~~We specify the pointers to the certificates and pass them to `native_config`, but the value referenced by the pointers is dropped at the end of the scope. Similar as the problem detailed here:~~
https://github.com/esp-rs/esp-idf-svc/blob/554cad178e079b75e837885dffcf71a833f53107/src/http/server.rs#L198-L208

~~Contrary of the Https Server, the values referenced by the pointers need to be valid when doing a request, even after the initialization has ended.~~


AnthonyGrondin:
Each mutual TLS requests seem to take a couple of seconds, I believe this is because we don't take advantage of the TLS sessions feature.

https://github.com/espressif/esp-idf/issues/8852

ivmarkov:
@AnthonyGrondin Sorry for the huge delay here, but as you've probably noticed, I was overly busy with finishing the `next` branch and merging it into `master`. Now that this is done, we can follow up on all the outstanding merge requests. I'm absolutely for merging this now, but can you rebase it on top of latest `master` and fix the merge conflicts?

ivmarkov:
@AnthonyGrondin The PR is unmergeable due to conflicts. Can you resolve these?

ivmarkov:
> @AnthonyGrondin The PR is unmergeable due to conflicts. Can you resolve these?

^^^ Disregard.

PR #125, opened 2022-08-31
Fix RawStorage::get_raw for EspNvsStorage

simmsb:
This would panic at `buf.copy_from_slice(&array[..min(buf.len(), len as usize)]);` if `buf` was longer than `len`


ivmarkov:
Fixed by https://github.com/esp-rs/esp-idf-svc/commit/b33d7951d167aabf9ae836a53491b362b1fc4384

PR #130, opened 2022-09-16
Fix build for IDF 'release/5.0'

usbalbin:
mqtt is disabled for IDF 5.0 for now


usbalbin:
This seems to work with https://github.com/esp-rs/esp-idf-sys/pull/133 and IDF 'release/5.0'

ivmarkov:
The `netif` changes are merged in, but obviously I could not just remove mqtt support when compiling against ESP IDF 5, so this, alongside other fixes were implemented instead, thanks.

usbalbin:
Makes sense 👍

PR #131, opened 2022-09-17
fix: show signal strength as signed integer

mfiumara:
This commit fixes displaying of the received signal strength, as these are signed values (dBs).

This depends on https://github.com/esp-rs/embedded-svc/pull/28.


ivmarkov:
Thanks!

PR #147, opened 2022-10-11
Fix #146 by providing std::ptr::null if payload is empty

### LINKED ISSUE #146: EspMqttClient::publish() with empty payload is causing a Guru Meditation Errors due to loading from 0x1
The function EspMqttClient::publish() accepts payload as `&[u8]` and uses it to feed the C function esp_mqtt_client_publish().
If an empty payload is to be sent ([] with length 0) the len is 0, so according to the documentation (and source code) the actual length is calculated from the payload string (with strlen()).

https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/mqtt.html#_CPPv423esp_mqtt_client_publish24esp_mqtt_client_handle_tPKcPKciii

> data – payload string (set to NULL, sending empty payload message)
> len – data length, if set to 0, length is calculated from payload string 

https://github.com/espressif/esp-mqtt/blob/master/mqtt_client.c#L1975-L1977

The issue here is, that in rust a pointer to the zero sized [] is actually 0x1, so the strlen() used in the c function is causing a load from 0x1...

https://github.com/esp-rs/esp-idf-svc/blob/6e8463074bbc0a4ab72b8dc484b3d1da646b5155/src/mqtt/client.rs#L453-L472

A fix could be to handle the payload.len() == 0 case to provide a NULL (ptr::null()) as payload.

mike1703:
this is fixing #146 the issue that a pointer to [] is actually 0x1 (not 0x0).

I fixed not only publish() but also enqueue() as this was also not working (but failed silent for me).

There is a new commit https://github.com/espressif/esp-idf/issues/9719 which fixes the enqueue() function in esp-idf which would cause an actual issue in the rust enqueue() function as well.

### DIFF
--- a/src/mqtt/client.rs
+++ b/src/mqtt/client.rs
@@ -459,11 +459,16 @@ impl<S> EspMqttClient<S> {
     ) -> Result<client::MessageId, EspError> {
         let c_topic = CString::new(topic).unwrap();
 
+        let payload_ptr = match payload.len() {
+            0 => core::ptr::null(),
+            _ => payload.as_ptr(),
+        };
+
         Self::check(unsafe {
             esp_mqtt_client_publish(
                 self.raw_client,
                 c_topic.as_ptr(),
-                payload.as_ptr() as _,
+                payload_ptr as _,
                 payload.len() as _,
                 qos as _,
                 retain as _,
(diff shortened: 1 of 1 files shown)


src/mqtt/client.rs
```diff
@@ -459,11 +459,16 @@ impl<S> EspMqttClient<S> {
     ) -> Result<client::MessageId, EspError> {
         let c_topic = CString::new(topic).unwrap();
 
+        let payload_ptr = match payload.len() {
+            0 => std::ptr::null(),
```
ivmarkov, line 463: Use core, not std.

src/mqtt/client.rs
```diff
@@ -480,11 +485,16 @@ impl<S> EspMqttClient<S> {
...
+            _ => payload.as_ptr(),
```
ivmarkov, line 490: Ditto.

ivmarkov (changes requested):

mike1703:
changed the commit to use core directly and not the std reexport anymore

ivmarkov:
Thanks!

PR #148, opened 2022-10-15
make `IpEvent` `Send`

zRedShift:
The `*mut esp_netif_t` makes `IpEvent` not implement `Send`, so it's either `unsafe impl Send for IpEvent {}` as I've done here, or to wrap the pointer in `RawHandleImpl`, and have it implement `RawHandle` just like `EspNetif`, I'm ok with that too. @ivmarkov, what do you think?


ivmarkov:
Let's go with your strategy for now. I did the same for `EthEvent`.

PR #150, opened 2022-10-18
fix: provide empty str for invalid utf-8 string

yplam:
If there are wifi access points with invalid utf-8 name, our firmware panic with message :

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 0, error_len: Some(1) }', /home/.cargo/registry/src/github.com-1ecc6299db9ec823/esp-idf-svc-0.42.3/src/private/cstr.rs:35:10

Try to fix it with this commit. Thank you.


MabezDev:
Hmm, I don't think this fix is the right way to go about this. The 802.11 spec says the SSID can either be valid UTF-8 _or_ unspecified data. Therefore I don't think we should be treating SSID as a string at all. Instead, we should treat it as a raw set of bytes and add helper methods for automatically converting strings into the array format.

ivmarkov:
> Hmm, I don't think this fix is the right way to go about this. The 802.11 spec says the SSID can either be valid UTF-8 _or_ unspecified data. Therefore I don't think we should be treating SSID as a string at all. Instead, we should treat it as a raw set of bytes and add helper methods for automatically converting strings into the array format.

I agree.

PR #152, opened 2022-10-29
Make EspOtaUpdate Send

wzhd:
This way when using async/await to receive ota data, the future can be `Send`, which some executors expect.


wzhd:
added a space for format

PR #153, opened 2022-10-30
Add timeout config to http client

kevinresol:
Expose timeout config for http client.

I am not very sure if it is better to use `Duration` type or just expose the config as-is, please advice.

Also I am not very sure if I should target master branch instead, but there seems to be a huge diff with conflicts if do that.


ivmarkov:
1. Please target the master branch. No extensions to the release branch are planned. Given that your change is a two- liner, you should just do it on top of master
2. Use `core::time::Duration` instead.

ivmarkov (changes requested):
As per my other comment.

PR #154, opened 2022-10-30
Fix Mdns, HTTPS server for `esp-idf` v5.0+, option for httpd wildcard matcher (available in v4 as well)

zRedShift:


src/mdns.rs
```diff
@@ -88,6 +88,17 @@ impl From<mdns_result_t> for QueryResult {
...
         }
 
+        #[cfg(esp_idf_version_major = "5")]
+        let interface = match unsafe { CStr::from_ptr(esp_netif_get_desc(result.esp_netif)) }
+            .to_bytes_with_nul()
+        {
+            b"sta\0" => Interface::STA,
+            b"ap\0" => Interface::AP,
+            b"eth\0" => Interface::ETH,
+            _ => unimplemented!(),
```
zRedShift, line 98: This is not very elegant, as the user can give custom descriptions to `esp_netif`, and also there are multiple new `netif` types that aren't covered by these.

src/mdns.rs
```diff
@@ -88,6 +88,17 @@ impl From<mdns_result_t> for QueryResult {
             p = unsafe { (*p).next };
         }
 
+        #[cfg(esp_idf_version_major = "5")]
```
zRedShift, line 91: `tcpip_if` -> `esp_netif` in v5.0+
see: https://github.com/espressif/esp-protocols/commit/ddc58e8220f8b4d8829cfe47cfecdf071e92c845#diff-b1f0aa635ad5c4ccef42c50cda8b8e441e9fc5b324d1ea3f7703af1ded32cb75L86-R79
maybe a better approach would be to get rid of the `Interface` enum

src/lib.rs
```diff
@@ -42,7 +42,7 @@ pub mod http;
 #[cfg(all(feature = "std", esp_idf_comp_esp_http_server_enabled))]
 pub mod httpd;
 pub mod log;
-#[cfg(feature = "alloc")]
+#[cfg(all(feature = "alloc", esp_idf_comp_mdns_enabled))]
```
zRedShift, line 45: let's leave the work of making sure that the `mdns` component is fetched to `esp-idf-sys`, no need to check `espidf` version here

src/http/server.rs
```diff
@@ -559,9 +578,7 @@ impl<'a> EspHttpConnection<'a> {
     pub fn uri(&self) -> &str {
         self.assert_request();
 
-        let c_uri = unsafe {
-            CStr::from_bytes_with_nul_unchecked(mem::transmute(self.request.0.uri.as_slice()))
-        };
+        let c_uri = unsafe { CStr::from_ptr(self.request.0.uri.as_ptr()) };
```
zRedShift, line 581: This was a bug, it violated the `CStr` safety contract, and ended up as a 512 byte long mostly-zeroed `&str`

src/http/server.rs
```diff
@@ -159,15 +161,23 @@ impl From<&Configuration> for Newtype<httpd_ssl_config_t> {
...
             cacert_pem: ptr::null(),
             cacert_len: 0,
             prvtkey_pem: ptr::null(),
             prvtkey_len: 0,
+            #[cfg(esp_idf_version_major = "4")]
             client_verify_cert_pem: ptr::null(),
+            #[cfg(esp_idf_version_major = "4")]
             client_verify_cert_len: 0,
+            #[cfg(esp_idf_version_major = "5")]
+            servercert: ptr::null(),
```
zRedShift, line 178: renamed in v5.0+
see: https://github.com/espressif/esp-idf/commit/94056fd4a547b63521f9b47fd313b8d1b8a1c7c5#diff-4ac40d704acb6c3ed7b266a6fe23a62209bd770ba94b340d5343381e45980e34L53-L68

src/http/server.rs
```diff
@@ -159,15 +161,23 @@ impl From<&Configuration> for Newtype<httpd_ssl_config_t> {
         Self(httpd_ssl_config_t {
             httpd: http_config.0,
             session_tickets: false,
+            #[cfg(esp_idf_version_major = "5")]
```
ivmarkov, line 164: Please do `not(esp_idf_version_major = "4")`. This way we are safe for ESP-IDF 6 :)

src/http/server.rs
```diff
@@ -159,15 +161,23 @@ impl From<&Configuration> for Newtype<httpd_ssl_config_t> {
...
             transport_mode,
             cacert_pem: ptr::null(),
             cacert_len: 0,
             prvtkey_pem: ptr::null(),
             prvtkey_len: 0,
+            #[cfg(esp_idf_version_major = "4")]
             client_verify_cert_pem: ptr::null(),
+            #[cfg(esp_idf_version_major = "4")]
             client_verify_cert_len: 0,
+            #[cfg(esp_idf_version_major = "5")]
```
ivmarkov, line 177: Ditto.

src/http/server.rs
```diff
@@ -159,15 +161,23 @@ impl From<&Configuration> for Newtype<httpd_ssl_config_t> {
...
+            servercert: ptr::null(),
+            #[cfg(esp_idf_version_major = "5")]
```
ivmarkov, line 179: Ditto.

src/http/server.rs
```diff
@@ -249,8 +259,17 @@ impl EspHttpServer {
...
 
-                config.0.cacert_pem = cert.as_ptr();
-                config.0.cacert_len = cert.len() as u32;
+                #[cfg(esp_idf_version_major = "4")]
+                {
+                    config.0.client_verify_cert_pem = cert.as_ptr();
+                    config.0.servercert_len = cert.len() as u32;
+                }
+
+                #[cfg(esp_idf_version_major = "5")]
```
ivmarkov, line 268: Ditto.

src/http/server.rs
ivmarkov: OK - thanks for fixing!

src/lib.rs
ivmarkov: But do we have `esp_idf_comp_mdns_enabled` as a valid cfg in ESP-IDF V4.4? I think so, but if you could confirm. I don't want that we are ending with regressions for V4.4.

src/mdns.rs
ivmarkov, line 91: !v4 :)

src/mdns.rs
ivmarkov: Very shaky. I would put a TODO here even though we won't fix it right now.

ivmarkov (changes requested):
If you can address the feedback. Almost all if not all of it is to switch the ESP IDF comparisons to not explicitly reference version "5", as we do everywhere else.

src/lib.rs
zRedShift: `esp_idf_comp_mdns_enabled` is a valid cfg in v4.4

PR #156, opened 2022-11-02
Add support for factory partition info

bshubenok-sigma:
Calling `get_running_slot` on `factory` partition fails as it does not support states, and state is requested via `get_slot`.


ivmarkov:
1. Please `cargo fmt` it
2. Shall we introduce `SlotState::Unsupported` instead?

bshubenok-sigma:
`SlotState::Factory` looks good to me. It will better describe `state` of a slot as `Unchangeable`.

ivmarkov:
Fixed in `master`.

PR #157, opened 2022-11-04
Fix new clippy lints, added null-by-default fields in `httpd_ssl_config_t`"

zRedShift:
- new fields added in `esp-idf/master`, all defaulting to `NULL`
- fix new clippy lints


ivmarkov:
Thanks!

PR #164, opened 2022-11-18
Add mqtt cert_pem, client_cert_pem and client_key_pem

rherrmannr:
I've used this implementation to connect an ESP32-C3 to an AWS endpoint.

This is only the implementation for `#[cfg(esp_idf_version_major = "4")]`. I'm not sure how to implement (and test) it for `#[cfg(not(esp_idf_version_major = "4"))]`

fix #162

Issue #162, opened 2022-11-18
MQTT Certificates


rherrmannr:
Hey, when do you plan to implement the certificates for MQTT?

https://github.com/esp-rs/esp-idf-svc/blob/master/src/mqtt/client.rs#L76
```
// TODO: Future

// pub cert_pem: &'a [u8],
// pub client_cert_pem: &'a [u8],
// pub client_key_pem: &'a [u8],
```

Do you have any suggested workaround to use them?

ivmarkov:
Ideally, whoever needs them should open a PR with the changes. :) For example, you might volunteer, and then the whole community would benefit.

rherrmannr:
I'm already trying to :P

ivmarkov:
Fixed.

PR #167, opened 2022-11-20
Add PSK support to MQTT client

sirhcel:
* I'd like to add support for PSK authentication to the MQTT client
* #164 and #165 have been created since I started working on this and maybe we should join forces to catch these two birds with one stone
    * I'd like to start a discussion here @rherrmannr and @ivmarkov
    * With #165, it looks to me, that [`PskConfiguration`](https://github.com/sirhcel/esp-idf-svc/blob/487a8d0d2e45573e429793888af3570216fbeb85/src/mqtt/client.rs#L48) should move to `tls.rs`
    * What about making the different authentication mechanisms mutually exclusive with an `enum`?
        * The IDF seems to process them in a certain order
        * I tricked myself at the beginning as I provided a CA bundle and a PSK and wondered why the PSK was not used.
        * Wouldn't something like the following more simple/intuitive?
            ```rust
            enum Authentication {
                None,
                GlobalCaStore,
                CaBundle(unsafe extern "C" fn(conf: *mut c_types::c_void) -> esp_err_t),
                Certificates(&str, &str, &str), // server certificate and client certificate with key
                Psk(PskConfiguration),
            }
            ```
* Adding PSK support seem more challenging than passwords and client certificat though
    * PSK data is passed as a pointer to `psk_hint_key_t` to [`esp_mqtt_client_init`](https://github.com/sirhcel/esp-idf-svc/blob/487a8d0d2e45573e429793888af3570216fbeb85/src/mqtt/client.rs#L439)
    * Needs to be valid when starting the connection
    * So I added a boxed instance to the client
        * [Created here](https://github.com/sirhcel/esp-idf-svc/blob/487a8d0d2e45573e429793888af3570216fbeb85/src/mqtt/client.rs#L131) if necessary
        * [Added to `esp_mqtt_client_config_t`]() if necessary
        * [Stored in `EspMqttClient`](https://github.com/sirhcel/esp-idf-svc/blob/487a8d0d2e45573e429793888af3570216fbeb85/src/mqtt/client.rs#L439) to have it available when starting the TLS connection in `esp_mqtt_client_start`
     * Having this data static would require a static `psk_hint_key_t` which does not look like something I'd like to have in this configuration interface at a first glance


sirhcel:
* This is just about my actual changes to PSK authentication which I will rebase later

ivmarkov:
Mr Obvious here, but you are contributing changes done on the tip of the released branch to the `master` branch? Please rebease all your work to the `master` branch. It will anyway be officially released later today.

ivmarkov:
>       ```rust
>       enum TlsVerificationMode {
>           None,
>           GlobalCaStore,
>           CaBundle(unsafe extern "C" fn(conf: *mut c_types::c_void) -> esp_err_t),
>           Certificates(&str, &str, &str), // server certificate and client certificate with key
>           Psk(PskConfiguration),
>       }
>       ```

Would that work across all of HTTP client, server and MQTT client?
Specifically, is PSK support by HTTP client? What does a `GlobalStore` even mean in the context of an HTTP server?

rherrmannr:
So the `MqqtClientConfiguration<'a>` would look like this? 
```rust
pub struct MqttClientConfiguration<'a> {
   pub protocol_version: Option<MqttProtocolVersion>,
    // ...

   pub tls_verification: Option<TlsVerificationMode>
}
```
and removes the members `cert_pem` `client_cert_pem` `client_key_pem` and `psk`?

Is `None` needed, if it's optional? Otherwise, could you specify the non-existing in different ways?
```rust
    let configuration = MqttClientConfiguration {
        client_id: Some("TBD"),

        tls_verification: Some(TlsVerificationMode::None()), // None in enum not needed?
        tls_verification: None(), // would be enough I guess or don't specify it all

        ..Default::default()
    };
```

sirhcel:
> Mr Obvious here, but you are contributing changes done on the tip of the released branch to the `master` branch? Please rebease all your work to the `master` branch. It will anyway be officially released later today.

Not intent to do land the PR this way - just wanted to start a discussion when I saw #165 and #162. Got some time to rebase them onto master meanwhile.

sirhcel:
* Thank you very much for your input! :)

> Would that work across all of HTTP client, server and MQTT client?

* I was just thinking about the clients at first
* The HTTPS server needs a slightly different configuration and I would not try to handle all of them the same

> Specifically, is PSK supported by HTTP client?

* Actually no, the IDF does not support it
* I don't see a reason why it shouldn't in the long term and adding the required configuration options looks feasible to me at a first glance

> What does a `GlobalCaStore` even mean in the context of an HTTP server?

* To be honest, I have no idea. :-D I just thought about clients.


* With your input, I would go for keeping a low profile in the short term
     * Just add PSK support to the MQTT client
     * Leave restructuring the TLS verification method for later (with addressing this issue at the IDF at first)

sirhcel:
> So the `MqqtClientConfiguration<'a>` would look like this?
> 
> ```rust
> pub struct MqttClientConfiguration<'a> {
>    pub protocol_version: Option<MqttProtocolVersion>,
>     // ...
> 
>    pub tls_verification: Option<TlsVerificationMode>
> }
> ```
> 
> and removes the members `cert_pem` `client_cert_pem` `client_key_pem` and `psk`?

This was my first thought. But with @ivmarkov's input, this looks like too much wished in the short term.


> Is `None` needed, if it's optional? Otherwise, could you specify the non-existing in different ways?

No, you are absolutely right. In this case, a dedicated `None` for the authentication method does not make sense.


> And a struct Certificates?
> 
> ```rust
> struct Certificates {
>    pub cert_pem: &'static str,
>    pub client_cert_pem: &'static str,
>    pub client_key_pem: &'static str,
> }
> ```

It was pretty late yesterday. A struct looks way more nice to me as the members have explicit names. I still like this struct more. What about adding it to `tls.rs` and using it for HTTP and MQTT client?

sirhcel:
* What about the latest update @ivmarkov?

ivmarkov:
> * What about the latest update @ivmarkov?
> * The adaption for IDF 5 is a maneuver in the dark
>   
>   * I'm using a modified [`rust-esp-std-demo`](https://github.com/sirhcel/rust-esp32-std-demo/tree/test/mqtt-psk) for testing
>   * How could I build this for IDF 5 or is this not supported by `embuild` yet as the comment in `Cargo.toml` suggests?

* All of the above should work for ESP IDF 5 (it is not that different in this regard from ESP IDF 4)
* I'm not aware of any `Cargo.toml` mentioning anything w.r.t. ESP IDF 5 not supported
* Building for ESP IDF 5 implies switching your ESP IDF branch to `master` and enabling the `espidf_time64` flag in `./cargo/config.toml`

ivmarkov:
> And a struct Certificates?
> 
> ```rust
> struct Certificates {
>    pub cert_pem: &'static str,
>    pub client_cert_pem: &'static str,
>    pub client_key_pem: &'static str,
> }
> ```

I'm not sure that having client authentication with certificates also means that the server should be authenticated (with a certificate) or vice versa. Moreover, you are missing the `ca_bundle` in the equation. I think the server authentication is either with a `ca_bundle` or with the `cert_pem` from the above structure.

To summarize, are you sure that we **have** to unite - in one structure - the client and server certificates authentication?
* If yes, you need to model the `ca_bundle` thing in the above structure
* If not, the above structure does not make sense, and client vs server certificates should be modeled separately. Yet, you still need to model `ca_bundle` vs directly-provided server certificate.

sirhcel:
>     * All of the above should work for ESP IDF 5 (it is not that different in this regard from ESP IDF 4)

* And it finally did (as described in #170). Thank you very much!

>     * I'm not aware of any `Cargo.toml` mentioning anything w.r.t. ESP IDF 5 not supported

* Sorry, I messed this up with https://github.com/ivmarkov/rust-esp32-std-demo/blob/09f1a899dbc112d19ffcf0eb3f0e66a69e3b2499/.cargo/config.toml#L39

>     * Building for ESP IDF 5 implies switching your ESP IDF branch to `master` and enabling the `espidf_time64` flag in `./cargo/config.toml`

* Yes, this did the trick. :)

sirhcel:
> I'm not sure that having client authentication with certificates also means that the server should be authenticated (with a certificate) or vice versa. Moreover, you are missing the `ca_bundle` in the equation. I think the server authentication is either with a `ca_bundle` or with the `cert_pem` from the above structure.

Me neither. I agree with you that server authentication should be either possible with a CA bundle or a single certificate.

This was my reflex after stepping into some pitfalls when adding PSK support and when I saw the other PRs. After having slept two nights over it, this is way more complex than I initially thought. For me this needs some more time, some careful thinking and likely small changes to the ESP IDF itself. I still think that an API which prevents inconsitent configuration would be nice but it is nothing to rush ahead with.

> To summarize, are you sure that we **have** to unite - in one structure - the client and server certificates authentication? As that would imply both use SSL authentication, or neither (client and server) uses it. I don't see such an indication in the MQTT protocol?
> 
>     * If yes, you need to model the `ca_bundle` thing in the above structure
> 
>     * If not, the above structure does not make sense, and client vs server certificates should be modeled separately. Yet, you still need to model `ca_bundle` vs directly-provided server certificate.

It would be nice. But now I see that this would require way more careful thinking.

What about adding PSK support as is and leaving the rest for now?

ivmarkov:
> What about adding PSK support as is and leaving the rest for now?

Open for that. And it can be an incremental, backward-compatible change.

sirhcel:
> Open for that.

Let's do it then.


> And it can be an incremental, backward-compatible change.

I have no idea how this could look like so far. But time might bring some. :) And i bet there will be some breaking changes coming from other corners and windows for getting a breaking change in the configuration into the project.

sirhcel:
While this PR contains the code for `esp-idf-svc` it still requires adding `CONFIG_ESP_TLS_PSK_VERIFICATION=y` to the SDK configuration of the actual project. Is there currently a standardized way of communicating this to dependent crates? Or is `.github/configs/sdkconfig.default` a hint for this?

ivmarkov:
> While this PR contains the code for `esp-idf-svc` it still requires adding `CONFIG_ESP_TLS_PSK_VERIFICATION=y` to the SDK configuration of the actual project. Is there currently a standardized way of communicating this to dependent crates? Or is `.github/configs/sdkconfig.default` a hint for this?

Adding a `CONFIG_FOO` configuration in `sdkconfig.defaults.*` automagically results in the definition of a Rust `#[cfg(...)]` flag with name `esp_idf_foo` available for consumption by all crates that have a direct or an indirect dependency on `esp-idf-sys`. So yeah, you can use `esp_idf_esp_tls_psk_verification` as a boolean cfg flag in `esp-idf-svc` to conditionally compile stuff which would only be available when the ESP IDF kconfig flag `CONFIG_ESP_TLS_PSK_VERIFICATION` is set to true ('y'). You can look around in the `esp-idf-svc` source-code and you would see plenty of such flags in use.

sirhcel:
Cool thing I did not know about before. Thank you!

This crate does not use conditional compilation for the TLS certificate bundle and the  options to enable it are present in [`.github/configs/sdkconfig.defaults`](https://github.com/esp-rs/esp-idf-svc/blob/336a8afe362f3835def7a12aea76969be2c8f419/.github/configs/sdkconfig.defaults). Just for CI builds as far as I understand it. Am I right with this assumption? If yes, shouldn't `CONFIG_ESP_TLS_PSK_VERIFICATION=y` be added to to the SDK configuration file too?

The configuration `CONFIG_ESP_TLS_PSK_VERIFICATION` for IDF is not required for building the Rust bindings against IDF but if one want to use TLS PSK they are [required at runtime](https://github.com/espressif/esp-idf/blob/6407ecb3f8d2cc07c4c230e7e64f2046af5c86f7/components/esp-tls/esp_tls_mbedtls.c#L635). As far as I understood things as of now, the application crate needs to enable it in their `sdkconfig.defaults.*` file. How does the user crating an applications gets to know about this configuration option? Good examples like `rust-esp32-std-demo`? Or is there any mechanism for automatically adding this config option? For example when initializing the template appilication?

src/tls.rs
```diff
@@ -1,6 +1,55 @@
...
+#[derive(Copy, Clone, Eq, PartialEq)]
+pub struct Psk<'a> {
+    pub key: &'a [u8],
+    pub hint: &'a str,
+}
+
+impl<'a> Debug for Psk<'a> {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> Result<(), core::fmt::Error> {
+        f.debug_struct("Psk")
+            .field("hint", &self.hint)
```
ivmarkov, line 17: You sure that's OK?

src/tls.rs
```diff
@@ -1,6 +1,55 @@
...
+///
+/// It could be easily converted from the public `Psk` configuration and holds the `psk_hint_key_t`
+/// along with its (string) data as this data typically needs to be around after initializing a TLS
+/// client until it has been started.
+pub(crate) struct TlsPsk {
+    pub(crate) psk: Box<psk_hint_key_t>,
+    pub(crate) _cstrs: RawCstrs,
+}
+
+impl Debug for TlsPsk {
```
ivmarkov, line 32: `#[cfg(feature = "alloc")]`

src/tls.rs
```diff
@@ -1,6 +1,55 @@
...
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> Result<(), core::fmt::Error> {
+        f.debug_struct("TlsPsk")
+            .field("psk", &self.psk)
+            .finish_non_exhaustive()
+    }
+}
+
+impl<'a> From<&'a Psk<'a>> for TlsPsk {
```
ivmarkov, line 40: `#[cfg(feature = "alloc")]`

src/mqtt/client.rs
```diff
@@ -80,8 +80,8 @@ pub struct MqttClientConfiguration<'a> {
     pub client_certificate: Option<X509<'static>>,
     pub private_key: Option<X509<'static>>,
     pub private_key_password: Option<&'a str>,
-    // TODO: Future
-    // pub psk_hint_key: KeyHint,
+
+    pub psk: Option<Psk<'a>>,
```
ivmarkov, line 84: Protect with `#[cfg(esp_idf_esp_tls_psk_verification)]`. This way the user **must** enable this config option to use the functionality. The fact that ESP IDF requires it only at runtime is bad design IMO, that we should not follow.

src/mqtt/client.rs
```diff
@@ -449,6 +458,15 @@ impl<S> EspMqttClient<S> {
             c_conf.broker.address.uri = cstrs.as_ptr(url);
         }
 
+        #[cfg(esp_idf_version_major = "4")]
```
ivmarkov, line 461: `#[cfg(esp_idf_esp_tls_psk_verification)]`

src/mqtt/client.rs
```diff
@@ -449,6 +458,15 @@ impl<S> EspMqttClient<S> {
...
+        if let Some(ref conf) = tls_psk_conf {
+            c_conf.psk_hint_key = &*conf.psk;
+        }
+        #[cfg(not(esp_idf_version_major = "4"))]
```
ivmarkov, line 465: `#[cfg(esp_idf_esp_tls_psk_verification)]`

ivmarkov (changes requested):
Some stylistic comments. Also you since you use the `alloc` crate in the `tls` module now, you need some conditional compilation there.

src/mqtt/client.rs
sirhcel: This looks like the thing I was looking for in my last commit. I like this way of being on the same page in esp-idf-svc and esp-idf about this feature gate. :)

src/tls.rs
sirhcel: It should and apparently does according to my test with logging a `Psk` struct. This support is available in _core_ and gives a debug output like `Psk { hint: "my-hint-is-longer-than-yours", .. }`.

sirhcel:
> Some stylistic comments. Also you since you use the `alloc` crate in the `tls` module now, you need some conditional compilation there.

Thank you for your feedback! Will do.

I had assumed that esp-idf-svc implies std due to its foundation ESP IDF relying on dynamic memory allocations. What targets the feature gate for `alloc`? Is this a `no_std` Rust environment on top on ESP IDF?

Could you please help me out with a configuration for non-alloc builds? I tried
```
$ export ESP_IDF_SDKCONFIG_DEFAULTS=.github/configs/sdkconfig.default
$ cargo build --no-default-features --features nightly,experimental --target riscv32imc-esp-espidf -Zbuild-std=core,panic_abort -Zbuild-std-features=panic_immediate_abort
$ cargo build --no-default-features --features nightly,experimental,alloc --target riscv32imc-esp-espidf -Zbuild-std=core,panic_abort -Zbuild-std-features=panic_immediate_abort
```
but they trigger a whole bunch of errors relatet to the crate `alloc` not being found in dependencies.

sirhcel:
> Some stylistic comments. Also you since you use the `alloc` crate in the `tls` module now, you need some conditional compilation there.

What about feature-gating all the TLS PSK support behind `#[cfg(all(esp_idf_esp_tls_psk_verification, feature = "alloc"))]`?

src/tls.rs
```diff
@@ -1,6 +1,55 @@
...
+            .finish_non_exhaustive()
+    }
+}
+
+/// Helper for holding PSK data for lately initialized TLS connections.
+///
+/// It could be easily converted from the public `Psk` configuration and holds the `psk_hint_key_t`
+/// along with its (string) data as this data typically needs to be around after initializing a TLS
+/// client until it has been started.
+pub(crate) struct TlsPsk {
```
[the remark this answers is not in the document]
sirhcel: Used `#[cfg(all(esp_idf_esp_tls_psk_verification, feature = "alloc"))]`

src/tls.rs
sirhcel: Used `#[cfg(all(esp_idf_esp_tls_psk_verification, feature = "alloc"))]`

src/mqtt/client.rs
sirhcel: Used `#[cfg(all(esp_idf_esp_tls_psk_verification, feature = "alloc"))]`

src/mqtt/client.rs
sirhcel: Used `#[cfg(all(esp_idf_esp_tls_psk_verification, feature = "alloc"))]`

src/mqtt/client.rs
sirhcel: Used `#[cfg(all(esp_idf_esp_tls_psk_verification, feature = "alloc"))]`

src/tls.rs
sirhcel: Used `#[cfg(all(esp_idf_esp_tls_psk_verification, feature = "alloc"))]`

src/tls.rs
```diff
@@ -1,6 +1,55 @@
...
+    pub(crate) psk: Box<psk_hint_key_t>,
+    pub(crate) _cstrs: RawCstrs,
+}
+
+impl Debug for TlsPsk {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> Result<(), core::fmt::Error> {
+        f.debug_struct("TlsPsk")
+            .field("psk", &self.psk)
```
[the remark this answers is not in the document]
sirhcel: Yes, see https://github.com/esp-rs/esp-idf-svc/pull/167#discussion_r1057358727

src/mqtt/client.rs
```diff
@@ -299,7 +313,9 @@ impl<'a> From<&'a MqttClientConfiguration<'a>> for (esp_mqtt_client_config_t, Ra
             }
         }
 
-        (c_conf, cstrs)
+        let psk_helper = conf.psk.as_ref().map(|psk| psk.into());
```
ivmarkov, line 316: This won't compile.

ivmarkov (changes requested):

src/mqtt/client.rs
sirhcel: Thank you for pointing this out! Fixed.

sirhcel:
* Thank you for starting the matrix builds! Now I've got the build configurations I was looking for.
* Added the conditional compilation with a single feature gate as layed out in https://github.com/esp-rs/esp-idf-svc/pull/167#issuecomment-1365863545
* Added new dimension for TLS PSK to have this configuration tested
    * I know, that this creates a significant CI load
    * Are there other things around which depend on the SDK configuration?
    * And how are they handled?

* Should be good to go from my side now
* Thanks for you patience with this PR!

ivmarkov:
Np - will try to review in the next days or latest by the weekend.

src/mqtt/client.rs
```diff
@@ -122,12 +123,17 @@ impl<'a> Default for MqttClientConfiguration<'a> {
...
+
+            #[cfg(all(esp_idf_esp_tls_psk_verification, feature = "alloc"))]
+            psk: None,
         }
     }
 }
 
 #[cfg(esp_idf_version_major = "4")]
-impl<'a> From<&'a MqttClientConfiguration<'a>> for (esp_mqtt_client_config_t, RawCstrs) {
+impl<'a> From<&'a MqttClientConfiguration<'a>>
```
ivmarkov, line 134: This is still NOK, sorry! You are missing the "alloc" protection in the `cfg` block. If `alloc` is not enabled, the `TlsPsk` struct won't be even defined. Please run a build with:
* Both std and alloc disabled
* Alloc enabled only
* Std enabled(which implies alloc)!

src/mqtt/client.rs
```diff
@@ -200,13 +206,20 @@ impl<'a> From<&'a MqttClientConfiguration<'a>> for (esp_mqtt_client_config_t, Ra
...
+        let tls_psk_conf = None;
+
+        (c_conf, cstrs, tls_psk_conf)
     }
 }
 
 #[allow(clippy::needless_update)]
 #[cfg(not(esp_idf_version_major = "4"))]
-impl<'a> From<&'a MqttClientConfiguration<'a>> for (esp_mqtt_client_config_t, RawCstrs) {
+impl<'a> From<&'a MqttClientConfiguration<'a>>
```
ivmarkov, line 220: Ditto.

ivmarkov (changes requested):

mbv:
@sirhcel 
Hello,

First of all, thank you for initiating the work on adding PSK support for MQTT. I appreciate your effort in bringing this functionality to the project.

I've noticed that there hasn't been much progress on this for the past few months. As someone who is interested in this feature, I'm curious about the status of the development. Are there any plans to complete it? If you need any assistance or support in getting it done, I'd be more than willing to help.

mbv:
@sirhcel 
Hello,

First of all, thank you for initiating the work on adding PSK support for MQTT. I appreciate your effort in bringing this functionality to the project.

I've noticed that there hasn't been much progress on this for the past few months. As someone who is interested in this feature, I'm curious about the status of the development. Are there any plans to complete it? If you need any assistance or support in getting it done, I'd be more than willing to help.

sirhcel:
Hello @mbv,

> I've noticed that there hasn't been much progress on this for the past few months. As someone who is interested in this feature, I'm curious about the status of the development. Are there any plans to complete it? If you need any assistance or support in getting it done, I'd be more than willing to help.

I'm pretty busy since a while, my local changes are working well for me and so I circumvented tackling the different build configurations since then. I'm still interested in getting this PR landed.

I really could use some help. The thing which hinders me most and is least fun for me is figuring out the relevant build configurations for building this crate locally for testing (like the ones shown in https://github.com/esp-rs/esp-idf-svc/pull/167#issuecomment-1365508652). With them at hand, getting the last things sorted out should be quick.

@ivmarkov, how do you build this crate for local development and how did you trigger the build issues from your last review?

Vollbrecht:
you can check the CI jobs for the different targets and feature flags to see how you can locally build the library only

mbv:
I've rebased sirhcel's changes onto the master branch and specified a fixed Rust nightly version, namely nightly-2023-08-07, which appears to be the latest version compatible with the CI without causing failures. 

All builds, including those involving no_std and alloc, for the code version containing sirhcel's changes, have been completed successfully: https://github.com/mbv/esp-idf-svc/actions/runs/5846647015

.github/configs/sdkconfig.tls-psk
```diff
@@ -0,0 +1,5 @@
+# Workaround for https://github.com/espressif/esp-idf/issues/7631
```
ivmarkov, line 1: Can we avoid introducing another sdkconfig file? This would basically double the CI time, which is already long.

ivmarkov (changes requested):

ivmarkov:
@mbv Sorry for reacting late. I'm fine with merging this, as long as we can avoid doubling the CI build matrix (see review my comment).

.github/configs/sdkconfig.tls-psk
sirhcel: My intent behind adding this file was to communicate the required SDK configuration fragments to users of MQTT PSK. If it makes explode CI times, I have no issue with removing it.

Is there another/better way of documenting the required sdk fragments?

sirhcel: What about adding `CONFIG_ESP_TLS_PSK_VERIFICATION=y` with a comment to `.github/configs/sdkconfig.defaults`?

sirhcel: Removed `sdkconfig.tls-psk` and replaced it with a hint in `sdkconfig.defaults`. Also removed TLS PSK support from the matrix build.

This results in TLS PSK no longer being regularly built in CI. I don't see how we could get these builds without the overhead from the matrix builds. But if this is a stepping stone towards getting this PR merged, let's do this step and increase CI build coverage for this feature in an upcoming one.

sirhcel:
> you can check the CI jobs for the different targets and feature flags to see how you can locally build the library only

Thank you very much for stating this one more time! I did not see the wood for the trees for a long time and learned today that the little triangles are the key to the build configurations I was looking for: environment and the complete Cargo command line was just a click away hiding in plain sight.

PR #169, opened 2022-11-21
Fix no-std compilation

taks:
Compilation errors occurred when compiling with no-std, so I fixed it.


ivmarkov:
Thanks! We were only doing a lip service to `no_std` / `alloc` builds, as these were not properly covered in the CI. I'm fixing this as well.

PR #172, opened 2022-11-22
fix dma channel

### LINKED ISSUE #171: invalid dma channel, chip only support spi dma channel auto-alloc
When I want to initialize `EthDriver` with spi, I get this error

```
E (332) spi: spi_bus_initialize(762): invalid dma channel, chip only support spi dma channel auto-alloc
```

I think I located the error in the code: https://github.com/esp-rs/esp-idf-svc/blob/e20c313f574ee5eacf6ddbc6c2827e961f002d1b/src/eth.rs#L526

As a comment, it states `SPI_DMA_CH_AUTO` which is defined as `3` in [pi_common.h](https://github.com/espressif/esp-idf/blob/439a709c42/components/driver/include/driver/spi_common.h).

teamplayer3:
fixes #171

### DIFF
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -523,7 +523,7 @@ impl<'d> EthDriver<'d> {
             ..Default::default()
         };
 
-        esp!(unsafe { spi_bus_initialize(P::device(), &bus_config, 1) })?; // SPI_DMA_CH_AUTO
+        esp!(unsafe { spi_bus_initialize(P::device(), &bus_config, 3) })?; // SPI_DMA_CH_AUTO
 
         Ok(())
     }


Vollbrecht:
to make it more nice we maybe want to get rid of the magic number and use the Dma enum from esp-idf-hal/spi https://github.com/esp-rs/esp-idf-hal/blob/master/src/spi.rs#L63 since esp-idf-svc already depends on esp-idf-hal ?

reinismu:
@ivmarkov  Any plans to introduce e2e tests, so these issues wouldn't happen again?

ivmarkov:
> @ivmarkov Any plans to introduce e2e tests, so these issues wouldn't happen again?

We have very limited number of resources, in that I am only doing this in my free time, and the Espressif folks are currently focused on the bare-metal portfolio, where there is a lot to catch on.

With that said, I'm absolutely for introducing tests (and documentation), however - given the resource availability - I think the community should step in and contribute in these areas. :)

reinismu:
> > @ivmarkov Any plans to introduce e2e tests, so these issues wouldn't happen again?
> 
> We have very limited number of resources, in that I am only doing this in my free time, and the Espressif folks are currently focused on the bare-metal portfolio, where there is a lot to catch on.
> 
> With that said, I'm absolutely for introducing tests (and documentation), however - given the resource availability - I think the community should step in and contribute in these areas. :)

I would love to help here, but from what I can tell to test this correctly would need to run on an actual device. To be fair I'm not too sure how to run tests on a real device :/ To do this properly I feel we would need to have some fleet of esp32/s2/s3/c3. This does feel like something Espressif could easily build.
 
Do you know any examples of how to run tests on real devices? If we had a framework here on how to do it, I would be willing to contribute more tests.

Interesting that Espressif doesn't bet more on Rust. I hope they do financially support this project at least :)

PR #177, opened 2022-11-24
cleanup stop and disconnection calls

teamplayer3:
Minor fix.

I think scanning can be implemented without disconnecting and stopping the driver in the future.

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -370,8 +370,6 @@ impl<'d> WifiDriver<'d> {
     pub fn stop(&mut self) -> Result<(), EspError> {
         info!("Stop requested");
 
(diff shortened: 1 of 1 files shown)


ivmarkov:
@zRedShift IS your fix similar to this one, or more involved?

zRedShift:
@ivmarkov I opted for a separate method for a non-blocking scan, with `esp!(esp_wifi_scan_start(ptr::null_mut(), false))?;` and waiting for a `ScanDone` event.  I have an event loop running in `async` that knows the state of the wifi, never disconnecting/stopping it when unnecessary. It's a hard requirement for meto juggle between Ap -> ApSta <-> Sta modes seamlessly without dropping connections.

This PR is just a double disconnect fix and should probably be merged as is.

PR #185, opened 2022-12-12
Ensure HTTP server responds to JSON requests in JSON

bsodmike:
It's a bit awkward handling errors client side as `text/html` when working in the context of machine-to-machine comms via JSON.

### DIFF
--- a/src/http/server.rs
+++ b/src/http/server.rs
@@ -13,7 +13,7 @@ use alloc::string::ToString;
 use alloc::sync::Arc;
 use alloc::vec::Vec;
 
-use ::log::{info, warn};
+use log::{info, warn};
 
 use embedded_svc::http::headers::content_type;
 use embedded_svc::http::server::{
@@ -25,6 +25,7 @@ use embedded_svc::utils::http::server::registration::{ChainHandler, ChainRoot};
 
 use esp_idf_sys::*;
 
+use serde_json::json;
 use uncased::{Uncased, UncasedStr};
 
 use crate::errors::EspIOError;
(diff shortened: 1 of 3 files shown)


src/http/server.rs
```diff
@@ -825,6 +859,26 @@ impl<'a> EspHttpConnection<'a> {
...
+    where
+        E: Display,
+    {
+        self.initiate_response(
+            500,
+            Some("Internal Error"),
+            &[content_type("application/json")],
+        )?;
+
+        let json = json!({
```
bsodmike, line 872: This may not be non-alloc/no-std friendly 🤔

ivmarkov:
Putting aside the re-introduction of the `json` crate into the codebase that I got rid of a month ago, I think such a code belongs to a utility method of yours, rather than to the core HTTP server. I've never seen such a thing in a generic HTTP server.

bsodmike:
> Putting aside the re-introduction of the `serde_json` crate into the codebase that I got rid of a month ago

Sorry!

> I think such a code belongs to a utility method of yours, rather than to the core HTTP server. I've never seen such a thing in a generic HTTP server.

Is that possible given that `to_native_handler` is implemented on the struct itself? How can one override it in the context of an app?

ivmarkov:
> Is that possible given that `to_native_handler` is implemented on the struct itself? How can one override it in the context of an app?

Of course. All you need to do is wrap your handlers with a middleware, that catches the potential error from the handler, renders it however you want and then returns an OK response to the server. The error handling in the server is a last resort so to say, and it does not use any ESP IDF internals.

bsodmike:
Got it. Thanks!

PR #187, opened 2022-12-13
Fix for `ESP-Now` with `esp-idf` master

zRedShift:
- fix `ESP NOW` for `esp-idf` `master` and `v5.1+`
- use `EspError::from_infallible`


ivmarkov:
@zRedShift There are a few wrong casts related to where `.EspIOError` is expected, while you provide plain `EspError`

zRedShift:
@ivmarkov yeah I noticed it, but now I see that I accidentally pushed the fixes to the wrong repository, starting CI again

PR #192, opened 2022-12-21
Fix esp_app_desc conversion

iH8c0ff33:
The Esp app description header was not read correctly due to a wrong offset.


ivmarkov:
Hopefully fixed by #205

PR #194, opened 2022-12-23
Wifi driver config only if changed

teamplayer3:
My scenario looks like the following:

```rust
let wifi = WifiDriver::new(...);
wifi.set_configuration(AP mode);
wifi.start();

let listener = TcpListener::new(AP IP Address);
task::spawn(move async { loop { listener.next().await } };

// scan for aps
wifi.set_configuration(AP + STA mode); // "upgrade" ap config to mixed config => old ap config + new sta config

wifi.start_scan(ScanConfig::default(), true);
let result = wifi.get_scan_result();

wifi.set_configuration(AP);
```

The scenario will fail at listening to the `TcpListener` while upgrading wifi config (ap -> mixed). This is because the ap config is set again while setting a mixed config.

The changes in the pull request will set the wifi interface config only if it changed.

In my opinion, this would not change the current behavior. I cannot think of a reason why the wifi interface should be reconfigured if the config has not changed.

Another possibility to make this work would be to set the wifi config to mixed at start. But this would start the sta interface from the beginning as well when it's not needed.

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -1134,6 +1134,7 @@ impl<'d> Drop for EspWifi<'d> {
     }
 }
 
+#[cfg(esp_idf_comp_esp_netif_enabled)]
 unsafe impl<'d> Send for EspWifi<'d> {}
 
 #[cfg(esp_idf_comp_esp_netif_enabled)]
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -370,8 +370,6 @@ impl<'d> WifiDriver<'d> {
     pub fn stop(&mut self) -> Result<(), EspError> {
         info!("Stop requested");
 
-        let _ = esp!(unsafe { esp_wifi_disconnect() });
-
         esp!(unsafe { esp_wifi_stop() })?;
 
         info!("Stopping");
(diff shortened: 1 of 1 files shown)


src/wifi.rs
```diff
@@ -742,6 +742,15 @@ impl<'d> WifiDriver<'d> {
         Ok(result)
     }
 
+    fn set_sta_conf_if_changed(&mut self, conf: &ClientConfiguration) -> Result<(), EspError> {
```
ivmarkov, line 745: Let's make this even simpler and just move the equality check inside the existing `set_sta_conf` / `set_ap_conf` methods?

src/wifi.rs
```diff
@@ -742,6 +742,15 @@ impl<'d> WifiDriver<'d> {
...
+        let curr_config = self.get_sta_conf()?;
+        if !conf.eq(&curr_config) {
```
ivmarkov, line 747: Nit: why not `==`?

ivmarkov (changes requested):

ivmarkov:
@teamplayer3 Are you sure that `get_ap_conf` and `get_sta_conf` do not fail when the driver is NOT configured in AP (respectively in STA) mode?

teamplayer3:
@ivmarkov sorry for late response. I checked it again and had no problems. If the driver was not configured in these modes, the function returns the default configs and does not fail.

PR #198, opened 2022-12-26
add support for different IDF versions in CI build

### LINKED ISSUE #197: add CI build support for different IDF versions
The current CI script doesn't build for different IDF versions. Having CI support for the supported IDF versions ([4.3, 4.4, 5.0, 5.1 (when released)) will make it easier to detect build breaks for specific IDF versions

taunusflieger:
This PR resolves #197. In addition it adds a build matrix to enable in the future to have CI build for multiple targets (currently only riscv32imc-esp-espidf is used).

The PR builds for the following IDF versions: 4.3, 4.4, 5.0 

It is expected that the build will not be successful for all listed IDF versions, as there are issues in the code which needs to be fixed, but this is exactly the point why I propose the PR. We need to become aware of those issues through the CI builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,25 +15,64 @@ jobs:
   compile:
     name: Compile
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        target:
+          - riscv32imc-esp-espidf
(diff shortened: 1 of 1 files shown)


MabezDev:
Closed in favour of #200

PR #200, opened 2022-12-29
Add IDF 5.0 and xtensa targets to CI

usbalbin:


src/espnow.rs
```diff
@@ -163,13 +163,12 @@ impl EspNow {
     }
 
     extern "C" fn recv_callback(
-        #[cfg(any(esp_idf_version_major = "4", esp_idf_version_minor = "0"))] mac_addr: *const u8,
-        #[cfg(not(any(esp_idf_version_major = "4", esp_idf_version_minor = "0")))]
-        esp_now_info: *const esp_now_recv_info_t,
+        #[cfg(any(esp_idf_version_major = "4"))] mac_addr: *const u8,
+        #[cfg(not(any(esp_idf_version_major = "4")))] esp_now_info: *const esp_now_recv_info_t,
         data: *const u8,
```
usbalbin, line 168: `#[cfg(any(esp_idf_version_major = "4", esp_idf_version_minor = "0"))]` leads to problems since
IDF 5.0 will satisfy `esp_idf_version_minor = "0"`. What was the thought behind that condition @zRedShift (added in #187)? Do I need to do anything in addition to my change?

zRedShift: There's a change that breaks this in esp-idf 5.1 (master), this condition means "esp-idf 4.x.x or esp-idf 5.0.x"

usbalbin: To me it looks like that change is also in 5.0. Assuming you mean using esp_now_recv_info_t

usbalbin: https://github.com/espressif/esp-idf/blob/release/v5.0/components/esp_wifi/include/esp_now.h#L99

zRedShift: https://github.com/espressif/esp-idf/commit/6818b3012d854508fc3a0870b93534a2fa36dfd7
this was only merged back to 5.0 10 days ago

zRedShift: https://github.com/espressif/esp-idf/blob/v5.0/components/esp_wifi/include/esp_now.h#L89
this is the official v5.0 release, it doesn't have it.
Not sure what to do, we don't have commit granularity here

usbalbin: Oh

usbalbin: So it might perhaps still be in the works then? Would in that case make sense to wait with my changes until that is merged into 5.0 (assuming it will be)

zRedShift: The v5.0 one is a tag, so it's never gonna get changed.
Both have v5.0.0 version, so until the version at release/v5.0 doesn't get bumped to v5.0.1 both won't be able to coexist
FYI @ivmarkov @igrr @MabezDev

usbalbin: Ok, thanks

taunusflieger: I would suggest to separate fixes for IDF 5.0 and the CI build matrix changes. There is also another PR https://github.com/esp-rs/esp-idf-svc/pull/198 which adresses enhancements to the CI build matrix

usbalbin: So one PR with the new device targets and fixes for the newly discovered problems there and another one with IDF 4.3, 4.4 and 5.0 etc.? The latter being #197? Am I understanding you right?

taunusflieger: sorry for not precise enough. My proposal is:
  - one PR for fixing the discovered problems in the code for IDF 5.0
  - one PR for enhancing the CI build (IDF 4.3, 4.4 and 5.0 + build target matrix) - this is covered in PR #198 (xtensa targets can be added to the matrix)

usbalbin: So more or less remove my changes to .github/workflows/ci.yml?

MabezDev:
Thanks! LGTM

MabezDev (approved):

ivmarkov:
@MabezDev @usbalbin CI fails after this merge. If you could check, thanks.

PR #207, opened 2023-01-04
Add wrapper to specify callback for SNTP

bsodmike:
This has been tested on a vanilla ESP32.

### DIFF
--- a/src/sntp.rs
+++ b/src/sntp.rs
@@ -117,10 +117,9 @@ impl<'a> Default for SntpConf<'a> {
     }
 }
 
-static TAKEN: mutex::Mutex<bool> = mutex::Mutex::wrap(mutex::RawMutex::new(), false);
-
 type SyncCallback = Box<dyn FnMut(Duration) + Send + 'static>;
 static SYNC_CB: Mutex<Option<SyncCallback>> = Mutex::new(None);
+static TAKEN: mutex::Mutex<bool> = mutex::Mutex::wrap(mutex::RawMutex::new(), false);
 
 pub struct EspSntp {
     // Needs to be kept around because the C bindings only have a pointer.
--- a/src/sntp.rs
+++ b/src/sntp.rs
@@ -1,4 +1,8 @@
+//! SNTP Time Synchronization
+
 use core::cmp::min;
+use core::time::Duration;
+use std::sync::Mutex;
 
 use ::log::*;
 
@@ -115,6 +119,9 @@ impl<'a> Default for SntpConf<'a> {
 
 static TAKEN: mutex::Mutex<bool> = mutex::Mutex::wrap(mutex::RawMutex::new(), false);
 
+type SyncCallback = Box<dyn FnMut(Duration) + Send + 'static>;
+static SYNC_CB: Mutex<Option<SyncCallback>> = Mutex::new(None);
+
 pub struct EspSntp {
     // Needs to be kept around because the C bindings only have a pointer.
     _sntp_servers: [CString; SNTP_SERVER_NUM],
@@ -138,6 +145,23 @@ impl EspSntp {
         Ok(sntp)
     }
 
+    pub fn new_with_callback<F>(conf: &SntpConf, callback: F) -> Result<Self, EspError>
+    where
+        F: FnMut(Duration) + Send + 'static,
+    {
+        let mut taken = TAKEN.lock();
+
+        if *taken {
+            esp!(ESP_ERR_INVALID_STATE)?;
+        }
+
+        *SYNC_CB.lock().unwrap() = Some(Box::new(callback));
+        let sntp = Self::init(conf)?;
+
+        *taken = true;
+        Ok(sntp)
+    }
+
     fn init(conf: &SntpConf) -> Result<Self, EspError> {
         info!("Initializing");
 
@@ -153,6 +177,7 @@ impl EspSntp {
 
         unsafe {
             sntp_set_time_sync_notification_cb(Some(Self::sync_cb));
+
             sntp_init();
         };
 
@@ -161,6 +186,10 @@ impl EspSntp {
         Ok(Self {
             _sntp_servers: c_servers,
         })
+    }
+
+    fn unsubscribe(&mut self) {
+        *SYNC_CB.lock().unwrap() = None;
     }
 
     pub fn get_sync_status(&self) -> SyncStatus {
@@ -173,6 +202,13 @@ impl EspSntp {
             (*tv).tv_sec,
             (*tv).tv_usec,
         );
+
+        if let Some(cb) = &mut *SYNC_CB.lock().unwrap() {
+            let duration = Duration::from_secs((*tv).tv_sec as u64)
+                + Duration::from_micros((*tv).tv_usec as u64);
+
+            cb(duration);
+        }
     }
 }
 
@@ -182,6 +218,7 @@ impl Drop for EspSntp {
             let mut taken = TAKEN.lock();
 
             unsafe { sntp_stop() };
+            self.unsubscribe();
             *taken = false;
         }
 
--- a/src/sntp.rs
+++ b/src/sntp.rs
@@ -2,7 +2,6 @@
 
 use core::cmp::min;
 use core::time::Duration;
-use std::sync::Mutex;
 
 use ::log::*;
 
@@ -123,7 +122,8 @@ impl<'a> Default for SntpConf<'a> {
 #[cfg(feature = "alloc")]
 type SyncCallback = alloc::boxed::Box<dyn FnMut(Duration) + Send + 'static>;
 #[cfg(feature = "alloc")]
-static SYNC_CB: Mutex<Option<SyncCallback>> = Mutex::new(None);
+static SYNC_CB: mutex::Mutex<Option<SyncCallback>> =
+    mutex::Mutex::wrap(mutex::RawMutex::new(), None);
 static TAKEN: mutex::Mutex<bool> = mutex::Mutex::wrap(mutex::RawMutex::new(), false);
 
 pub struct EspSntp {
@@ -160,7 +160,7 @@ impl EspSntp {
             esp!(ESP_ERR_INVALID_STATE)?;
         }
 
-        *SYNC_CB.lock().unwrap() = Some(alloc::boxed::Box::new(callback));
+        *SYNC_CB.lock() = Some(alloc::boxed::Box::new(callback));
         let sntp = Self::init(conf)?;
 
         *taken = true;
@@ -195,7 +195,7 @@ impl EspSntp {
 
     #[cfg(feature = "alloc")]
     fn unsubscribe(&mut self) {
-        *SYNC_CB.lock().unwrap() = None;
+        *SYNC_CB.lock() = None;
     }
 
     pub fn get_sync_status(&self) -> SyncStatus {
@@ -210,7 +210,7 @@ impl EspSntp {
         );
 
         #[cfg(feature = "alloc")]
-        if let Some(cb) = &mut *SYNC_CB.lock().unwrap() {
+        if let Some(cb) = &mut *SYNC_CB.lock() {
             let duration = Duration::from_secs((*tv).tv_sec as u64)
                 + Duration::from_micros((*tv).tv_usec as u64);
 
--- a/src/sntp.rs
+++ b/src/sntp.rs
@@ -11,6 +11,9 @@ use esp_idf_sys::*;
 use crate::private::cstr::CString;
 use crate::private::mutex;
 
+#[cfg(feature = "alloc")]
+extern crate alloc;
+
 const SNTP_SERVER_NUM: usize = SNTP_MAX_SERVERS as usize;
 
 const DEFAULT_SERVERS: [&str; 4] = [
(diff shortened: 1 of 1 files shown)


bsodmike:
It's probably better to replace `new/init` rather than introducing the `new_with...` methods.  The reason I did this was to test these changes in isolation, without breaking or altering the existing public API.

I will make some updates that will require breaking the public API.

src/sntp.rs
```diff
@@ -122,23 +122,29 @@ pub struct EspSntp {
 
 impl EspSntp {
     pub fn new_default() -> Result<Self, EspError> {
-        Self::new(&Default::default())
+        Self::new(&Default::default(), None)
     }
 
-    pub fn new(conf: &SntpConf) -> Result<Self, EspError> {
+    pub fn new(
```
ivmarkov, line 128: * Rather than asking the user to deal with an unsafe "C" fn and with an unsafe `*mut timeval`, how about asking the user to provide a safe callback? As in e.g. `FnMut(core::time::Duration) + Send + 'static`? This would also be symmetric with what `SysTime` returns. Which is a `core::time::Duration`.
* You are introducing a backwards incompatible change. Shall we instead keep backwards compatibility and introduce a new `subscribe` method that takes the callback? As in: `pub fn subscribe<F: FnMut(core::time::Duration) + Send + 'static>(&mut self) -> EspSntpSubscription`? And then, dropping the `SntpSubscription would mean that the user is unsubscribing?

src/sntp.rs
```diff
@@ -152,7 +158,12 @@ impl EspSntp {
         }
 
         unsafe {
-            sntp_set_time_sync_notification_cb(Some(Self::sync_cb));
+            if let Some(provided_cb) = callback {
```
ivmarkov, line 161: I think the user callback should work *in addition to* `Self::sync_cb`, not instead of `Self::sync_cb`.

ivmarkov (changes requested):

src/sntp.rs
bsodmike: Thanks for the advice, but I'm running into this - how do I get around the type conversion for `F`?

```rs
error[E0308]: mismatched types
   --> /home/mdesilva/esp/esp-idf-svc_fork/src/sntp.rs:172:53
    |
167 |     pub fn subscribe<F: FnMut(core::time::Duration) + Send + 'static>(
    |                      - this type parameter
...
172 |             sntp_set_time_sync_notification_cb(Some(callback));
    |                                                ---- ^^^^^^^^ expected fn pointer, found type parameter `F`
    |                                                |
    |                                                arguments to this enum variant are incorrect
    |
    = note:  expected fn pointer `unsafe extern "C" fn(*mut esp_idf_sys::timeval)`
            found type parameter `F`
```

A backwards incompatible change may be required (I need to test again), but you'll notice earlier `sntp_init()` is called after the callback is set.  I need to check if (1) we can set more than one callback, which relates to your other PR comment and (2) if these need to be set before the `sntp_init()` call.

bsodmike: Hmm, for the type conversion I could do

```rs
#[allow(dead_code)]
pub struct Newtype<T>(pub T);

impl<T> Into<unsafe extern "C" fn(*mut esp_idf_sys::timeval)> for Newtype<T> {
    fn into(self) -> unsafe extern "C" fn(*mut esp_idf_sys::timeval) {
        todo!()
    }
}
```

with

```rs
    pub fn subscribe<F: FnMut(core::time::Duration) + Send + 'static>(
        &mut self,
        callback: F,
    ) -> EspSntpSubscription {
        unsafe {
            sntp_set_time_sync_notification_cb(Some(Newtype(callback).into()));
        }

        EspSntpSubscription {}
    }
    ```
    
    Let me know if this is an acceptable approach (not fully completed yet of course), thanks!

File: src/sntp.rs [hunk 3]
```diff
@@ -188,3 +200,22 @@ impl Drop for EspSntp {
...
+#[allow(dead_code)]
+struct Newtype<T>(pub T);
+
+impl<T> From<Newtype<T>> for unsafe extern "C" fn(*mut esp_idf_sys::timeval) {
+    fn from(ntype: Newtype<T>) -> Self {
+        let boxed = Box::new(ntype.0);
+        let ptr: *mut T = Box::into_raw(boxed);
+
+        unsafe {
+            std::mem::transmute::<*mut T, unsafe extern "C" fn(*mut esp_idf_sys::timeval)>(ptr)
```
bsodmike, line 218: @ivmarkov is this approach acceptable, or is there a simpler way to go about this?

bsodmike: Ignore the above, the transmute causes the closure to loose context of its argument - so this is a bad way to approach this.  Open to suggestions, @ivmarkov thanks.

src/sntp.rs
bsodmike: I don't think that's possible https://github.com/espressif/esp-idf/blob/6cfa88ed49b7d1209732347dae55578f4a679c98/components/lwip/apps/sntp/sntp.c#L64-L67

src/sntp.rs
```diff
@@ -163,6 +168,22 @@ impl EspSntp {
         })
     }
 
+    pub fn subscribe<F: FnMut(core::time::Duration) + Send + 'static>(
+        &mut self,
+        callback: F,
+    ) -> EspSntpSubscription {
```
ivmarkov, line 174: The original idea of `EspSntpSubscription` was to avoid the need for having an `unsubscribe` method (dropping `EspSntpSubscription` should unsubscribe). However to model that you can have only one subscription at a time (or zero), you need to lifetime `EspSntpSubscription` as `EspSntpSubscription<'_>` where `'_` is the lifetime of the `&mut self` reference in the `subscribe` method. Unfortunately, this means you cannot call `get_sync_status` while you have an active subscription!

So I suggest you remove the `EspSntpSubscription` struct altogether.

Furthermore, and in retrospective, there is unfortunately a time race issue with the separate `subscribe` method I did not foresee: since `EspSntp::new` immediately starts the SNTP service, you might miss the first time sync, as it might happen before you call the `subscribe` method. Not sure how much of a problem that would be in practice, but if it is a problem, then unfortunately we need to revert to your                                                        original idea of pushing the callback to one of the constructors. That would imply:
* a third constructor - say - `new_with_callback` that takes `F`
* `unsubscribe` probably becomes a private fn and the only way to unsubscribe is to drop the `EspSntp` instance.

src/sntp.rs
```diff
@@ -163,6 +168,22 @@ impl EspSntp {
...
+        unsafe {
+            *SYNC_CB.lock().unwrap() = Some(Box::new(callback));
```
ivmarkov, line 176: Move out of the `unsafe` block. Setting the value of the mutex is not unsafe.

src/sntp.rs
```diff
@@ -163,6 +168,22 @@ impl EspSntp {
...
+            sntp_set_time_sync_notification_cb(Some(Self::sync_cb_for_subscribe));
```
ivmarkov, line 177: I suggest calling `sntp_set_time_sync_notification_cb` only once in the constructor of `EspSntp`, as it used to be the case.

src/sntp.rs
```diff
@@ -174,6 +195,17 @@ impl EspSntp {
...
+            duration + core::time::Duration::from_micros((*tv).tv_usec as u64);
+
+        // FIXME: catch_unwind, so you don't panic across `extern "C"`
+        if let Some(cb) = &mut *SYNC_CB.lock().unwrap() {
+            cb(duration);
+        }
+    }
 }
 
 impl Drop for EspSntp {
```
ivmarkov, line 211: You should call `unsubscribe` from `drop`.

src/sntp.rs
```diff
@@ -188,3 +220,8 @@ impl Drop for EspSntp {
         info!("Dropped");
     }
 }
+
+pub struct EspSntpSubscription {
```
ivmarkov, line 224: As per above - remove.

src/sntp.rs
```diff
@@ -174,6 +195,17 @@ impl EspSntp {
             (*tv).tv_usec,
         );
     }
+
+    unsafe extern "C" fn sync_cb_for_subscribe(tv: *mut timeval) {
+        let duration: core::time::Duration = core::time::Duration::from_secs((*tv).tv_sec as u64);
```
ivmarkov, line 200: Move the duration calculations inside the `if` statement below. t is pointless to compute `duration` if nobody is subscribed.

src/sntp.rs
```diff
@@ -174,6 +195,17 @@ impl EspSntp {
...
+        let duration: core::time::Duration =
+            duration + core::time::Duration::from_micros((*tv).tv_usec as u64);
+
+        // FIXME: catch_unwind, so you don't panic across `extern "C"`
```
ivmarkov, line 204: What exactly is this FIXME `catch_unwind` supposed to mean? ESP IDF has zero support for unwinding the call stack anyway?

src/sntp.rs
```diff
@@ -174,6 +195,17 @@ impl EspSntp {
             (*tv).tv_usec,
         );
     }
+
+    unsafe extern "C" fn sync_cb_for_subscribe(tv: *mut timeval) {
```
ivmarkov, line 199: Please rename back to `sync_cb` as it used to be and restore the `debug!` logs.

ivmarkov (changes requested):

src/sntp.rs
bsodmike: Thanks for the review and feedback. Yes, will proceed with the third constructor and follow rest of your points.

bsodmike:
@ivmarkov I've made the changes requested, thanks.

src/sntp.rs
```diff
@@ -1,4 +1,8 @@
+//! SNTP Time Synchronization
+
 use core::cmp::min;
+use core::time::Duration;
+use std::sync::Mutex;
```
ivmarkov, line 5: Sorry, one more thing that you need to fix: please use the Mutex from `crate::private::mutex::Mutex`. It allows folks to use `esp-idf-svc` in no_std environments as well (not that there is much of a use case for that though).

You can see how to use that mutex in the other modules, i.e. `EspNow`, `Eth` etc. Or just look at the mutex used by the `TAKEN` static var.

ivmarkov (changes requested):

src/sntp.rs
```diff
@@ -113,6 +117,8 @@ impl<'a> Default for SntpConf<'a> {
     }
 }
 
+type SyncCallback = Box<dyn FnMut(Duration) + Send + 'static>;
```
ivmarkov, line 120: ... and while we are on the `no_std` topic, please:
* declare `#[cfg(feature = "alloc")] extern crate alloc;`
* Protect with `#[cfg(feature = "alloc")]` your `SyncCallback` (note that you should turn `Box` into `alloc::boxed::Box`)
* Ditto for `SYNC_CB`

src/sntp.rs
```diff
@@ -138,6 +144,23 @@ impl EspSntp {
         Ok(sntp)
     }
 
+    pub fn new_with_callback<F>(conf: &SntpConf, callback: F) -> Result<Self, EspError>
```
ivmarkov, line 147: Protect with `#[cfg(feature = "alloc")]`

src/sntp.rs
```diff
@@ -138,6 +144,23 @@ impl EspSntp {
...
+    where
+        F: FnMut(Duration) + Send + 'static,
+    {
+        let mut taken = TAKEN.lock();
+
+        if *taken {
+            esp!(ESP_ERR_INVALID_STATE)?;
+        }
+
+        *SYNC_CB.lock().unwrap() = Some(Box::new(callback));
```
ivmarkov, line 157: `alloc::boxed::Box`

src/sntp.rs
```diff
@@ -163,6 +187,10 @@ impl EspSntp {
         })
     }
 
+    fn unsubscribe(&mut self) {
```
ivmarkov, line 190: Protect with `#[cfg(feature = "alloc")]`

src/sntp.rs
```diff
@@ -182,6 +217,7 @@ impl Drop for EspSntp {
             let mut taken = TAKEN.lock();
 
             unsafe { sntp_stop() };
+            self.unsubscribe();
```
ivmarkov, line 220: Protect the call to `unsubscribe` with `#[cfg(feature = "alloc")]`.

The `alloc` changes make sure that:
* You are using Box from the `alloc` module in case user has selected `no_std` but has still enabled `alloc`.  The STD `Box` is a type-alias for `alloc::boxed::Box`, but this type alias is not available in `no_std`.
* The SNTP module can still work, but without a callback in case user has opt out even from `alloc`.

ivmarkov (changes requested):

src/sntp.rs
bsodmike: Thanks for the explanation!

I also added it here, as SYNC_CB shouldn't be available without `alloc`.
```rs
        #[cfg(feature = "alloc")]
        if let Some(cb) = &mut *SYNC_CB.lock().unwrap() {
```

bsodmike:
Thanks @ivmarkov - this should be ready now :)

PR #208, opened 2023-01-04
Add EspOta.get_update()

Nukesor:
This commit adds the `EspOta.get_update()` function to improve the ergonomics of implementing OTA update logic.

The reasoning behind this, that the current approach requires holding a mutable reference to EspOtaUpdate whilste holding a the reference to EspOta, which is tricky when working with message handling event loops.

Our use-case is as follows.
We implemented OTA via Bluetooth in a way that allows background updates while continuing normal operation.

Previously, our approach worked fine, as `EspOta<Update>` was returned as a owned struct.
This allowed us to do something like this:
```rs
    let mut ota: EspOta = EspOta::new().unwrap();
    let mut ota_update: Option<EspOta<Update>> = None;
    let mut ota_bytes_written = 0;
    loop {
        match rx.recv() {
            Ok(message) => {
                handle_message(
                    message,
                    &mut ota,
                    &mut ota_update,
                    &mut ota_bytes_written,
                )?;
            }
            Err(_) => {}
        }
    }
```
`handle_message` handled various cases of normal traffic, ota initialization, ota update packets and ota update aborts.
With this approach, it's possible to abort and restart an update at any time, and it's possible for an update to take a very long time.

However, since `EspOtaUpdate` is now returned as a mutable reference, we can no longer use this approach, as we would have two different mutable references to the same object (`EspOta` & `EspOtaUpdate`).

As a workaround, I propose to introduce a new function `get_update`, which returns an `Option<&mut EspOtaUpdate>`.
That way, we can always only pass the actual `EspOta` object around and access it's unit on-demand, instead of having to handle a mutable reference for the whole duration of the update.

I considered two possible workarounds for this:
- Stop normal operation and only receive OTA update related packages. This is sadly a no-go for our usecase.
- Spawn a dedicated update thread and send messages to that thread via a dedicated channel.
    It would be great if we would not have to do this, as we're already using quite a few threads and channels, which adds a lot of overhead due to channel buffers and context switching.

Both alternatives are rather unconvenient and/or problematic.

### DIFF
--- a/src/ota.rs
+++ b/src/ota.rs
@@ -232,6 +232,9 @@ impl EspOta {
     pub fn initiate_update(&mut self) -> Result<&mut EspOtaUpdate, EspError> {
         self.check_read()?;
 
+        // This might return a null pointer in case no valid partition can be found.
+        // We don't have to handle this error in here, as this will implicitly trigger an error
+        // as soon as the null pointer is provided to `esp_ota_begin`.
         let partition = unsafe { esp_ota_get_next_update_partition(ptr::null()) };
 
         let mut handle: esp_ota_handle_t = Default::default();
@@ -242,6 +245,17 @@ impl EspOta {
         self.0.update_handle = handle;
 
         Ok(&mut self.0)
+    }
+
+    /// Get a handle to the [EspOtaUpdate].
+    /// This is `None`, if [Self.initiate_update] hasn't been called yet or the update has
+    /// finished.
+    pub fn get_update(&mut self) -> Option<&mut EspOtaUpdate> {
+        if self.0.update_partition.is_null() {
+            None
+        } else {
+            Some(&mut self.0)
+        }
     }
 
     pub fn mark_running_slot_valid(&mut self) -> Result<(), EspError> {


src/ota.rs
```diff
@@ -325,7 +335,7 @@ impl EspOta {
     }
 
     fn check_read(&self) -> Result<(), EspError> {
-        if self.0.update_partition.is_null() {
```
Nukesor, line 328: This section was previously a bit problematic, as `esp_ota_get_next_update_partition` was capable of returning null pointers in case the a update partition couldn't be found.

Keeping this in mind, working with an `Option<EspOtaUpdate>` seems like the cleaner solution anyway, as the default null pointer holds a bit of a special meaning

src/ota.rs
```diff
@@ -220,16 +217,29 @@ impl EspOta {
     pub fn initiate_update(&mut self) -> Result<&mut EspOtaUpdate, EspError> {
         self.check_read()?;
 
-        let partition = unsafe { esp_ota_get_next_update_partition(ptr::null()) };
+        // TODO: The documentation of `esp_ota_get_next_update_partition` states, that the returned
+        // pointer may be a null pointer, which indicates that:
+        // "indicates invalid OTA data partition, or that no eligible OTA app slot partition was found."
+        //
+        // This should probably be handled in here and throw a proper EspError.
+        let update_partition = unsafe { esp_ota_get_next_update_partition(ptr::null()) };
```
Nukesor, line 225: I wasn't sure what kind of error should be thrown in case of a null ptr, hence the `TODO` block.

Nukesor:
Edit: I just stumbled upon the `EspOtaUpdate::abort` function.
This complicates things a bit, as this function would only reset the `EspOtaUpdate`, without setting the update to `None`.

To be consistent, we the `complete` and `abort` functions would have to be moved to `EspOta`, which doesn't make a lot of sense from a semantic point of view and would also be a breaking change.

I'm starting to think, whether it makes sense to split this into a `EspOta` and `EspOtaUpdate` struct in the first place.
I assume this is important to stay consistent with the `embedded_svc` crate?

Adding a separate `EspOta::abort/complete` function without adjusting `EspOtaUpdate` is somewhat out of question , as this would result in an inconsistent state.

Do you have any ideas on how to solve this?

Nukesor:
A possible solution to this could be to add a flag to the `EspOtaUpdate`, which is set as soon as a `complete` or `abort` call has happened.

This would then be used to indicate the `EspOta` object, that the current instance should be discarded.

ivmarkov:
I'll inspect upon my return, but something in your arguments sounds fishy anyway (or maybe the previous implementation was incorrect). The thing is, while the previous implementation returned an owned object, this object was a GAT type containing a &mut reference to `EspOta`. Which goes to say that even with the previous impl, `EspOta` was supposed to be unavailable while `EspOtaUpdate` was alive.

Nukesor:
You're right. I overtook the task of updating our firmware version half-way through and I seem to got a wrong impression on how the previous version worked.

Previously it were some `EspOta<Read>` and `EspOta<Update>` references.
But it seems like the previous instance was faulty, if the original `EspOta<Read>` should've no longer been used., as `EspOta<Update>` didn't contain an `&mut` reference to it.

I adjusted my example accordingly.

ivmarkov:
I think with the current implementation (or the previous GAT one, if it was correct anyway) you can still do your event loop, except it needs to be a nested one. I.e., spin the outer loop while you are not in an OTA update state. If you get an event request to initiate an OTA update, start a second, inner loop, with the mutable borrow of `EspOtaUpdate` and your event loop receiver. Spin the second loop until the update is running. Once it is complete or aborted, exit the inner loop, drop the `&mut` ref to `EspOtaUpdate` and get back into the outer loop. No?

Nukesor:
That's the first possible solution for the problem I proposed.
This is possible, but it's really unwieldy from a development perspective.

We aim to keep up normal operations during the update. The above example is probably bit misleading, as our Bluetooth message handling code inside the loop is quite complex and has many edge-cases.

I hope the following pseudocode will give you a better understanding of our use case:
```
loop {
    match rx.recv() {
        // Handle OTA write
        Ok(BleChannelCallbackData::OtaWrite(data)) => {
            handle_ota_write(
                    data,
                    events_storage.clone(),
                    &mut ota_update,
                    &mut ota_bytes_written,
                    );
        }
        /// Multiple other callback cases
        ...

        Ok(BleChannelCallbackData::AppMessage(data)) => {
            /// Data handling code
            ...
            ...

            if let Err(e) = communication::process_app_message(
                    message,
                    &mut ota,
                    &mut ota_update,
                    &mut ota_bytes_written,
                    ) {
                error!("Failed to process app message:{}", e);
            }
        }

        ...
    }
}

fn process_app_message(message: Message, ota: &mut EspOta, ota_update: &mut EspOtaUpdate, ota_bytes_written: usize, ...) {
    match message {
        AppMessagePayload::OtaInit(payload) => {
            begin_ota(payload)
        }

        /// About ~40 other message types, that're handled in here.
        ...
    }
}
```

We would have to create a separate `process_app_message`, which doesn't contain the `begin_ota` message handling and doesn't take a `&mut ota` reference.
That function would then need to be called from a virtually identical duplication of the `match` over the `BleChannelCallbackData`.

This approach is propably possible, but it introduces a lot of code duplication and makes the message handling flow pretty ugly and obfuscated.

How about we just consume the `EspOta` struct as soon as the update is started? This would make things easier as well.

ivmarkov:
> How about we just consume the `EspOta` struct as soon as the update is started? This would make things easier as well.

Well if you have such a large amount of match cases you probably have to split those into smaller functions anyway? But with that said, I'm not opposed to having a cosumption-based API in addition to the existing one. If you would like to implement one in a PR, I'll review and merge.

Nukesor:
One quick qestion:

Why exactly is the `TAKEN` lock acquired as soon as `EspOta` is created?
Couldn't we just move the Lock instantiation to the `EspOtaUpdate` struct? That way we wouldn't even have to consume the `EspOta` struct, as there wouldn't be any downside to keeping it.

You explained to me, that the `EspOta` struct may only ever be instantiated once, but as far as I can see, the only critical operations are on the `EspOtaUpdate` struct. I feel like I'm missing something.

ivmarkov:
> ```rust
>     fn initiate_update(&mut self) -> Result<&mut Self::Update, Self::Error>;
> ```
> 
> ->
> 
> ```rust
>     fn initiate_update(self) -> Result<Self::Update, Self::Error>;
> ```
> 
> If this would be ok for you, I'll go ahead and prepare two PRs accordingly.

That would mean losing backwards compat and I hope there are at least a few other folks that use the current API successfully. Do you have to go via the traits API by all means? Are you targetting - with your use case - anything else besides ESP32? If not, just expose `into_update(self)` onto `EspOta` itself and then - onto `EspOtaUpdate` - something like `into_ota_complete` and `into_ota_abort`. I also suggest you think twice if you really cannot live with the current API. I can't really judge objectively, as I do not have a perspective on your whole app design, but it does sound a bit as if you are trying to twist the API around your current code a bit too much. Which is a red herring.

ivmarkov:
> One quick qestion:
> 
> Why exactly is the `TAKEN` lock acquired as soon as `EspOta` is created? Couldn't we just move the Lock instantiation to the `EspOtaUpdate` struct? That way we wouldn't even have to consume the `EspOta` struct, as there wouldn't be any downside to keeping it.
> 
> You explained to me, that the `EspOta` struct may only ever be instantiated once, but as far as I can see, the only critical operations are on the `EspOtaUpdate` struct. I feel like I'm missing something.

There can only be one or zero  EspOtaUpdate instances (or mut refs) at a time. If you have multiple EspOta instances, how do you guarantee that?

Nukesor:
> There can only be one or zero EspOtaUpdate instances (or mut refs) at a time. If you have multiple EspOta instances, how do you guarantee that?

Right now, we lock a global mutex when instantiating `EspOta`, however, it would make sense to move that locking logic to the `EspOtaUpdate` struct, as this is the struct we're actually caring about.

That way`, we could call `EspOta::initiate_update()`, which returns either an `EspOtaUpdate` in case there isn't already a running update or a `EspError` in case the flag is already locked.

I created two POC branches that can be found over here:
[esp-idf-svc draft](https://github.com/Nukesor/esp-idf-svc/commit/9491ee665c32a271dfb08a455e4dcfe56041cf59)
[embedded-svc draf](https://github.com/Nukesor/embedded-svc/commit/784ab46ca68f5c295a508d7788214780b0b3c020)

ivmarkov:
With this change you just lost the benefit of ensuring - at compile time - that you are having a single `EspOtaUpdate` instance; you can then also remove the `&mut` from `initiate_update(&mut self)` as it does not make any sense anymore? And the OTA API is now different from everything else in the crate which operates on singletons, as `EspOta` is no longer a singleton.

ivmarkov:
One more thing: these changes are not backwards compatible (as I asked for) in one of my previous comments.

Nukesor:
In understand, that you're hesitant to change the API in a backward compatible way that breaks a existing convention in this project.

My main intention with this PR was to think about a way to make this API a bit easier to work with, as it felt like there're a few limitations that aren't strictly necessary.

In case no of the proposed approaches match the project's guidelines, just say the word and I'll close the PR :)

Nukesor:
One thing that should be addressed under all circumstances, is that a null pointer can be returned if no partition can be found. This behavior should be caught and a error should be thrown.

ivmarkov:
OK. I've written detailed justifications below why I'm not OK with your proposal. I'll appreciate if you check these, as I don't think this is only a matter of taste or "ad hoc" guidelines though having some sort of symmetry in the whole crate is a good thing nevertheless. The summary is:
1. I don't believe that shifting the mutex from acquiring `EspOta` to `EspOtaUpdate` is giving that much of an ergonomics improvement, if any. And you've missed some APIs. Details below
2. Abandoning the singleton pattern is dangerous. You seem to be ignoring the elephant in the room which is the ESP IDF OTA code - a pile of unsafe C. Details below
3. (Less important but still a bit annoying) You are jumping from a relatively smaller enhancement (`self`-consuming variation of `initiate_update`)  into shifting the lock elsewhere - and - ignoring my request not to introduce backwards incompatible changes. Also: did you even try my suggestion with the inner loop or just disregarded it based on theoretical arguments?

On (1):
Assuming that your changes - i.e. shifting the lock and removing the singleton pattern on `EspOta` - are safe - and that's one big 'if' - you end up with something like (excuse the possible syntax typos, I'm on a vacation and time is precious):

```rust
let ota = EspOta;
// do various "read only" stuff with `ota`
// ...
ota.into_update()?; // This MIGHT fail at runtime, if you try to do this on another `ota` instance, in the same or in a different thread
// ...
```

As opposed to the current status quo:
```rust
let mut ota = EspOta::take()?; // This MIGHT fail at runtime if the singleton object is already taken elsewhere
// do various "read only" stuff with `ota`
// ...
let update = ota.initiate_update()?; // Will always succeed 
// ...
let update2 = ota.initiate_update()?; // Will always fail compile-time unless you have already dropped `update` 
```

So what gives? You are shifting the lock elsewhere and you don't have to "thread" the `ota` var throughout all the code that needs to do a R/O OTA operations? Is this that much of an ergonomics improvement (again - assuming it is OK to implement) in the first place?

Also: you seem to have forgotten about:
- `factory_reset`
- `mark_running_slot_invalid_and_reboot`
- `mark_running_slot_valid`

All of the ^^^ take a `&mut EspOta` reference. With your de-singleton-ification, I can now call these ops from, **separate threads** and they would be data-racing each other. Would that be safe? Or we would be making these thread safe too?

On (2):

By abandoning the singleton pattern, you are basically proclaiming that / relying upon the ESP OTA code to be *thread safe* - at least in its "read only" part (as I can now call the EspOta R/O API from multiple threads). This is often a false assumption with C code, as the fact that this code is only doing "read only" "innocent" operations does not mean it is thread safe, as it might have a **hidden unsafe interior mutability**. Like [this one](https://github.com/espressif/esp-idf/blob/master/components/app_update/esp_ota_ops.c#L546). Sure, this is just a pointer, but did you check that pointer load/store on xtensa variants are atomic, including future variants (on riscv32imc I *think* it is)? Further, the C code has no sign of any mutexes / critical sections usage. It has other static variables. It does not have any "thread safety" statement in the ESP IDF docu (that is, that I was able to google at the time). Most importantly, this is your **app self-update code**. The last place where we should be taking risks, right?

So yeah, given the circumstances, I think the singleton pattern - even for `EspOta` - should stay. I hope you also realize `EspOta` is `Send` but **not** `Sync` (hence not thread safe). That's of course on purpose.

Hope these details help clarify my concerns. :)

ivmarkov:
> One thing that should be addressed under all circumstances, is that a null pointer can be returned if no partition can be found. This behavior should be caught and a error should be thrown.

If you have experienced such a bug, that's a great thing to file a PR for :)

Nukesor:
First of all, I'm sorry for disturbing your holiday! 

Your last post was indeed the missing piece to understand your concerns. I'm pretty new to this kind of low-level programming and obviously didn't think about all possible impacts of this change.

In the meantime, I tried to implement the nested loop pattern.
In message handling code, it's a very common (and in my opinion good) pattern, to have single large enums that contain all possible messages that can be sent.
These are then handled in a single large match clause, which simply calls handler functions in most of the cases. The main benefit of doing this in a single match clause is, that we can be 100% sure that we handled all possible message variants.
Splitting this up into several functions, changes unhandled messages from a compiletime error to a runtime error, which is frankly a no-go.

As a result, the way to implement this, is to virtually copy these message handling `match` statements, and adjust the calls where a reference to `EspOta` was previously passed to.

In our Codebase, it resulted in additional ~500 lines of code, with most of it being duplicate code.
This could be shrunk to about ~100 lines of code, if we would give up compile-time guarantees and live with runtime errors. 

I still think, that it's worth improving this case, as OTA updates can take a while and handling two seperate mutable references to the same object for a long time is always a bit tricky.

If it would be ok for you, I'll go back to your original proposal, which I should have picked up earlier.

> If not, just expose into_update(self) onto EspOta itself and then - onto EspOtaUpdate - something like into_ota_complete and into_ota_abort

Im sorry if I had the appearance of being rude, this wasn't my intention.
I simply wasn't getting the whole picture.

src/ota.rs
ivmarkov: Just pick something appropriate from the `ESP_ERR_*` error constants, and return it wrapped as an `EspError` instance. This way we will ensure, that `initiate_update` will complete successfully only if there was an update partition found.

src/ota.rs
```diff
@@ -161,7 +164,7 @@ impl EspOtaUpdate {
 }
 
 #[derive(Debug)]
-pub struct EspOta(EspOtaUpdate);
+pub struct EspOta(Option<EspOtaUpdate>);
```
ivmarkov, line 167: Can you elaborate why introducing `Option` here is necessary? It increases the payload of the `EspOta` struct (not that this is the end of the world), yet I'm not sure what problem it is addressing?

src/ota.rs
```diff
@@ -147,6 +149,7 @@ impl EspOtaUpdate {
 
         self.update_partition = core::ptr::null();
         self.update_handle = 0;
+        self.discarded = true;
```
ivmarkov, line 152: I don't think this field is necessary as `discarded` <=> not in update mode <=> `update_partition` != `null`.

src/ota.rs
```diff
@@ -220,16 +220,43 @@ impl EspOta {
...
+        // "indicates invalid OTA data partition, or that no eligible OTA app slot partition was found."
+        //
+        // This should probably be handled in here and throw a proper EspError.
+        let update_partition = unsafe { esp_ota_get_next_update_partition(ptr::null()) };
 
-        let mut handle: esp_ota_handle_t = Default::default();
+        let mut update_handle: esp_ota_handle_t = Default::default();
 
-        esp!(unsafe { esp_ota_begin(partition, OTA_SIZE_UNKNOWN as usize, &mut handle) })?;
+        esp!(unsafe {
```
ivmarkov, line 232: Return back the original code, as we would be dropping the `Option` stuff from above.

src/ota.rs
ivmarkov: I think - and as per above - we need a much much simpler change - which is - if `esp_ota_get_next_update_partition` is returning a null ptr, `initiate_update` should just fail returning an `EspError`.

I'm unsure as to what is the special meaning of a default null pointer, if you could elaborate?

src/ota.rs
```diff
@@ -220,16 +220,43 @@ impl EspOta {
...
+        });
 
-        self.0.update_partition = partition;
-        self.0.update_handle = handle;
+        Ok(self.0.as_mut().unwrap())
+    }
+
+    /// Get a handle to the [EspOtaUpdate].
+    /// This is `None`, if [Self.initiate_update] hasn't been called yet.
+    pub fn get_update(&mut self) -> Option<&mut EspOtaUpdate> {
```
ivmarkov, line 251: Check if `update_partition` is not null and then return `Some(...)` else `None`.

ivmarkov (changes requested):

src/ota.rs
ivmarkov: `ESP_ERR_NOT_SUPPORTED` and `ESP_ERR_INVALID_STATE` are both good candidates to return as an error code if you get a null pointer from ESP IDF. Perhaps just pick one of these two and return it.

ivmarkov (changes requested):

Nukesor:
Thanks for the review, but I didn't get to updating the code to your `into` proposal yet :)
I'm currently sick at home and will probably only start working on this next week.

ivmarkov:
> Thanks for the review, but I didn't get to updating the code to your `into` proposal yet :) I'm currently sick at home and will probably only start working on this next week.

Indeed, but now that I looked into the details of what you actually suggest as a PR, I'm starting to think that it might be as good as the `into` idea:
* There is no sign of moving the lock to `EspOtaUpdate` so I guess this was just a brain-storming suggestion;
* `EspOta` is still a singleton;
* Your `get_update()` method allows you to optionally return a mut ref to `EspOtaUpdate` if `EspOta` was switched to updating mode by `initiate_update`

So I guess I'm saying I'm fine with this approach as well! The review feedback can be summarized as "implement a few simplifications and reduce the change surface, and we can merge this too".

ivmarkov:
Oh and it is all still completely backwards compatible. That is, if we put aside the fact that `initiate_update` will now error out in case it cannot find a valid OTA partition; but I rather consider the current behavior of `initiate_update` a bug (where it does not return an error on invalid OTA partition).

src/ota.rs
Nukesor: As it turns out, this isn't even necessary, as a null pointer will trigger an error when provided to [esp_ota_begin](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/ota.html#_CPPv413esp_ota_beginPK15esp_partition_t6size_tP16esp_ota_handle_t).
The current logic is sound and I added an inline comment to prevent further confusion

Nukesor:
Once again, thanks for your review!

tldr: The change is now very minimalistic as proposed by your review and works perfectly fine in our usecase.

I went back to the drawing board and took a closer look at my initial proposal and your proposed fixes.
You were completely right, simply adding a `get_update` function works perfectly fine. There's no need for an `Option` or flags.

If I remember correctly, I added those because I thought that it was somehow possible to end up with a null pointer on the `EspOtaUpdate` struct even though the struct was still initialized.
Taking a closer look at the error handling logic, this is not the case, which makes this whole thing a lot easier. Sorry for all of the confusion and unnecessary discussions, I definitely learned quite a bit while working on this :laughing: .

ivmarkov:
It seems we have exceeded the GitHub rate limit, so CI does not run. But the change is so minimal, that I'll just merge. Thanks!

PR #210, opened 2023-01-06
X509 should have same lifetime as MqttClientConfiguration

swanandx:
certificates do not need to live for `'static` lifetime. As other fields like `private_key_password` live for lifetime of our config, same should be with `X509`.

`'static` lifetime gets in our way when we have to parse the certs at runtime, e.g. from a JSON file. Changing it to have same lifetime as `MqttClientConfiguration` should make it much easier for users.

### DIFF
--- a/src/mqtt/client.rs
+++ b/src/mqtt/client.rs
@@ -75,10 +75,10 @@ pub struct MqttClientConfiguration<'a> {
     #[cfg(not(esp_idf_version = "4.3"))]
     pub crt_bundle_attach: Option<unsafe extern "C" fn(conf: *mut c_void) -> esp_err_t>,
 
-    pub server_certificate: Option<X509<'static>>,
+    pub server_certificate: Option<X509<'a>>,
 
-    pub client_certificate: Option<X509<'static>>,
-    pub private_key: Option<X509<'static>>,
+    pub client_certificate: Option<X509<'a>>,
+    pub private_key: Option<X509<'a>>,
     pub private_key_password: Option<&'a str>,
     // TODO: Future
     // pub psk_hint_key: KeyHint,


ivmarkov:
This can only work if you parameterize the MQTT client wrapper over the `'a` lifetime as well. Also, if you take the effort to actually look into the ESP IDF MQTT  C implementation, you'll realize that your analogy between the password field and the certificate fields is wrong g as well. The password field (being a short one) is `memcpy`d in a memory held by the MQTT client, so for this field having `'a` living as long as `&self` during the client co struction is OK. Not so with the certificates, which are bigger and not memcpy'd yet need to be alive as long as the client is alive.

Rather than introducing lifetime in the mqtt client (which is possible but backwards incompatible and will require some serious thinking on your side), perhaps you can look I to the static-cell crate and co sidereal if you can apply it for your use case. While I realize you are parsing at runtime.e (btw - why? It is as inefficient as it gets), you might co sidereal putting the parsed certificate in a static context after parsing via static-cell.

swanandx:
Hey, yeah that analogy might be wrong haha. I wanted it to have `'a` lifetime so I can parse them from JSON file.
I am writing an library ( sdk ) which uses mqtt to communicate with bytebeam cloud platform. So I was thinking, users will flash them to SPIFFS partitions of esp separately if they want to use sdk. We will just directly read from SPIFFS.
If I was writing a bin application, I can parse them in build script and include them in as static, but that was not case here. Also it will increase the firmware size as well right?

ivmarkov:
Of course static-cell will consume spam, but you don't have a choice with your json spiffs parsing strategy anyway. And this will avoid you having to parameterize the mqtt client over 'a lifetime.

swanandx:
> Did you look into static-cell?
I did! it looks like what I was looking for.
Thanks for your feedback, I will definitely explore more about the other alternative! I should have put more thought about backwards compatibility before opening PR 😅

ivmarkov:
Np. Feel free to reopen / write an issue request if the static-cell approach does not work. But that would mean a more intrusive change then of course, touching the client struct.

PR #214, opened 2023-01-24
added nvs get/set for primitive types

### LINKED ISSUE #206: Current nvs_get_blob() based NVS access can't read NVS partition created with IDF nvs_partition_gen.py
I use the IDF nvs_partition_gen.py tool to generate the nvs partition content using the following definition file:
```
key,type,encoding,value
certificates,namespace,,
ca_cert,file,string,../../certificates/aws-iot-rootCA.crt
cert,file,string,../../certificates/cert.pem
priv_key,file,string,../../certificates/privkey.pem
device_data,namespace,,
device_id,data,string,XXXXX-YYY-ZZZZZ
```
When checking access to the data using ```EspCustomNvs::contains()``` none of the keys are found. When replacing in ```EspNvs::len()``` ```nvs_get_blob()```  with ```nvs_get_str()``` the keys are found. While the IDF documentation states that both functions have a identical behavior except that they return different data types, there seems to be a difference.

I would propose to add additional functions which use ```nvs_get_str()``` / ```nvs_set_str()``` for handling access to string based key/value pairs. I'm open to create a PR to resolve the issue.

taunusflieger:
This PR resolves #206. It contains a set of set / get functions for Rust primitive types to access NVS. With this PR ```esp-idf-svc``` is closer to the native IDF NVS functionality.

```get_str()``` is special, as it does not return a string, but a slice of u8. The reason is that it should be usable for heap/heapless scenarios. The conversion into the destination string type is left for the user of this function.

### DIFF
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -353,32 +353,43 @@ impl<T: NvsPartitionId> EspNvs<T> {
         Ok(true)
     }
 
-    pub fn get_blob<'a>(
-        &self,
-        name: &str,
-        buf: &'a mut [u8],
-    ) -> Result<Option<&'a [u8]>, EspError> {
+    pub fn blob_len(&self, name: &str) -> Result<Option<usize>, EspError> {
         let c_key = CString::new(name).unwrap();
 
-        // check for blob value, by getting blob length
+        #[allow(unused_assignments)]
         let mut len = 0;
+
         match unsafe { nvs_get_blob(self.1, c_key.as_ptr(), ptr::null_mut(), &mut len as *mut _) } {
             ESP_ERR_NVS_NOT_FOUND => Ok(None),
             err => {
                 // bail on error
                 esp!(err)?;
 
-                len = buf.len();
+                Ok(Some(len))
+            }
+        }
+    }
 
-                // fetch value if no error
-                esp!(unsafe {
-                    nvs_get_blob(
-                        self.1,
-                        c_key.as_ptr(),
-                        buf.as_mut_ptr() as *mut _,
-                        &mut len as *mut _,
-                    )
-                })?;
+    pub fn get_blob<'a>(
+        &self,
+        name: &str,
+        buf: &'a mut [u8],
+    ) -> Result<Option<&'a [u8]>, EspError> {
+        let c_key = CString::new(name).unwrap();
+        let mut len = buf.len();
+
+        match unsafe {
+            nvs_get_blob(
+                self.1,
+                c_key.as_ptr(),
+                buf.as_mut_ptr() as *mut _,
+                &mut len as *mut _,
+            )
+        } {
+            ESP_ERR_NVS_NOT_FOUND => Ok(None),
+            err => {
+                // bail on error
+                esp!(err)?;
 
                 Ok(Some(&buf[..len]))
             }
@@ -418,10 +429,8 @@ impl<T: NvsPartitionId> EspNvs<T> {
     pub fn get_str<'a>(&self, name: &str, buf: &'a mut [u8]) -> Result<Option<&'a str>, EspError> {
         let c_key = CString::new(name).unwrap();
 
-        #[allow(unused_assignments)]
-        let mut len = 0;
+        let mut len = buf.len();
         match unsafe {
-            len = buf.len();
             nvs_get_str(
                 self.1,
                 c_key.as_ptr(),
@@ -583,6 +592,7 @@ impl<T: NvsPartitionId> EspNvs<T> {
     pub fn get_i32(&self, name: &str) -> Result<Option<i32>, EspError> {
         let c_key = CString::new(name).unwrap();
         let mut result: [i32; 1] = [0; 1];
+
         match unsafe { nvs_get_i32(self.1, c_key.as_ptr(), &mut result[0] as *mut _) } {
             ESP_ERR_NVS_NOT_FOUND => Ok(None),
             err => {
@@ -607,6 +617,7 @@ impl<T: NvsPartitionId> EspNvs<T> {
     pub fn get_u64(&self, name: &str) -> Result<Option<u64>, EspError> {
         let c_key = CString::new(name).unwrap();
         let mut result: [u64; 1] = [0; 1];
+
         match unsafe { nvs_get_u64(self.1, c_key.as_ptr(), &mut result[0] as *mut _) } {
             ESP_ERR_NVS_NOT_FOUND => Ok(None),
             err => {
@@ -631,6 +642,7 @@ impl<T: NvsPartitionId> EspNvs<T> {
     pub fn get_i64(&self, name: &str) -> Result<Option<i64>, EspError> {
         let c_key = CString::new(name).unwrap();
         let mut result: [i64; 1] = [0; 1];
+
         match unsafe { nvs_get_i64(self.1, c_key.as_ptr(), &mut result[0] as *mut _) } {
             ESP_ERR_NVS_NOT_FOUND => Ok(None),
             err => {
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -353,7 +353,52 @@ impl<T: NvsPartitionId> EspNvs<T> {
         Ok(true)
     }
 
-    pub fn len_str(&self, name: &str) -> Result<Option<usize>, EspError> {
+    pub fn get_blob<'a>(
+        &self,
+        name: &str,
+        buf: &'a mut [u8],
+    ) -> Result<Option<&'a [u8]>, EspError> {
+        let c_key = CString::new(name).unwrap();
+
+        // check for blob value, by getting blob length
+        let mut len = 0;
+        match unsafe { nvs_get_blob(self.1, c_key.as_ptr(), ptr::null_mut(), &mut len as *mut _) } {
+            ESP_ERR_NVS_NOT_FOUND => Ok(None),
+            err => {
+                // bail on error
+                esp!(err)?;
+
+                len = buf.len();
+
+                // fetch value if no error
+                esp!(unsafe {
+                    nvs_get_blob(
+                        self.1,
+                        c_key.as_ptr(),
+                        buf.as_mut_ptr() as *mut _,
+                        &mut len as *mut _,
+                    )
+                })?;
+
+                Ok(Some(&buf[..len]))
+            }
+        }
+    }
+
+    pub fn set_blob(&mut self, name: &str, buf: &[u8]) -> Result<(), EspError> {
+        let c_key = CString::new(name).unwrap();
+
+        // start by just clearing this key
+        unsafe { nvs_erase_key(self.1, c_key.as_ptr()) };
+
+        esp!(unsafe { nvs_set_blob(self.1, c_key.as_ptr(), buf.as_ptr().cast(), buf.len()) })?;
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(())
+    }
+
+    pub fn str_len(&self, name: &str) -> Result<Option<usize>, EspError> {
         let c_key = CString::new(name).unwrap();
 
         #[allow(unused_assignments)]
@@ -370,7 +415,7 @@ impl<T: NvsPartitionId> EspNvs<T> {
         }
     }
 
-    pub fn get_str<'a>(&self, name: &str, buf: &'a mut [u8]) -> Result<Option<&'a [u8]>, EspError> {
+    pub fn get_str<'a>(&self, name: &str, buf: &'a mut [u8]) -> Result<Option<&'a str>, EspError> {
         let c_key = CString::new(name).unwrap();
 
         #[allow(unused_assignments)]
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/nvs.rs [hunk 1]
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+            err => {
+                // bail on error
+                esp!(err)?;
+
+                Ok(Some(len))
+            }
+        }
+    }
+
+    pub fn get_str<'a>(&self, name: &str, buf: &'a mut [u8]) -> Result<Option<&'a [u8]>, EspError> {
```
ivmarkov, line 373: Ideally, this should return an `&'a str`, not `&'a [u8]`. You can go from `&'a [u8]` to `&'a str` by using `core::str::from_utf8_unchecked`.

ivmarkov: By the way... given that the get_XXX / set_XXX methods are *not* trait implementations, and that inside them we are full of allocations (you know, the `CString` stuff), we can just as well surrender to `alloc` and require the `alloc` module for this API. (And not compile this code if the `alloc` feature is not enabled.)

This would allow you to use `String` and `Vec`, rather than suffering with the `&[u8]` bufs.

Maybe food for thought. Not certain myself which way to go.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+        unsafe { nvs_erase_key(self.1, c_key.as_ptr()) };
+
+        esp!(unsafe { nvs_set_str(self.1, c_key.as_ptr(), c_val.as_ptr(),) })?;
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
+    }
+
+    pub fn get_u8<'a>(&self, name: &str, out_val: &'a mut u8) -> Result<Option<&'a u8>, EspError> {
```
ivmarkov, line 411: `u8` is `Copy` so having the user to provide an `out_val` buf is unnecessary. Just return the `u8` by value, and use a `[u8; 1]` array allocated on the stack by the function itself.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+        let c_val = CString::new(val).unwrap();
+
+        // start by just clearing this key
+        unsafe { nvs_erase_key(self.1, c_key.as_ptr()) };
+
+        esp!(unsafe { nvs_set_str(self.1, c_key.as_ptr(), c_val.as_ptr(),) })?;
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
```
ivmarkov, line 408: The returned `bool` does not seem to indicate anything, remove.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+    }
+
+    pub fn set_u8(&self, name: &str, val: u8) -> Result<bool, EspError> {
+        let c_key = CString::new(name).unwrap();
+
+        esp!(unsafe { nvs_set_u8(self.1, c_key.as_ptr(), val) })?;
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
```
ivmarkov, line 432: The returned `bool` does not seem to indicate anything, remove.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+    }
+
+    pub fn get_i8<'a>(&self, name: &str, out_val: &'a mut i8) -> Result<Option<&'a i8>, EspError> {
+        let c_key = CString::new(name).unwrap();
```
ivmarkov, line 436: Ditto - get rid of the `out_val` buf.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+    }
+
+    pub fn set_u16(&self, name: &str, val: u16) -> Result<bool, EspError> {
+        let c_key = CString::new(name).unwrap();
+
+        esp!(unsafe { nvs_set_u16(self.1, c_key.as_ptr(), val) })?;
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
```
ivmarkov, line 482: The returned `bool` does not seem to indicate anything, remove.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+    }
+
+    pub fn set_i16(&self, name: &str, val: i16) -> Result<bool, EspError> {
+        let c_key = CString::new(name).unwrap();
+
+        esp!(unsafe { nvs_set_i16(self.1, c_key.as_ptr(), val) })?;
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
```
ivmarkov, line 509: The returned `bool` does not seem to indicate anything, remove.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+    }
+
+    pub fn set_u32(&self, name: &str, val: u32) -> Result<bool, EspError> {
+        let c_key = CString::new(name).unwrap();
+
+        esp!(unsafe { nvs_set_u32(self.1, c_key.as_ptr(), val) })?;
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
```
ivmarkov, line 536: The returned `bool` does not seem to indicate anything, remove.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+    }
+
+    pub fn set_i64(&self, name: &str, val: i64) -> Result<bool, EspError> {
+        let c_key = CString::new(name).unwrap();
+
+        esp!(unsafe { nvs_set_i64(self.1, c_key.as_ptr(), val) })?;
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
```
ivmarkov, line 617: The returned `bool` does not seem to indicate anything, remove.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+    }
+
+    pub fn set_u64(&self, name: &str, val: u64) -> Result<bool, EspError> {
+        let c_key = CString::new(name).unwrap();
+
+        esp!(unsafe { nvs_set_u64(self.1, c_key.as_ptr(), val) })?;
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
```
ivmarkov, line 590: The returned `bool` does not seem to indicate anything, remove.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+    }
+
+    pub fn set_i32(&self, name: &str, val: i32) -> Result<bool, EspError> {
+        let c_key = CString::new(name).unwrap();
+
+        esp!(unsafe { nvs_set_i32(self.1, c_key.as_ptr(), val) })?;
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
```
ivmarkov, line 563: The returned `bool` does not seem to indicate anything, remove.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+    }
+
+    pub fn get_u16<'a>(
+        &self,
+        name: &str,
+        out_val: &'a mut u16,
```
ivmarkov, line 461: Ditto - get rid of the `out_val` buf.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
+    }
+
+    pub fn get_i16<'a>(
+        &self,
+        name: &str,
+        out_val: &'a mut i16,
```
ivmarkov, line 488: Ditto - get rid of the `out_val` buf.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
+    }
+
+    pub fn get_u32<'a>(
+        &self,
+        name: &str,
+        out_val: &'a mut u32,
```
ivmarkov, line 515: Ditto - get rid of the `out_val` buf.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
+    }
+
+    pub fn get_i32<'a>(
+        &self,
+        name: &str,
+        out_val: &'a mut i32,
```
ivmarkov, line 542: Ditto - get rid of the `out_val` buf.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
+    }
+
+    pub fn get_u64<'a>(
+        &self,
+        name: &str,
+        out_val: &'a mut u64,
```
ivmarkov, line 569: Ditto - get rid of the `out_val` buf.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(true)
+    }
+
+    pub fn get_i64<'a>(
+        &self,
+        name: &str,
+        out_val: &'a mut i64,
```
ivmarkov, line 596: Ditto - get rid of the `out_val` buf.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
 
         Ok(true)
     }
+
+    pub fn len_str(&self, name: &str) -> Result<Option<usize>, EspError> {
```
ivmarkov, line 356: Maybe `str_len`? But then also see my comments on whether we should surrender to `alloc` for this API.

src/nvs.rs
```diff
@@ -352,6 +352,270 @@ impl<T: NvsPartitionId> EspNvs<T> {
 
         Ok(true)
     }
+
```
ivmarkov, line 355: I don't see a get_blob/set_blob API. I think we need it as well.

ivmarkov (changes requested):

ivmarkov:
> This PR resolves #206. It contains a set of set / get functions for Rust primitive types to access NVS. With this PR `esp-idf-svc` is closer to the native IDF NVS functionality.
> 
> `get_str()` is special, as it does not return a string, but a slice of u8. The reason is that it should be usable for heap/heapless scenarios. The conversion into the destination string type is left for the user of this function.

Going from `&[u8]` to `&str` is still possible without alloc (with some caveats). Details below. `str` is unsized just like `[u8]` and `&str` is a fat pointer, just like `&[u8]`.

src/nvs.rs
taunusflieger: I think keeping the function makes sense independent of the alloc topic

src/nvs.rs
taunusflieger: I tend to lean towards an approach which does not require heap allocation. There might be usage scenarios where reading for example config values from the nvs happens often (no app side caching of the values). For such a scenario having heap allocation might not be optimal. I would prefere stack alloc (not to fragment the heap over time too much). What do you think?

src/nvs.rs
```diff
@@ -352,6 +352,305 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+    pub fn get_blob<'a>(
```
ivmarkov, line 356: You also need `blob_len`, similarly to `str_len`.

src/nvs.rs
```diff
@@ -352,6 +352,305 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+        &self,
+        name: &str,
+        buf: &'a mut [u8],
+    ) -> Result<Option<&'a [u8]>, EspError> {
+        let c_key = CString::new(name).unwrap();
+
+        // check for blob value, by getting blob length
```
ivmarkov, line 363: I don't think you need to call `nvs_get_blob` twice. Perhaps you copied from the current Rust-only impl? Just follow the same logic as in `get_str`.

src/nvs.rs
```diff
@@ -352,6 +352,305 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+        }
+    }
+
+    pub fn get_str<'a>(&self, name: &str, buf: &'a mut [u8]) -> Result<Option<&'a str>, EspError> {
+        let c_key = CString::new(name).unwrap();
+
+        #[allow(unused_assignments)]
+        let mut len = 0;
+        match unsafe {
+            len = buf.len();
```
ivmarkov, line 424: Nit: merge `len = buf.len()` with `let mut len = 0;`. There is no reason why the former should be in the `unsafe` block, and separate from the initializer. This will allow you to get rid of the `unused-assigments` warning, which is more or less telling you exactly that.

ivmarkov (changes requested):
Code looks OK. Hopefully a final set of comments and then we can merge.

taunusflieger:
@ivmarkov thanks for your support!

ivmarkov:
There are some really small nits remaining, like I just noticed that you are using arrays for `u8`, `i8`, `u6` etc., which is unnecessary. Also you have `allow(unused_assignments)` at two places which is also probably no longer necessary. But these are minor, so I'll merge. Thanks.

PR #218, opened 2023-01-28
make set_target_level() available and fix should_log() to use target not hard coded `rust-logging`

liebman:
- log::EspLogger::set_target_level make function not method
    Made this a function so that it can be called via log::EspLogger::set_target_level()
- log::EspLogger::should_log implement cache for CStrings
    so it turns out that  esp-idf `function esp_log_level_get` builds a cache using the address
    of the target and not doing a string compare.  This means we need to
    build a cache of our own mapping the target string value to a consistant 
    CString value.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,6 @@ esp-idf-sys = { version = "0.32.1", default-features = false, features = ["nativ
 esp-idf-hal = { version = "0.40", default-features = false, features = ["esp-idf-sys"] }
 embassy-sync = { version = "0.1", optional = true }
 embassy-time = { version = "0.1", optional = true, features = ["tick-hz-1_000_000"] }
-once_cell = "1.17.0"
 
 [build-dependencies]
 embuild = "0.31"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ esp-idf-sys = { version = "0.32.1", default-features = false, features = ["nativ
 esp-idf-hal = { version = "0.40", default-features = false, features = ["esp-idf-sys"] }
 embassy-sync = { version = "0.1", optional = true }
 embassy-time = { version = "0.1", optional = true, features = ["tick-hz-1_000_000"] }
+once_cell = "1.17.0"
 
 [build-dependencies]
 embuild = "0.31"
--- a/src/log.rs
+++ b/src/log.rs
@@ -99,7 +99,7 @@ impl EspLogger {
         LevelFilter::from(Newtype(CONFIG_LOG_MAXIMUM_LEVEL))
     }
 
-    pub fn set_target_level(&self, target: impl AsRef<str>, level_filter: LevelFilter) {
+    pub fn set_target_level(target: impl AsRef<str>, level_filter: LevelFilter) {
         let ctarget = CString::new(target.as_ref()).unwrap();
 
         unsafe {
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -34,6 +34,7 @@ esp-idf-sys = { version = "0.32.1", default-features = false, features = ["nativ
 esp-idf-hal = { version = "0.40", default-features = false, features = ["esp-idf-sys"] }
 embassy-sync = { version = "0.1", optional = true }
 embassy-time = { version = "0.1", optional = true, features = ["tick-hz-1_000_000"] }
+once_cell = "1.17.0"
```
ivmarkov, line 37: This introduces a hard dependency on a STD-only crate so that won't fly. See below why we don't need it anyway.

src/log.rs
```diff
@@ -139,10 +139,28 @@ impl EspLogger {
...
     fn should_log(record: &Record) -> bool {
+        use crate::private::mutex::Mutex;
+        use alloc::collections::BTreeMap;
+        use once_cell::sync::Lazy;
+
+        // esp-idf `function esp_log_level_get` builds a cache using the address
+        // of the target and not doing a string compare.  This means we need to
+        // build a cache of our own mapping the string value to a consistant
+        // c-string value.
+        static TARGET_CACHE: Lazy<Mutex<BTreeMap<String, CString>>> = Lazy::new(|| Mutex::new(BTreeMap::new()));
```
ivmarkov, line 150: Why do you need `Lazy` in the first place? `Mutex::wrap(RawMutex::new(), BTreeMap::new())` is a `const` expr, so you can just assign it to the static.

src/log.rs
```diff
@@ -139,10 +139,28 @@ impl EspLogger {
...
         let level = Newtype::<esp_log_level_t>::from(record.level()).0;
-        let max_level = unsafe {
-            esp_log_level_get(b"rust-logging\0" as *const u8 as *const _) // TODO: use record target?
+        let mut cache = TARGET_CACHE.lock();
+        let ctarget = match cache.get(record.target()) {
```
ivmarkov, line 153: Has this passed `cargo fmt`?

src/log.rs
```diff
@@ -139,10 +139,28 @@ impl EspLogger {
...
+            Some(value) => value,
+            None => {
+                // println!("EspLogger::should_log: add to cache target={}", record.target());
```
ivmarkov, line 156: Remove pls.

src/log.rs
```diff
@@ -139,10 +139,28 @@ impl EspLogger {
...
+                let value = CString::new(record.target()).unwrap();
```
ivmarkov, line 157: I think there was a way to do the insert + get with a single map op.

src/log.rs
```diff
@@ -139,10 +139,28 @@ impl EspLogger {
...
+                cache.insert(record.target().to_string(), value);
+                cache.get(record.target()).unwrap()
+            }
         };
+        let max_level = unsafe { esp_log_level_get(ctarget.as_c_str().as_ptr()) };
+        // println!("EspLogger::should_log: target={} max_level={}", record.target(), max_level);
```
ivmarkov, line 163: Remove pls.

ivmarkov (changes requested):

src/log.rs
```diff
@@ -99,7 +99,7 @@ impl EspLogger {
         LevelFilter::from(Newtype(CONFIG_LOG_MAXIMUM_LEVEL))
     }
 
-    pub fn set_target_level(&self, target: impl AsRef<str>, level_filter: LevelFilter) {
+    pub fn set_target_level(target: impl AsRef<str>, level_filter: LevelFilter) {
```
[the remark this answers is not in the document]
liebman: Ok, I'll put it back.  Any thoughts on how to access it since LOGGER is static? What about a `pub fn` local to this module that calls `LOGGER.set_target_level()`?

src/log.rs
liebman: found it

src/log.rs
```diff
@@ -139,10 +139,21 @@ impl EspLogger {
...
     #[cfg(not(all(esp_idf_version_major = "4", esp_idf_version_minor = "3")))]
     fn should_log(record: &Record) -> bool {
+        use crate::private::mutex::{Mutex, RawMutex};
+        use alloc::collections::BTreeMap;
+
+        // esp-idf `function esp_log_level_get` builds a cache using the address
+        // of the target and not doing a string compare.  This means we need to
+        // build a cache of our own mapping the string value to a consistant
+        // c-string value.
+        static TARGET_CACHE: Mutex<BTreeMap<String, CString>> =
```
ivmarkov, line 149: `alloc::string::String` or import the `alloc::string` module.

ivmarkov (changes requested):

PR #227, opened 2023-02-14
Hide wifi password from info prints

ydayan:
Printing the Wi-Fi password in plain-text over UART is a security risk.
We didn't want to prevent others from printing it if they really want to (hence we didn't override the Debug implementation), so we found this to be the simplest solution

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -1348,7 +1348,7 @@ impl WifiWait {
     }
 }
 
-fn format_config_without_password(config: &Configuration) -> String {
+fn format_config_without_password(config: &Configuration) -> alloc::string::String {
     match config {
         Configuration::None => format!("{config:?}"),
         Configuration::Client(client) => format_client_config_without_password(client),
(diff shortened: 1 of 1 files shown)


MabezDev:
This feels somewhat pointless, logs should probably be disabled for a production device. However, I don't feel too against the idea of hiding the password by default, but I do feel strongly about the implementation of it, we should not be pulling in alloc just to hide the password. 

What should be done instead is customizing the `Debug` impl inside `embedded-svc` to not print the password. Please open a PR with the changes there.

ydayan:
We do want to have some logs during production. If it's not met to be shown in production build as you say, shouldn't these prints be in debug level? Would that make sense? @MabezDev

MabezDev:
Even at debug level, they could be leaked in the future.

> What should be done instead is customizing the Debug impl inside embedded-svc to not print the password. Please open a PR with the changes there.

You should do this if you want to properly hide the password. Instead of this debug derive: https://github.com/esp-rs/embedded-svc/blob/85c32e9b8cdd903593bfd64bac4a6091f73c3964/src/wifi.rs#L184, you would do a custom implementation that doesn't print the password.

Issue #176, opened 2022-11-24
EthDriver not working with dma disabled


teamplayer3:
When I use the `EthDriver` with `Dma::Disabled`, an error as follows is logged and for example a ping does not succeed.

```
I (2474) esp_idf_svc::ping: Ping timeout callback invoked
I (2474) esp_idf_svc::ping: From ??? icmp_seq=2 timeout
E (2854) spi_master: check_trans_valid(689): txdata transfer > host maximum
E (2854) w5500.mac: w5500_write(74): spi transmit failed
E (2864) w5500.mac: w5500_write_buffer(173): write TX buffer failed
E (2864) w5500.mac: emac_w5500_transmit(528): write frame failed
```

Related to 790fee01f9b70743dcfa46724b7ed34943832c66.

Vollbrecht:
Dma::disabled imply that the frame_size is 64byte most Eth Frames with an MTU of 1500 or so are properbly to large here. If an Frame is longer than 64 bytes it can only be send in chucks or need Dma

teamplayer3:
Ah ok, good to know. So, info like this belongs to docs at the end, I guess? Or is this knowledge which should be obtained from the esp-idf docs.

Vollbrecht:
yeah thats documented in esp-idf doc. usually this is not a problem though because the esp-idf-hal driver should create chunks that are not larger than this  and send it chunk by chunk

reinismu:
@teamplayer3 managed to get it working? 

Now struggling myself with ETH w5500.

```
let eth = w5500::init_w5500(
        &sysloop,
        Box::new(EspEth::wrap(EthDriver::new_spi(
            peripherals.spi2,
            pins.gpio4, // int
            pins.gpio12, // sclk
            pins.gpio13, // sdo
            pins.gpio11, // sdi
            esp_idf_hal::spi::Dma::Auto(1500), // esp_idf_hal::spi::Dma::Disabled
            Some(pins.gpio10), // cs
            Option::<gpio::Gpio28>::None, //Some(pins.gpio39), // rst
            SpiEthChipset::W5500,
            36.MHz().into(),
            Some(&[0x02, 0x00, 0x00, 0x12, 0x34, 0x56]),
            Some(1),
            sysloop.clone(),
        )?)?),
    )?;
```
Basic example from `esp-idf` works. 
At first it fails of not getting HW id. and then after restart I get

```
Start w5500 mdns!
[...]

PR #236, opened 2023-03-08
Use SpiDriver to setup ETH

reinismu:
This should allow to share ETH with other devices on the same SPI bus

### DIFF
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -295,13 +295,9 @@ impl<'d> EthDriver<'d> {
     esp_idf_eth_spi_ethernet_ksz8851snl
 ))]
 impl<'d> EthDriver<'d> {
-    pub fn new_spi<P: spi::Spi + 'd>(
-        spi: impl Peripheral<P = P> + 'd,
+    pub fn new_spi(
+        spi_driver: &spi::SpiDriver<'d>,
         int: impl Peripheral<P = impl gpio::InputPin> + 'd,
-        sclk: impl Peripheral<P = impl gpio::OutputPin> + 'd,
-        sdo: impl Peripheral<P = impl gpio::OutputPin> + 'd,
-        sdi: impl Peripheral<P = impl gpio::InputPin + gpio::OutputPin> + 'd,
-        dma: esp_idf_hal::spi::Dma,
         cs: Option<impl Peripheral<P = impl gpio::OutputPin> + 'd>,
         rst: Option<impl Peripheral<P = impl gpio::OutputPin> + 'd>,
         chipset: SpiEthChipset,
@@ -310,27 +306,23 @@ impl<'d> EthDriver<'d> {
         phy_addr: Option<u32>,
         sysloop: EspSystemEventLoop,
     ) -> Result<Self, EspError> {
-        esp_idf_hal::into_ref!(spi, int, sclk, sdo, sdi);
+        esp_idf_hal::into_ref!(int);
 
-        let (mac, phy, spi_device) = Self::init_spi::<P>(
+        let (mac, phy, spi_device) = Self::init_spi(
+            spi_driver.host(),
             chipset,
             baudrate,
             int.pin(),
-            sclk.pin(),
-            sdo.pin(),
-            sdi.pin(),
-            dma,
             cs.map(|pin| pin.into_ref().pin()),
             rst.map(|pin| pin.into_ref().pin()),
             phy_addr,
         )?;
 
         let eth = Self::init(
-            spi,
             mac,
             phy,
             mac_addr,
-            Some(P::device()),
+            Some(spi_driver.host()),
             spi_device,
             sysloop,
         )?;
@@ -338,14 +330,11 @@ impl<'d> EthDriver<'d> {
         Ok(eth)
     }
 
-    fn init_spi<P: spi::Spi>(
+    fn init_spi(
+        host: spi_host_device_t,
         chipset: SpiEthChipset,
         baudrate: Hertz,
         int: i32,
-        sclk: i32,
-        sdo: i32,
-        sdi: i32,
-        dma: esp_idf_hal::spi::Dma,
         cs: Option<i32>,
         rst: Option<i32>,
         phy_addr: Option<u32>,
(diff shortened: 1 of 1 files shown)


src/eth.rs
```diff
@@ -295,13 +295,9 @@ impl<'d> EthDriver<'d> {
     esp_idf_eth_spi_ethernet_ksz8851snl
```
ivmarkov, line 295: In general: does not build for ESP IDF 5. Please check

src/eth.rs
```diff
@@ -357,7 +346,7 @@ impl<'d> EthDriver<'d> {
         ),
         EspError,
     > {
-        Self::init_spi_bus::<P>(sclk, sdo, sdi, dma)?;
+        unsafe { gpio_install_isr_service(0) };
```
ivmarkov, line 349: Why is this necessary?

src/eth.rs
```diff
@@ -295,13 +295,9 @@ impl<'d> EthDriver<'d> {
     esp_idf_eth_spi_ethernet_ksz8851snl
 ))]
 impl<'d> EthDriver<'d> {
-    pub fn new_spi<P: spi::Spi + 'd>(
-        spi: impl Peripheral<P = P> + 'd,
+    pub fn new_spi(
+        spi_driver: &spi::SpiDriver<'d>,
```
ivmarkov, line 299: The SPI driver should be configured with DMA mode enabled and with transfer size > 1500. Shall we check that?

ivmarkov, line 299: You are re-using the `SpiDriver` thing, but you are not re-using `SpiDeviceDriver`. The latter would allow you to get rid of `init_spi_device` and `get_spi_conf`. Since this is an obtrusive change, why not bite the bullet and pass-in an `SpiDeviceDriver` instance instead?

ivmarkov (changes requested):

src/eth.rs
reinismu: It was there before. Not sure why, but I guess it suppresses this error `E (473) gpio: gpio_isr_handler_add(465): GPIO isr service is not installed, call gpio_install_isr_service() first`

src/eth.rs
reinismu: Is it the case for all chips or just W5500?

src/eth.rs
ivmarkov: Pretty sure all chips.

src/eth.rs
ivmarkov: Pretty sure all chips.

reinismu: Plan was to just change eth.rs API. Passing in `SpiDeviceDriver` instead of `SpiDriver` would mean that caller need to do all the configuration on his own. Wouldn't want to expose that complexity.

I do see how we could create `SpiDeviceDriver` internally to get rid of `init_spi_device` and `get_spi_conf`, but that can be done later as well.

ivmarkov: What additional configuration do you mean specifically? Please point me to it.

As for this being done later - I'm not feeling comfortable breaking the API twice just because. So we take either the current  route (`SpiDriver`) or what I suggested (`SpiDeviceDriver`). With good arguments for one or the other.

reinismu: `max_transfer_size` is not exposed from `SpiDriver`. Any suggestions how I could assert it?

ivmarkov: Oh well. Forget it. Users will get a cryptic error message, they need to decypher. :) Exposing bits and pieces from the configuration of the `SpiDriver` just to assert it here does not feel OK.

src/eth.rs
```diff
@@ -295,13 +291,9 @@ impl<'d> EthDriver<'d> {
...
```
ivmarkov, line 295: I think we are in a pretty good shape now. The only thing remaining I think which is a bit annoying is that we do not have the flexibility to pass the SPI driver either by reference, or by value (The latter is more convenient when you don't plan to share the SPI bus with multiple devices - see for reference `SpiDeviceDriver`; but of course means you pay the extra verbosity of having the `EthDriver` having a non-lifetime generic now.). Also, folks using `EthDriver` with the RMII and the `openeth` impls will also pay the price (I guess, for them the generic would be resolved as `()`).

Any thoughts?

teamplayer3:
closes #176

src/eth.rs
reinismu: Hmm ye not sure what would be best course of action here. What do you think?

reinismu: I looked into this again and even together with GPT-4 I couldn't find a way :)
Most likely will need to redo whole ETH from ground up to make it play nice

ivmarkov: I'll also try to look over the weekend at it. Ideally we should merge this before the upcoming release of the esp-idf-* crates.

ivmarkov: Weekend keeps postponing, but also we have not released either. Sorry for the delay - perhaps *this* weekend. I basically want to see if the `BorrowMut` pattern we use elsewhere can be ergonomic enough here as well. I think it can if we define
```rust
pub struct RMII {}
pub struct OpenEth {}
pub struct SPI<T> where T: BorrowMut<SpiDriver>`
```

and then we parameterize `EthDriver` by `EthDriver<'d, Q>` where Q: is one of `RMII`, `OpenEth` or `SPI<T>`. This might result in a small win (don't remember Rust's method overloading rules exactly) where we can rename back `new_spi`, `new_rmii`, `new_openeth` all to just `new` except that they would be implemented on different parameterizations of `Q`. Might or might not work, have to check.

reinismu: Maybe make common trait `EthDriver` for all of them? Then pass it in `EspEth`. Then have `OpenEthDriver`, `W5500Driver` and `RMIIDriver` that will implement it.

ivmarkov: This way we might have to duplicate a lot of code across the the three implementations, and I'm not sure how that's fixing the main issue: which is - when using SPI - provide the user the capability to either pass the SPI driver by reference, or move it into the driver, where the latter would be like 99% of the cases (where they don't need to share the SPI bus with something else).

Anyway - my approach actually works. Ahem, type-checks.
I'll be merging your patch shortly, and then committing my approach on top of it.

Issue #235, opened 2023-03-08
EspHttpConnection::read can never return an error (incomplete response)


schurwanzn-stabl:
## Problem
`EspHttpConnection::read` returns a Result, suggesting (since it is not documented) that it would return an error when reading the response body fails.

However the ESP-IDF function it calls in it's implementation, `esp_http_client_read_response`, is not documented to return any errors and is implemented to silently ignore any errors that might occur.

## Impact
We have implemented a simple HTTP GET request by calling `EspHttpConnection::new`, then `initiate_request`, followed by `initiate_response`. After which we just `.read()` from EspHttpConnection until it returns Ok(0), imitating the http_request example.

Testing this approach under adverse network conditions [while reading the data in small chunks (4KiB) and simulatenously writing the read data into flash (EspOtaUpdate)] has ended with the code regarding EspHttpConnection terminating it's read normally, even though the read data was incomplete.

ivmarkov:
Unfortunately, without providing a reproducible code snippet, I can only guess what is the problem, and whether the problem is in the Rust HTTP client wrapper, or in the ESP IDF HTTP client itself, or in your code, or in some in-between proxy, as another poster suggested.

I get it that you don't have time (as per the forum message) to implement a reproducible code snippet, but I wonder how you expect to get something actionable from the above problem description?

schurwanzn-stabl:
I can also provide a simple code example to demonstrate how this discrepancy impacts real code:
```rust
// [... initialize logging, connect to network etc. ...]

  let mut http_client = EspHttpConnection::new(&Configuration {
      crt_bundle_attach: Some(esp_idf_sys::esp_crt_bundle_attach),
      ..Default::default()
  })?;

  log::info!("Doing HTTP request...");
  http_client.initiate_request(Method::Get, "https://httpbin.org/drip?duration=30&numbytes=40&code=200&delay=2", &[])?;
  http_client.initiate_response()?;

  if http_client.status() != 200 {
      return Err(anyhow::anyhow!(
          "request to {} returned HTTP status code {}",
          OTA_API_DOWNLOAD_URL,
          http_client.status()
      ));
  }

  log::info!("Content length: {:?}", http_client.content_len());
  log::info!("Reading response body...");

  let mut total_bytes_read = 0;
  let mut buffer = [0u8; 4];

  loop {
      let bytes_read = match http_client.read(&mut buffer) {
          Ok(bytes_read) => bytes_read,
          Err(err) => {
[...]

PR #237, opened 2023-03-10
Improve error handling in EspHttpConnection::read (Issue #235)

schurwanzn-stabl:
See issue for further information.

https://github.com/esp-rs/esp-idf-svc/issues/235#issuecomment-1463590779

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -289,12 +289,19 @@ impl EspHttpConnection {
 
         // workaround since esp_http_client_read does not yet return EAGAIN error in ESP-IDF v4.
         // in ESP-IDF v5 esp_http_client_read will return EAGAIN and this should not be needed.
-        match result {
-            Ok(0) if unsafe { !esp_http_client_is_complete_data_received(self.raw_client) } => {
-                // no error but read 0 bytes and body is not yet complete, probably caused by EAGAIN
-                Err(EspError::from_infallible::<ESP_ERR_HTTP_EAGAIN>())
+        #[cfg(esp_idf_version_major = "4")]
+        {
+            match result {
+                Ok(0) if unsafe { !esp_http_client_is_complete_data_received(self.raw_client) } => {
+                    // no error but read 0 bytes and body is not yet complete, probably caused by EAGAIN
+                    Err(EspError::from_infallible::<ESP_ERR_HTTP_EAGAIN>())
+                }
+                other => other,
             }
-            other => other,
+        }
+        #[cfg(not(esp_idf_version_major = "4"))]
+        {
+            result
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/http/client.rs
```diff
@@ -283,9 +283,19 @@ impl EspHttpConnection {
     pub fn read(&mut self, buf: &mut [u8]) -> Result<usize, EspError> {
         self.assert_response();
 
-        Self::check(unsafe {
-            esp_http_client_read_response(self.raw_client, buf.as_mut_ptr() as _, buf.len() as _)
-        })
+        let result = Self::check(unsafe {
```
ivmarkov, line 286: Hmmmm. I was thinking that the loop which retries on `EAGAIN` would be _inside_ the `read` method, rather than pushing it back to the shoulders of the user?

src/http/client.rs
```diff
@@ -283,9 +283,19 @@ impl EspHttpConnection {
...
+            esp_http_client_read(self.raw_client, buf.as_mut_ptr() as _, buf.len() as _)
+        });
+
+        // workaround since esp_http_client_read does not yet return EAGAIN error in ESP-IDF v4.
+        // in ESP-IDF v5 esp_http_client_read will return EAGAIN and this should not be needed.
+        match result {
```
ivmarkov, line 292: I'm kind of OK with this approach, but isn't it better to actually branch the code for esp idf 5+ vs older esp idf-s and then do different things for each?

ivmarkov (changes requested):

src/http/client.rs
schurwanzn-stabl: Hmm yeah, I just re-read the documentation of `std::io::Read::read` and noticed that it doesn't encourage either way specifically but leaves it completely open to decide how to do it:
> This function does not provide any guarantees about whether it blocks waiting for data, but if an object needs to block for a read and cannot, it will typically signal this via an [Err](https://doc.rust-lang.org/std/result/enum.Result.html#variant.Err) return value.

I suppose it's a question of design philosophy. Either you hide complexity from the user and make it simpler for them to use your API or you pass the complexity on but give them more choice to handle things their way. In this case maybe they don't want to retry immediately but instead do something else first.

I think it is quite clear which one I prefer in this context (embedded Rust) but in the end I would give you the final say.

Just one more thing to consider is what would happen if the server keeps the connection open but just doesn't send any more data? Would this effectively deadlock the thread?
You can specify a timeout on the EspHttpConnection but from the limited testing I did for this issue it seemed like this only determines after which time period an EAGAIN is returned.

src/http/client.rs
ivmarkov: It is all over the place but here you go:
`#[cfg(not(esp_idf_version_major = "4"))]`
^^^ For code which should run on 5+

and
`#[cfg(esp_idf_version_major = "4")]`
^^^ For code which should run on 4. We don't support ES:P IDF earlier than 4.3.

ivmarkov: With the above `cfg` hints written, I'm (as said earlier) kind of fine with the existing way too, so I can merge the current shape of the patch too.

schurwanzn-stabl:
I added conditional compilation to the workaround

Issue #213, opened 2023-01-18
`EspWifi::start_scan` returns `ESP_FAIL` on `master` (succeeds on release)


hawkw:
On an ESP32-C3 (the Adafruit QT Py ESP32-C3 devboard), and on the `master` branch of this repo (commit a866ff3f7d68990f5f4b63d764b177a66b772740), calling `EspWifi::scan`  fails with the `ESP_FAIL` error code. On the current crates.io release (v0.45.0), the WiFi scan succeeds.

The error appears to be returned by the `EspWifi::start_scan` function, rather than by `get_scan_results` --- I determined this by copying the implementation of `EspWifi::scan` into my project, and logging the results of the calls to both `start_scan` and `get_scan_results`.

hawkw:
It seems like several commits to `master` since v0.45.0 touched the WiFi scan code: 3d69a6522cb1bc71e899271ce7410754f4fa528d, 8aac6d0dd41953f16c58c85f14ebb7720c719309, 93591c1c57ba0e217ae33678e70212175648dde0, and b052a0d227da0e234c9d395f38aa9e4118aeaa5e in particular.

dragazo:
I just ran into this issue in my project as well with a different board: ESP32-S3 DevKitC-1. Unfortunately it seems like the only workaround has been to downgrade to latest release, which also forces me to use esp-idf v4.4 rather than v5.0.

ivmarkov:
@hawkw @dragazo we accept PRs!

hawkw:
> @hawkw @dragazo we accept PRs!

I'm happy to try and put a patch together, but I'm not at all sure what actually caused the regression --- if you have any suggestions for a starting point, I can try to get a fix together.

dragazo:
@hawkw From binary searching through the commits with a minimal test program, it looks like the culprit is 3d69a65 (later changes might also be problematic, but this was the first one to break it). I don't have time to look into it now, but I might get a chance next week.

teamplayer3:
Try to configure the wifi in station mode before scanning. The new scan interface can only be used after configuring the wifi in station mode manually.

```rust
wifi.set_configuration(...);
wifi.start();
let res = wifi.scan();
```

ivmarkov:
@teamplayer3 can it also be configured in ap or mixed mode during scanning?

teamplayer3:
> @teamplayer3 can it also be configured in ap or mixed mode during scanning?

[...]

PR #238, opened 2023-03-10
Add docs for wifi scan

teamplayer3:
Added some doc and examples, to clarify the use of the WiFi scan API to avoid #213.


ivmarkov:
We start it and then configure it in the example, rather than the other way around. That's OK, @teamplayer3 ?

teamplayer3:
> We start it and then configure it in the example, rather than the other way around. That's OK, @teamplayer3 ?

Yes, the wrong way around. Fixed it.

PR #240, opened 2023-03-17
Added async Wifi implementation with stubs

pilotniq:
The stubs just block on the sync variants of the functions, and return a 'ready' future. In a later iteration, the async functions that block can be replaced with proper futures. Also added an example of using the async Wifi.

Feedback and comments welcome!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ embassy-time-isr-queue = [
     "embassy-time",
     "esp-idf-hal/embassy-sync",
 ]
+async = ["embedded-svc/nightly", "embedded-svc/experimental"]
 
 [dependencies]
 heapless = { version = "0.7", default-features = false }
@@ -60,3 +61,7 @@ anyhow = "1"
 [[example]]
 name = "http_request"
 required-features = ["experimental"]
+
+[[example]]
+name = "wifi_async"
+required-features = ["async"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,6 @@ nightly = ["embedded-svc/nightly"]
 experimental = ["embedded-svc/experimental"]
 embassy-time-driver = ["embassy-time"]
 embassy-time-isr-queue = ["embassy-sync", "embassy-time", "esp-idf-hal/embassy-sync"]
-async = ["embedded-svc/nightly", "embedded-svc/experimental", "nightly"]
 
 [dependencies]
 heapless = { version = "0.7", default-features = false }
@@ -47,4 +46,4 @@ required-features = ["experimental"]
 
 [[example]]
 name = "wifi_async"
-required-features = ["async"]
+required-features = ["nightly"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,24 +16,13 @@ rust-version = "1.61"
 [features]
 default = ["std"]
 
-std = [
-    "alloc",
-    "anyhow/std",
-    "log/std",
-    "esp-idf-sys/std",
-    "esp-idf-hal/std",
-    "embedded-svc/std",
-]
+std = ["alloc", "anyhow/std", "log/std", "esp-idf-sys/std", "esp-idf-hal/std", "embedded-svc/std"]
 alloc = ["anyhow", "esp-idf-hal/alloc", "embedded-svc/alloc"]
 nightly = ["embedded-svc/nightly"]
 experimental = ["embedded-svc/experimental"]
 embassy-time-driver = ["embassy-time"]
-embassy-time-isr-queue = [
-    "embassy-sync",
-    "embassy-time",
-    "esp-idf-hal/embassy-sync",
-]
-async = ["embedded-svc/nightly", "embedded-svc/experimental"]
+embassy-time-isr-queue = ["embassy-sync", "embassy-time", "esp-idf-hal/embassy-sync"]
+async = ["embedded-svc/nightly", "embedded-svc/experimental", "nightly"]
 
 [dependencies]
 heapless = { version = "0.7", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,16 +31,10 @@ log = { version = "0.4", default-features = false }
 uncased = "0.9.7"
 anyhow = { version = "1", default-features = false, optional = true } # Only used by the deprecated httpd module
 embedded-svc = { git = "https://github.com/esp-rs/embedded-svc.git", branch = "master" }
-esp-idf-sys = { version = "0.32.1", default-features = false, features = [
-    "native",
-] }
-esp-idf-hal = { version = "0.40", default-features = false, features = [
-    "esp-idf-sys",
-] }
+esp-idf-sys = { version = "0.32.1", default-features = false, features = ["native"] }
+esp-idf-hal = { version = "0.40", default-features = false, features = ["esp-idf-sys"] }
 embassy-sync = { version = "0.1", optional = true }
-embassy-time = { version = "0.1", optional = true, features = [
-    "tick-hz-1_000_000",
-] }
+embassy-time = { version = "0.1", optional = true, features = ["tick-hz-1_000_000"] }
 futures = "0.3.27"
 
 [build-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,12 +16,23 @@ rust-version = "1.61"
 [features]
 default = ["std"]
 
-std = ["alloc", "anyhow/std", "log/std", "esp-idf-sys/std", "esp-idf-hal/std", "embedded-svc/std"]
+std = [
+    "alloc",
+    "anyhow/std",
+    "log/std",
+    "esp-idf-sys/std",
+    "esp-idf-hal/std",
+    "embedded-svc/std",
+]
 alloc = ["anyhow", "esp-idf-hal/alloc", "embedded-svc/alloc"]
 nightly = ["embedded-svc/nightly"]
 experimental = ["embedded-svc/experimental"]
 embassy-time-driver = ["embassy-time"]
-embassy-time-isr-queue = ["embassy-sync", "embassy-time", "esp-idf-hal/embassy-sync"]
+embassy-time-isr-queue = [
+    "embassy-sync",
+    "embassy-time",
+    "esp-idf-hal/embassy-sync",
+]
 
 [dependencies]
 heapless = { version = "0.7", default-features = false }
@@ -29,11 +40,18 @@ enumset = { version = "1", default-features = false }
 log = { version = "0.4", default-features = false }
 uncased = "0.9.7"
 anyhow = { version = "1", default-features = false, optional = true } # Only used by the deprecated httpd module
-embedded-svc = { version = "0.24", default-features = false }
-esp-idf-sys = { version = "0.32.1", default-features = false, features = ["native"] }
-esp-idf-hal = { version = "0.40", default-features = false, features = ["esp-idf-sys"] }
+embedded-svc = { git = "https://github.com/esp-rs/embedded-svc.git", branch = "master" }
+esp-idf-sys = { version = "0.32.1", default-features = false, features = [
+    "native",
+] }
+esp-idf-hal = { version = "0.40", default-features = false, features = [
+    "esp-idf-sys",
+] }
 embassy-sync = { version = "0.1", optional = true }
-embassy-time = { version = "0.1", optional = true, features = ["tick-hz-1_000_000"] }
+embassy-time = { version = "0.1", optional = true, features = [
+    "tick-hz-1_000_000",
+] }
+futures = "0.3.27"
 
 [build-dependencies]
 embuild = "0.31"
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -1515,7 +1515,7 @@ where
         self.base_driver.borrow_mut().scan_n()
     }
 
-    pub async fn scan(&mut self) -> Result<Vec<AccessPointInfo>, EspError> {
+    pub async fn scan(&mut self) -> Result<alloc::vec::Vec<AccessPointInfo>, EspError> {
         let base: &mut WifiDriver<'d> = self.base_driver.borrow_mut();
         base.scan()
     }
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -11,7 +11,7 @@ use ::log::*;
 
 use enumset::*;
 
-#[cfg(all(feature = "nightly", feature = "experimental"))]
+#[cfg(feature = "async")]
 use embedded_svc::wifi::asynch;
 
 use embedded_svc::wifi::*;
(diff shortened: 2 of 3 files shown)


src/wifi.rs
```diff
@@ -1355,3 +1358,110 @@ impl WifiWait {
...
+        nvs: Option<EspDefaultNvsPartition>,
+    ) -> Result<Self, EspError> {
+        Ok(EspAsyncWifiDriver {
+            sync: WifiDriver::new(modem, sysloop, nvs)?,
+        })
+    }
+}
+
+#[cfg(all(feature = "nightly", feature = "experimental"))]
+impl<'d> asynch::Wifi for EspAsyncWifiDriver<'d> {
```
ivmarkov, line 1382: The current philosophy of the esp-idf-hal crate is that everything accessible via the e-hal traits should also be doable via an esp-idf-hal API which is straight in the driver struct. This way you don't need nightly experimental features if you don't like these. Therefore, please move your stubs to the driver (as regular async methods) and delegate to these from the trait impl.

src/wifi.rs
```diff
@@ -1355,3 +1358,110 @@ impl WifiWait {
...
+
+#[cfg(all(feature = "async"))]
+pub struct EspAsyncWifiDriver<'d> {
+    sync: WifiDriver<'d>,
+}
+
+#[cfg(all(feature = "async"))]
+impl<'d> EspAsyncWifiDriver<'d> {
+    #[cfg(all(feature = "alloc", esp_idf_comp_nvs_flash_enabled))]
+    pub fn new<M: WifiModemPeripheral>(
```
ivmarkov, line 1370: In terms of reusability, reimplementing everything which is currently in `WifiDriver` from scratch is not a great idea. What you might or might not have noticed (not really sure as the current code is just a bunch of stubs), is that the `WifiDriver` methods are in fact **non-blocking**, in that e.g. calling `start` does not mean that the driver would've started when the method returns. Quite the opposite - the driver would only *initiate* the start procedure, and then you have to subscribe to the event bus and wait for a "start" event notification. Ditto for `connect`.

Hence, what makes sense is to have your `AsyncWifiDriver` taking a `T: BorrowMut<WifiDriver>` parameter (as per above). All of your methods would just call the corresponding methods on `WifiDriver` and then asynchronously wait on the system event bus until the corresponding event is received. Also note that the system event bus (and all event buses in fact) have an async adaptor.

So you can possibly implement a helper struct - `AsyncWifiWait`  - similar in spirit to the blocking `WifiWait` - except that it would asynchronously wait. And then use it in your `AsyncWifiDriver.

src/wifi.rs
```diff
@@ -1355,3 +1358,110 @@ impl WifiWait {
         }
     }
 }
+
+#[cfg(all(feature = "async"))]
+pub struct EspAsyncWifiDriver<'d> {
+    sync: WifiDriver<'d>,
```
ivmarkov, line 1364: Ideally, this should be `sync: T` `where T: BorrowMut<WifiDriver>`, so that the user has the freedom to pass a `&mut` ref as well. Also the driver you are wrapping here is NOT "sync" - see below, so you might want to change the field name.

src/wifi.rs
```diff
@@ -1355,3 +1358,110 @@ impl WifiWait {
         }
     }
 }
+
+#[cfg(all(feature = "async"))]
+pub struct EspAsyncWifiDriver<'d> {
```
ivmarkov, line 1363: Strip the `Esp` prefix please. This stuff should be symmetrically named to `WifiDriver`.

ivmarkov (changes requested):

src/wifi.rs
```diff
@@ -18,6 +18,11 @@ use esp_idf_hal::peripheral::Peripheral;
 
 use esp_idf_sys::*;
 
+#[cfg(feature = "async")]
```
ivmarkov, line 21: * The feature that you should be using is `nightly`, not `async`. Also see my other comments below, you only need to protect with `nightly` where you use a nightly feature. As in TAIT. Not here.

src/wifi.rs
```diff
@@ -327,6 +332,30 @@ pub struct WifiDriver<'d> {
     _p: PhantomData<&'d mut ()>,
 }
 
+#[cfg(feature = "async")]
```
ivmarkov, line 335: None of these are necessary. `async fn` is supported on the stable compiler for structs. Remove *ALL* of these type aliases.

src/wifi.rs
```diff
@@ -439,6 +468,11 @@ impl<'d> WifiDriver<'d> {
         Ok(caps)
     }
 
+    #[cfg(feature = "async")]
```
ivmarkov, line 471: Remove the `async` feature protection and just use plain old `pub async fn get_capabilities`. Note also that I suggested these functions to be on the `AsyncWifiDriver`, not on the `WifiDriver`. So you don't need the `_async` suffix as you won't have name clashes once you move them.

src/wifi.rs
```diff
@@ -449,6 +483,11 @@ impl<'d> WifiDriver<'d> {
         Ok(())
     }
 
+    #[cfg(feature = "async")]
```
ivmarkov, line 486: Ditto everywhere below.

src/wifi.rs
```diff
@@ -1355,3 +1439,90 @@ impl WifiWait {
         }
     }
 }
+
+#[cfg(feature = "async")]
```
ivmarkov, line 1530: As per above, you **do not have to protect** with `feature = async` anything here.

src/wifi.rs
```diff
@@ -1442,3 +1526,90 @@ impl WifiWait {
...
+    _p: PhantomData<WifiDriver<'d>>,
+}
+
+#[cfg(feature = "async")]
+impl<'d, T> AsyncWifiDriver<'d, T>
+where
+    T: BorrowMut<WifiDriver<'d>>,
+{
+    #[cfg(all(feature = "alloc", esp_idf_comp_nvs_flash_enabled))]
+    pub fn new(base_driver: T) -> Result<Self, EspError> {
```
ivmarkov, line 1545: As per above, move all async functions from `WifiDriver` to `AsyncWifiDriver` and turn them into plain ol `pub async fn` variants, and strip their `_async` suffix.

src/wifi.rs
```diff
@@ -1442,3 +1526,90 @@ impl WifiWait {
...
+        Ok(AsyncWifiDriver {
+            base_driver,
+            _p: PhantomData,
+        })
+    }
+}
+
+#[cfg(all(feature = "async"))]
```
ivmarkov, line 1553: Now THIS trait impl is the only thing you need to protect with a feature, but the feature should be `nightly` (as it needs nightly features), not `async`.

ivmarkov (changes requested):

Cargo.toml
```diff
@@ -22,18 +22,26 @@ nightly = ["embedded-svc/nightly"]
...
 [dependencies]
 heapless = { version = "0.7", default-features = false }
 enumset = { version = "1", default-features = false }
 log = { version = "0.4", default-features = false }
 uncased = "0.9.7"
 anyhow = { version = "1", default-features = false, optional = true } # Only used by the deprecated httpd module
-embedded-svc = { version = "0.24", default-features = false }
-esp-idf-sys = { version = "0.32.1", default-features = false, features = ["native"] }
-esp-idf-hal = { version = "0.40", default-features = false, features = ["esp-idf-sys"] }
+embedded-svc = { git = "https://github.com/esp-rs/embedded-svc.git", branch = "master" }
```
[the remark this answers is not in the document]
pilotniq: You are too fast (on a Sunday morning!), I realized that there was formatting I had missed to fix after push, fixed it, but by then you had already reviewed ;-)

src/wifi.rs
```diff
@@ -1442,3 +1445,144 @@ impl WifiWait {
...
+impl<'d, T> asynch::Wifi for AsyncWifiDriver<'d, T>
+where
+    T: BorrowMut<WifiDriver<'d>>,
+{
+    type Error = EspError;
+
+    type GetCapabilitiesFuture<'a> = futures::future::Ready<Result<EnumSet<Capability>, EspError>> where Self: 'a;
+    type GetConfigurationFuture<'a> = futures::future::Ready<Result<Configuration, EspError>> where Self: 'a;
+    type SetConfigurationFuture<'a> = futures::future::Ready<Result<(), EspError>> where Self: 'a;
+    type StartFuture<'a> = Pin<Box<dyn futures::future::Future<Output = Result<(), EspError>> + 'a>> where Self: 'a;
```
ivmarkov, line 1535: Hint: here you need the TAIT future which is nightly only. No Pin, no Box. Basically, `= impl Future<Output = Result<(), Self::Error>>`.

src/wifi.rs
```diff
@@ -1442,3 +1445,144 @@ impl WifiWait {
...
+
+    fn get_configuration(&self) -> Self::GetConfigurationFuture<'_> {
+        futures::future::ready(self.get_configuration())
+    }
+
+    fn set_configuration(&mut self, conf: &Configuration) -> Self::SetConfigurationFuture<'_> {
+        futures::future::ready(self.set_configuration(conf))
+    }
+
+    fn start(&mut self) -> Self::StartFuture<'_> {
```
ivmarkov, line 1556: Hint: and here you just need to do `async move { AsyncWifiDriver::start(self).await }`. In fact, even simply `AsyncWifiDriver::start(self)` without the async block and without the await should work too, but you should check.

ivmarkov (changes requested):

src/wifi.rs
```diff
@@ -1442,3 +1445,144 @@ impl WifiWait {
         }
     }
 }
+
+#[cfg(feature = "nightly")]
```
ivmarkov, line 1449: And again: you **don't** need `nightly` in the `AsyncWifiDriver`, as it does not use ANY nightly features! That's the whole point why I wanted plain Ole pub async functions on the driver - because they do work in the stable compiler. You can hide it behind `experimental`, but that's orthogonal to `nightly`. The one thing you need nightly for is the `impl Future` trick that you have to use when instantiating your future GATs when implementing the async Wifi trait. So only the async WiFi trait implneeds to be hidden behind nightly, nothing else.

src/wifi.rs
```diff
@@ -1442,3 +1446,141 @@ impl WifiWait {
...
+    }
+
+    pub async fn stop(&mut self) -> Result<(), EspError> {
+        self.base_driver.borrow_mut().stop()
+        // TODO: wait until stop is complete?
+    }
+
+    pub async fn connect(&mut self) -> Result<(), EspError> {
+        self.base_driver.borrow_mut().connect()
+        // TODO: wait until connect is complete
```
MabezDev, line 1502: FYI if you would like a pointer on how you might implement this, you can take a look at what I did for the bare metal esp-wifi: https://github.com/esp-rs/esp-wifi/blob/76ba312b55b841cbbe9090c4f8dfb12ad379bb38/esp-wifi/src/wifi/mod.rs#L1437-L1453

MabezDev: Once you get one event working, it should be straight forward for the rest

pilotniq: Thanks, I'll have a look!

src/wifi.rs
```diff
@@ -322,6 +340,7 @@ static mut TX_CALLBACK: Option<Box<dyn FnMut(WifiDeviceId, &[u8], bool) + 'stati
 pub struct WifiDriver<'d> {
     status: Arc<mutex::Mutex<(WifiEvent, WifiEvent)>>,
     _subscription: EspSubscription<System>,
+    sysloop: EspSystemEventLoop,
```
ivmarkov, line 343: I feel it is clearer if the `AsyncWifiDriver` itself takes on constructor a `&SystemEventL:oop` or a `SystemEventLoop`, if it needs it, rather than keeping the sysloop of the `WifiDriver` as a state, just so that `AsyncWifiDriver` can re-use it. `SystemEventLoop` is cloneable, and cloning is cheap, as it is just an Arc-ed "()" under the hood, so to say.

src/wifi.rs
```diff
@@ -336,10 +355,11 @@ impl<'d> WifiDriver<'d> {
     ) -> Result<Self, EspError> {
         Self::init(nvs.is_some())?;
 
-        let (status, subscription) = Self::subscribe(&sysloop)?;
+        let (status, subscription) = Self::subscribe_statuses(&sysloop)?;
```
ivmarkov, line 358: Nit but... if we don't change `WifiDriver` because of my previous comment, let's not rename anything for now.

src/wifi.rs
```diff
@@ -1442,3 +1462,342 @@ impl WifiWait {
...
+        internal_data.subscription = Some(sysloop.subscribe(move |event: &WifiEvent| {
+            if *event == trigger_event {
+                WifiEventFutureData::resolve(&mut copied_data, ())
+            }
+        })?);
+        Ok(())
+    }
+}
+
+impl core::future::Future for WifiEventFuture {
```
ivmarkov, line 1730: All of this (albeit in a more complex way as in the general case you don't want just to be notified if an event happened, but also get the *actual* event) is already implemented [here](https://github.com/esp-rs/embedded-svc/blob/master/src/utils/asyncify/event_bus.rs#L63) and available for you to use, as `esp-idf-svc` has a hard dependency on `embedded-svc`. Why not making use of it? That's what I meant by saying, that the event buses (a.k.a. event loops) in `esp-idf-svc` [are already async-ified](https://github.com/esp-rs/esp-idf-svc/blob/master/src/eventloop.rs#L896) - no need to implement your own low level future plumbing code using closure-based low level subscriptions? 

Now, if for some reason we *cannot* make use of the async-ified version of the event loop, we should perhaps re-use [this piece](https://github.com/esp-rs/esp-idf-hal/pull/218/files#diff-91b00c8963785bf5c4253a4b32505486ee18bf1509f30b31a94513ba3a16d26fR1418), which will likely land soon in `esp-idf-hal` which is saving you from having to deal with atomic wakers & stuff. But with that said, I would really first give a chance to the first reference, as it was designed for cases like this one.

I have also some smaller stylistic comments (again per my earlier comment, how about we implement a `AsyncWifiWait` struct that does the waiting magic by delegating to the async event loop code and then you just re-use it here?) but these can wait and they are much easier to address post-factum IMO.

ivmarkov (changes requested):

ivmarkov:
See #243

ivmarkov:
@pilotniq If you are wondering what is happening... I started to realize that some of my feedback might be well above the complexity level for an external contributor. So what I did is to take your PR, and then apply on top all of the changes suggested here. As a result, we now have #243 .

#243 DOES type-check and build successfully, but I have not yet flashed the outcome on an actual ESP. If you feel adventurous, feel free to build your example and flash it. Who knows - it might work out of the box, or with only minimal changes. :)

Feedback & suggestions in the new PR always welcome, of course!

PR #241, opened 2023-03-18
Wifi: Split status tuple for AP and STA status into two separate fields

pilotniq:
This is in preparation for async, which will want to wait for one of these to change but not the other.

Also, I think it increases legibility and decreases risk of error due to confusing which of .0 and .1 in the tuple is AP and STA statuses.

Note: WifiDriver::is_connected() returns false if the AP is enabled but not started even if the STA is connected. This seems odd to me, shouldn't is_connected() only be about if the Wifi is connected as STA?

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -319,8 +319,12 @@ static mut RX_CALLBACK: Option<
 #[allow(clippy::type_complexity)]
 static mut TX_CALLBACK: Option<Box<dyn FnMut(WifiDeviceId, &[u8], bool) + 'static>> = None;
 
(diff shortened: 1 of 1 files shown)


ivmarkov:
@pilotniq I feel you might not need these changes. Therefore I strongly suggest you first and foremost actually try to turn your stubs into something working, as that would inform you if you need (any) changes to the underlying `WifiDriver`.

ivmarkov:
@pilotniq And please check my review comments in #240 first.

ivmarkov:
See #243

ivmarkov:
See #240

PR #243, opened 2023-03-26
Async & blocking adaptors for the Wifi driver (supersedes #240)

ivmarkov:
* Took the existing PR from @pilotniq
* Applied all my feedback
* Made the wifi_async example buildable
* Made all examples buildable, also as part of the CI

To build, use, e.g.:
`cargo build --example wifi_async --target xtensa-esp32-espidf --features experimental,nightly`


src/wifi.rs
```diff
@@ -1235,6 +1333,36 @@ impl<'d> Drop for EspWifi<'d> {
 #[cfg(esp_idf_comp_esp_netif_enabled)]
 unsafe impl<'d> Send for EspWifi<'d> {}
 
+#[cfg(esp_idf_comp_esp_netif_enabled)]
```
Vollbrecht, line 1336: why is this implemented two times? onetime behind this feature flag and first time without on on line 1020

src/wifi.rs
```diff
@@ -319,6 +325,57 @@ static mut RX_CALLBACK: Option<
 #[allow(clippy::type_complexity)]
 static mut TX_CALLBACK: Option<Box<dyn FnMut(WifiDeviceId, &[u8], bool) + 'static>> = None;
 
+pub trait NonBlocking {
+    fn is_scan_done(&self) -> Result<bool, EspError>;
+
+    fn start_scan(
+        &mut self,
+        scan_config: &config::ScanConfig,
+        blocking: bool,
```
Vollbrecht, line 334: why does this fn still take an blocking bool ? Wouldn't it be nicer to create an WifiDriver in an blocking or nonblocking state and depending on that it will be set true in the blocking impl and false in the non blokcking? Or is it to not break the api to hard?

ivmarkov: The base driver is in fact nonblocking (event loop based) but not async. It implements the `Wifi` trait because, well, the trait actually does not specify if its methods block or not. What you are asking for is now implemented in the `AsyncWifi` and `BlockingWifi` wrappers, which are just wrapping the base nonblocking driver and wait on the event loop either asynchronously or in a blocking way. And *also* implement either the `Wifi` or the `WifiAsync` trait.

The actual problem here is that the `NonBlocking` trait is marked as `pub` but it is not an API - it is just an internal implementation trick that allows AsyncWifi and BlockingWifi to wrap both the nw level 2  driver, as well as the nw level 3 `EspWifi` thing. I'll see if I can hide this trait so that folks are not confused.

src/wifi.rs
ivmarkov: Because both the nw level 2 `WifiDriver` as well as the ip nw level 3 `EspWifi` thing can be wrapped in `AsyncWifi` or `BlockingWifi`? Basically, if the user does not need the nw level 3 layer (ip) he can still consume the nw level 2 with the async / blocking wrappers.

examples/wifi_async.rs
```diff
@@ -0,0 +1,48 @@
...
+
+use embedded_svc::wifi::{AuthMethod, ClientConfiguration, Configuration};
+use esp_idf_hal::prelude::Peripherals;
+use esp_idf_svc::timer::EspTaskTimerService;
+use esp_idf_svc::wifi::{AsyncWifi, EspWifi};
+use esp_idf_svc::{eventloop::EspSystemEventLoop, nvs::EspDefaultNvsPartition};
+use esp_idf_sys::{self as _}; // If using the `binstart` feature of `esp-idf-sys`, always keep this module imported
+use futures::executor::block_on;
+
+const SSID: &'static str = "<Your SSID here>";
```
MabezDev, line 15: Shall we make this use the `env!` macro instead?

src/wifi.rs
ivmarkov: Yeah. not great, but I can't make `NonBlocking` non-`pub` without bloating the existing implementation even further. And I just bloated it a lot - in that - besides `BlockingWifi` and `AsyncWifi`, we now also have `BlockingEth`, `AsyncEth`, `BlockingNetif`, `AsyncNetif` and `NetifStatus` (the latter we probably need to promote to `embedded-svc` at some point when we believe we have a good, trait-able abstraction for a Network Interface.

PR #249, opened 2023-04-11
nvs: support StorageIterate

codyps:
This implements `StorageIterate` for `EspNvs`. Note that while the `Error` on iteration is setup to work, the actual implementation in esp-idf does not return any errors during iteration (other than "Not Found", indicating iteration is complete). As a result, `next()` will always return `Some(Ok(...))` or `None` and never `Some(Err(...))`.

---
This depends on the esp-idf 5 iteration APIs (the ones in esp-idf 4 aren't supported) plus the change in https://github.com/espressif/esp-idf/pull/11118. (Which we can apply in esp-idf-sys).

Plan on landing this:

 1. restrict to esp-idf 5
 2. have esp-idf-sys apply https://github.com/espressif/esp-idf/pull/11118 to the esp-idf we build (5 & master)

---
embedded-svc PR: https://github.com/esp-rs/embedded-svc/pull/48

### DIFF
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -767,25 +767,34 @@ impl<'a, T: NvsPartitionId> Iterator for EspNvsIterEntries<'a, T> {
     type Item = Result<NvsEntry, EspError>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        if self.iter.is_null() {
-            return None;
-        }
-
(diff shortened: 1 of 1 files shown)


ivmarkov:
See [this](https://github.com/esp-rs/embedded-svc/pull/48#issuecomment-1509854469) though. Let me know what you think.

ivmarkov:
Closing as #599 was merged.

Issue #254, opened 2023-04-19
Rework README.md to contain better information for new users / simple how to get started 


Vollbrecht:

ivmarkov:
Easiest is to just put there references to the `esp-idf-template` project, which already does have a how-to-get-started info, if by that you mean how to assemble a minimal project that depends on `esp-idf-svc`. We should probably put the same link in `esp-idf-hal` and `esp-idf-sys` readmes.

ivmarkov:
We can also put mini.al info how to build the crate with examples (at least one more is coming next to the existing one). Ditto for `esp-idf-hal`.

Vollbrecht:
yeah, i was thinking something along this line - maybe also add a link to the wokwi hello-world for a zero setup example. i will come up with something and make a PR

ivmarkov:
Hopefully addressed now.

PR #256, opened 2023-04-20
readme draft

Vollbrecht:
addresses #254

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,9 +80,21 @@ For the XTENSA case, you need to remove the XTENSA toolchain installed via espup
 
 # Chat
 Join the ESP-RS community on Matrix chat for help or questions: https://matrix.to/#/#esp-rs:matrix.org
+
+# Documentation
+Check out the `./examples/` folder for some basic examples.
+
+Please use either [this docs on github pages](https://esp-rs.github.io/esp-idf-hal/esp_idf_hal/index.html) or build the documentation yourself with 
+```
+cargo doc --target riscv32imc-esp-espidf -Zbuild-std=std,panic_abort -Zbuild-std-features=panic_immediate_abort
+```
+ to get a basic API overview, since current docs.rs have certain build limitations.
+
+ The current state of the documentation is pretty minimal, please try to check out the aditional information provided below. Feel free to ask questions around the API on the matrix chat. 
 # Aditional Information
 
 * The [Rust on ESP Book](https://esp-rs.github.io/book/)
+* The [ESP Embedded Training](https://github.com/esp-rs/espressif-trainings)
 * The [embedded-svc](https://github.com/esp-rs/embedded-svc) project
 * The [esp-idf-template](https://github.com/esp-rs/esp-idf-template) project
 * The [esp-idf-sys](https://github.com/esp-rs/esp-idf-sys) project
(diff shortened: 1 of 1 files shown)


ivmarkov:
My one concern with this PR is that while it refers to `esp-idf-template`, it is nevertheless *also* containing bits and pieces of information from the installation prerequisites of `esp-idf-template`. To maintain the information up to date, I think for everything related to how to use this crate we do need to simply redirect to `esp-idf-template` and nothing more.

At most, we might provide on top a sample command line how to build the examples.

Vollbrecht:
> My one concern with this PR is that while it refers to `esp-idf-template`, it is nevertheless _also_ containing bits and pieces of information from the installation prerequisites of `esp-idf-template`. To maintain the information up to date, I think for everything related to how to use this crate we do need to simply redirect to `esp-idf-template` and nothing more.
> 
> At most, we might provide on top a sample command line how to build the examples.

My reasoning behind it. 
1) A new user is most likely to hit either esp-idf-hal/svc repository and not the template. Presenting all necessary  setup in a "minimal" way helps winning a potential user to decide if using esp-idf-* is viable. I think this can help highlight that for most people the setup is really fast. I already reduced the setup guide compared to the one in esp-idf-template, and i think this part will not change a lot over the time. 
2) All information that is needed to be going, should be found in one place, not spread out over 10 different places [as long as the information itself is not the size of an book]. ( In that case i think duplication of that information is not bad)  

But if you see it, still otherwise i can remove that section.

ivmarkov:
Yeah. But if you look at the size of the README in `esp-idf-template`... we might be fooling ourselves that we can squeeze this stuff in the README of `esp-idf-svc`. It will simply be incomplete, and over time it will grow all the way to the size of `esp-idf-template`, and then we have to do the same for `esp-idf-hal`, and then probably for `esp-idf-sys` too, and then we'll have the same stuff spread all over 4 places, except slightly different everywhere and probably outdated at most places.

I strongly prefer instead to hyperlink the `esp-idf-template` in the other `esp-idf-*` crates w.r.t. "how to build". This is the purpose of the `esp-idf-template` after all - to get people going with the nitty gritty stuff of installing all the dependencies and so on.

I would rather describe - in the READMEs of `esp-idf-svc` and `esp-idf-hal` the features those crates offer, rather than "new to Embedded, Rust and esp-rs, how to build??" all over again.

ivmarkov:
Not to mention that folks don't want to build `esp-idf-svc` / `esp-idf-hal` per se. These are library crates. What they want instead, is to *use* them in a binary crate which shifts the topic to "how do I build a binary crate that happens to depend on the `esp-idf-*` ecosystem, which is `esp-idf-template` all over again.

The one thing worth mentioning in the readme is the command line which builds the examples. Which is in the CI btw. But even that detail should be mentioned after the hyperlink to `esp-idf-template`, and the wording should assume that the user had succeeded in installing all dependencies depscribed in `esp-idf-template` as well as even a template-generated binary crate first.

Vollbrecht:
Ok, i got you. I will change both README's here and potentialy the esp-idf-sys one to all have same wording and style 1)  directly linking to esp-idf-template. 2) Emphasize your point in all 3 README's that it's an library crate -> setup and usage through esp-idf-template. 3) Updating some info in the esp-idf-template README regarding uninstall + " mention `cargo run` " support

ivmarkov:
I've taken your suggestions (modulo build instructions) and incorporated in the README. Small wording fixes here and there.

PR #262, opened 2023-05-31
Added docstrings for wifi module

IvanSanchez:
I recently became slightly frustrated with some of the documentation of the Wifi module. I noted that `WifiDriver` just calls functions from `esp_idf_sys` (which do have more detailed notes about their behaviour and preconditions), and also noted that `EspWifi`, `BlockingWifi` and `AsyncWifi` are just wrappers, adding links to the functions being wrapped.

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -374,8 +374,13 @@ where
     }
 }
 
-/// Low(-ish) level WiFi driver. Wraps the Wifi calls from `esp-idf-sys` in a
-/// structured way.
+/// This struct provides a safe wrapper over the ESP IDF Wifi C driver. The driver
+/// works on Layer 2 (Data Link) in the OSI model, in that it provides facilities
+/// for sending and receiving ethernet packets over the WiFi radio.
+///
+/// For most use cases, utilizing `EspWifi` - which provides a networking (IP)
+/// layer as well - should be preferred. Using `WifiDriver` directly is beneficial
+/// only when one would like to utilize a custom, non-STD network stack like `smoltcp`.
 pub struct WifiDriver<'d> {
     status: Arc<mutex::Mutex<(WifiEvent, WifiEvent)>>,
     _subscription: EspSubscription<System>,
@@ -1211,7 +1216,15 @@ impl<'d> Wifi for WifiDriver<'d> {
     }
 }
 
-/// Wraps a [`WifiDriver`] and handles the associated [`EspNetif`] network interface(s).
+/// `EspWifi` wraps a `WifiDriver` Data Link layer instance, and binds the OSI
+/// Layer 3 (network) facilities of ESP IDF to it. In other words, it connects
+/// the ESP IDF AP and STA Netif interfaces to the Wifi driver. This allows users
+/// to utilize the Rust STD APIs for working with TCP and UDP sockets.
+///
+/// This struct should be the default option for a Wifi driver in all use cases
+/// but the niche one where bypassing the ESP IDF Netif and lwIP stacks is
+/// desirable. E.g., using `smoltcp` or other custom IP stacks on top of the
+/// ESP IDF Wifi radio.
 #[cfg(esp_idf_comp_esp_netif_enabled)]
 pub struct EspWifi<'d> {
     ap_netif: EspNetif,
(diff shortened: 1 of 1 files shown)


Vollbrecht:
:heart:  thanks for helping out documenting stuff !

src/wifi.rs
```diff
@@ -374,6 +374,8 @@ where
     }
 }
 
+/// Low(-ish) level WiFi driver. Wraps the Wifi calls from `esp-idf-sys` in a
```
ivmarkov, line 377: "Low-(-ish)" is not providing enough information. Suggestion:
```
This struct provides a safe wrapper over the ESP IDF Wifi C driver. The driver works on Layer 2 (Data Link) in the OSI model, in that it provides facilities for sending and receiving ethernet packets over the WiFi radio. 

For most use cases, utilizing `EspWifi` - which provides a networking (IP) layer as well - should be preferred. Using `WifiDriver` directly is beneficial only when one would like to utilize a custom, non-STD network stack like `smoltcp`.
```

src/wifi.rs
```diff
@@ -1184,6 +1211,7 @@ impl<'d> Wifi for WifiDriver<'d> {
     }
 }
 
+/// Wraps a [`WifiDriver`] and handles the associated [`EspNetif`] network interface(s).
```
ivmarkov, line 1214: Suggestion:
```
`EspWifi` wraps a `WifiDriver` Data Link layer instance, and binds the OSI Layer 3 (network) facilities of ESP IDF to it. In other words, it connects the ESP IDF AP and STA Netif interfaces to the Wifi driver. This allows users to utilize the Rust STD APIs for working with TCP and UDP sockets.

This struct should be the default option for a Wifi driver in all use cases but the niche one where bypassing the ESP IDF Netif and lwIP stacks is desirable. E.g., using `smoltcp` or other custom IP stacks on top of the ESP IDF Wifi radio.
```

src/wifi.rs
```diff
@@ -1625,6 +1664,8 @@ fn matches_wifi_event(event: &WifiEvent) -> bool {
     )
 }
 
+/// Wraps a [`WifiDriver`] or [`EspWifi`], and offers strictly synchronous (blocking)
+/// function calls for their functionality.
```
ivmarkov, line 1668: You might want to add an example here in that wrapping is possible both by moving the `WifiDriver` / `EspWifi` instance into `BlockingWifi`, as well as by mutable borrowing it (i.e. `&mut wifi_driver`), where the latter allows one to re-use the driver elsewhere once `BlockingWifi` is dropped.

src/wifi.rs
```diff
@@ -1684,22 +1735,27 @@ where
...
         self.wifi.scan_n()
     }
 
+    /// As per [`WifiDriver::scan()`]
     #[cfg(feature = "alloc")]
     pub fn scan(&mut self) -> Result<alloc::vec::Vec<AccessPointInfo>, EspError> {
         self.wifi.scan()
     }
 
+    /// Waits for [`WifiEvent`]s. The wait ends when a wifi event happens, and
```
ivmarkov, line 1757: Correction: not for that event. The function provided by the user does not have access to the actual event.
Perhaps a more correct wording would be:
```
Waits for a certain condition provided by the user in the form of a `matcher` callback to become true. Most often than not that condition would be related to the state of the Wifi driver. In other words, whether the driver is started, stopped, (dis)connected and so on.

Note that the waiting is not done internally using busy-looping and/or timeouts. Rather, the condition (`matcher`) is evaluated initially, and if it returns `false`, it is re-evaluated each time the ESP IDF C 
Wifi driver posts a Wifi event on the system event loop. The reasoning behind this is that changes to the state of the Wifi driver are always accompanied by posting Wifi events.  
```

src/wifi.rs
```diff
@@ -1897,6 +1973,8 @@ where
         self.wifi.get_scan_result()
     }
 
+    /// Waits for [`WifiEvent`]s. The wait ends when a wifi event happens, and
```
ivmarkov, line 1976: Ditto here: you might want to reword similarly to the suggestion in `BlockingWifi`.

src/wifi.rs
```diff
@@ -1917,15 +1995,20 @@ impl<T> AsyncWifi<T>
...
         self.wifi.is_up()
     }
 
+    /// Waits until the underlaying network interface is up.
     pub async fn wait_netif_up(&self) -> Result<(), EspError> {
         self.ip_wait_while(|| self.wifi.is_up().map(|s| !s), Some(CONNECT_TIMEOUT))
             .await
     }
 
+    /// Waits for [`IpEvent`]s on the network stack. The wait ends when a
```
ivmarkov, line 2009: Ditto.

ivmarkov (changes requested):

src/wifi.rs
IvanSanchez: Alas, I wasn't even aware that `BlockingWifi` was wrapping a `EspWifi` instance until I read the source code - I don't have a use case where I'd reuse the driver by dropping `BlockingWifi`, much less I feel confident to write a reliable example.

I'd rather leave this for a future improvement.

PR #267, opened 2023-06-18
Add methods to query content length before receiving

jordanhalase:
With this code it will now be possible to query for incoming content size before attempting to receive any data. Users can query the content length and ensure a buffer capacity or send a graceful fail message to the client.

Calling `content_len()` does not require invariant checks so it can be inlined, but I can add the invariant check if you insist. :)

## Example usage: JSON POST handler

```Rust
#[derive(Deserialize)]
struct FormData<'a> {
    first_name: &'a str,
    last_name: &'a str
}

// ...

server.fn_handler("/post", Method::Post, |mut req| {
    let len = req.connection().content_len();

    if len > MAX_LEN {
        req.into_status_response(413)?
            .write_all("Request too big".as_bytes())?;
        return Ok(());
    }

    let mut buf = vec![0; len];
    req.read_exact(&mut buf)?;
    let mut resp = req.into_ok_response()?;

    if let Ok(form) = serde_json::from_slice::<FormData>(&buf) {
        write!(resp, "Your first name is `{}` and last name is `{}`.", form.first_name, form.last_name)?;
    } else {
        resp.write_all("JSON error".as_bytes())?;
    }

    Ok(())
}).unwrap();
```

### DIFF
--- a/src/http/server.rs
+++ b/src/http/server.rs
@@ -1001,6 +1001,20 @@ pub mod ws {
             }
         }
 
+        pub fn recv_frame_type(&mut self) -> Result<(FrameType, usize), EspError> {
+            match self {
+                Self::New(_, _) => Err(EspError::from_infallible::<ESP_FAIL>()),
+                Self::Receiving(_, raw_req) => {
+                    let mut raw_frame: httpd_ws_frame_t = Default::default();
+
+                    esp!(unsafe { httpd_ws_recv_frame(*raw_req, &mut raw_frame as *mut _, 0) })?;
+
+                    Ok(Self::create_frame_type(&raw_frame))
+                }
+                Self::Closed(_) => Ok((FrameType::SocketClose, 0)),
+            }
+        }
+
         fn create_raw_frame(frame_type: FrameType, frame_data: &[u8]) -> httpd_ws_frame_t {
             httpd_ws_frame_t {
                 type_: match frame_type {
--- a/src/http/server.rs
+++ b/src/http/server.rs
@@ -689,6 +689,11 @@ impl<'a> EspHttpConnection<'a> {
         self.headers.is_none()
     }
 
+    #[inline(always)]
+    pub fn content_len(&self) -> usize {
+        self.request.0.content_len
+    }
+
     pub fn read(&mut self, buf: &mut [u8]) -> Result<usize, EspError> {
         self.assert_request();
 


src/http/server.rs
```diff
@@ -689,6 +689,11 @@ impl<'a> EspHttpConnection<'a> {
         self.headers.is_none()
     }
 
+    #[inline(always)]
```
MabezDev, line 692:
```suggestion
    #[inline]
```

Let let the optimizer do its job :D.

src/http/server.rs
```diff
@@ -996,6 +1001,20 @@ pub mod ws {
             }
         }
 
+        pub fn recv_frame_type(&mut self) -> Result<(FrameType, usize), EspError> {
```
MabezDev, line 1004: Did you mean to add this in this PR?

ivmarkov:
@jordanhalase I do not understand what this PR is implementing which does not exist already?
* You can already call `request.content_len()`
* You can already call `connection.content_len()`

Moreover, the above two calls will return something only if the client actually DID set the `Content-Length` header. Which might not be the case at all.

ivmarkov:
If that's not clear, the `Headers` trait from `embedded-svc` already does have a `fn content_len(&self) -> Option<&str>` method. Moreover, both `Connection` and `Request` implement the `Headers` trait. So just add `use` for the relevant traits.

jordanhalase:
> @jordanhalase I do not understand what this PR is implementing which does not exist already?
> 
>     * You can already call `request.content_len()`
> 
>     * You can already call `connection.content_len()`
> 
> 
> Moreover, the above two calls will return something only if the client actually DID set the `Content-Length` header. Which might not be the case at all.

Hmm, you're right. I'm not sure how I missed that, but I can revert commit c09c6d4 once I get home.

However, I'm having trouble finding the frame type and payload length of a WebSocket connection (without calling `recv`). The C documentation recommends calling `httpd_ws_recv_frame()` with `max_len` 0 to get frame size without receiving any data, hence why I implemented `recv_frame_type()` in this PR.

src/http/server.rs
jordanhalase: I did. I need a way to get the WebSocket frame type and size **without** receiving any data yet, to ensure size invariants and to allocate a buffer big enough for the incoming data.

ivmarkov:
> > @jordanhalase I do not understand what this PR is implementing which does not exist already?
> > ```
> > * You can already call `request.content_len()`
> > 
> > * You can already call `connection.content_len()`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Moreover, the above two calls will return something only if the client actually DID set the `Content-Length` header. Which might not be the case at all.
> 
> Hmm, you're right. I'm not sure how I missed that, but I can revert commit [c09c6d4](https://github.com/esp-rs/esp-idf-svc/commit/c09c6d4ac8e8597726202418b41629dbebc88630) once I get home.
> 
> However, I'm having trouble finding the frame type and payload length of a WebSocket connection (without calling `recv`). The C documentation recommends calling `httpd_ws_recv_frame()` with `max_len` 0 to get frame size without receiving any data, hence why I implemented `recv_frame_type()` in this PR.
> 
> It's essentially a copy of `EspHttpWsConnection::recv()` without the internal second call to fill a buffer.

Isn't calling `EspHttpWsConnection::recv(self, &mut [])` doing exactly what you want?

jordanhalase:
> > > @jordanhalase I do not understand what this PR is implementing which does not exist already?
> > > ```
> > > * You can already call `request.content_len()`
> > > 
> > > * You can already call `connection.content_len()`
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Moreover, the above two calls will return something only if the client actually DID set the `Content-Length` header. Which might not be the case at all.
> > 
> > 
> > Hmm, you're right. I'm not sure how I missed that, but I can revert commit [c09c6d4](https://github.com/esp-rs/esp-idf-svc/commit/c09c6d4ac8e8597726202418b41629dbebc88630) once I get home.
> > However, I'm having trouble finding the frame type and payload length of a WebSocket connection (without calling `recv`). The C documentation recommends calling `httpd_ws_recv_frame()` with `max_len` 0 to get frame size without receiving any data, hence why I implemented `recv_frame_type()` in this PR.
> > It's essentially a copy of `EspHttpWsConnection::recv()` without the internal second call to fill a buffer.
> 
> Isn't calling `EspHttpWsConnection::recv(self, &mut [])` doing exactly what you want?

Hah, somehow I thought this would call the internal function twice but I didn't look hard enough.

I'll close this PR and probably open a new one later on with some POST and WebSocket example code showing what you just told me. :)

PR #269, opened 2023-06-22
Add higher level server examples

### LINKED ISSUE #266: Session context for websockets
I found the websocket code for the HTTP server at [esp_idf_svc::http::server::ws](https://github.com/esp-rs/esp-idf-svc/blob/b97f7a8d3eb513fcee5cad9d90bf8f0a22a675a1/src/http/server.rs#L893), but am having a little trouble figuring out how to add the session context to this.

When writing code in C I'm used to setting `sess_ctx` in [httpd_req_t](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/esp_http_server.html#_CPPv49httpd_req) to some `malloc()`'d struct on open (and maybe safely mutating some global data) and setting the `free_ctx` member to `free()` the data (and maybe safely mutating some global data).

I'm not sure how to achieve this with esp-idf-svc. Is there currently a way to have custom websocket session data or is it to be implemented? I need websocket session contexts to send some metadata/setup information before continuing.

Thanks

jordanhalase:
This shows example usage of a few concepts:

* Querying payload size in a POST handler before receiving data
* Parsing JSON in a POST handler
* Querying frame size in a WebSocket handler before receiving data (using @ivmarkov's fix)
* WebSocket sessions

Closes #266


examples/ws_guessing_game.rs
```diff
@@ -0,0 +1,236 @@
...
+        SSID, PASSWORD
+    );
+
+    let server_configuration = esp_idf_svc::http::server::Configuration {
+        stack_size: STACK_SIZE,
+        ..Default::default()
+    };
+
+    // Keep wifi running forever
+    core::mem::forget(wifi);
```
Vollbrecht, line 233: its fine to use mem::forget to make the example more concise and show the need trick with ending the main etc, but in the light that this is an example and many non-trained people will read it - people need to be careful using it. Many people are still not fully aware of the threaded nature, forgetting that they are running on an OS and that its only the underlying C part that is still running. 
What i am try to say is maybe we should add a link to https://doc.rust-lang.org/stable/core/mem/fn.forget.html , and elaborate a small note why its usage is "save" here and people should think before adapting it in there code.

examples/ws_guessing_game.rs
```diff
@@ -0,0 +1,236 @@
...
+            // may only be called with an empty buffer exactly once to receive the
+            // incoming buffer size, then must be called exactly once to receive the
+            // actual payload.
+
+            let (_frame_type, len) = match ws.recv(&mut []) {
+                Ok(frame) => frame,
+                Err(e) => return Err(e),
+            };
+
+            if len > MAX_LEN {
```
Vollbrecht, line 152: with this specific MAX_LEN can  you even trigger a Request too big? A frame should be bounded by the underlying packet MTU size or am i wrong here, in which case we would get spitted packages, and would never come close to 10KB

Vollbrecht (commented):
Thank you for tainting the repository with JS :joy: 

Jokes aside in a first fast glance it looks good to me.

examples/ws_guessing_game.rs
ivmarkov: WS frames run on top of TCP/IP so MTU does not apply to them.

examples/ws_guessing_game.rs
jordanhalase: Trying to think of a concise way to explain it without too many words. How about

>Keep wifi running forever
>Do not call this if you ever want
>to stop or access it later
>https://doc.rust-lang.org/stable/core/mem/fn.forget.html

?

examples/ws_guessing_game.rs
jordanhalase: You can return any `esp_error_t` from any handler and the socket will be closed as per the IDF documentation. The specific error type used is pretty much up to the socket handler. ESP_FAIL is good but it's better to be more descriptive.

jordanhalase:
I just made some small improvements and explained the usage of `core::mem::forget()`. :)

Vollbrecht:
Thanks for the PR :heart:

PR #271, opened 2023-06-22
Fixes #270: add EspNvs::set_str CString error handling

### LINKED ISSUE #270: Return Err instead of panic when EspNvs::set_str called with no CString compatible strings
being "with no CString compatible strings" strings with nulls inside (allowed in rust but not in c).

The fix would be around line https://github.com/esp-rs/esp-idf-svc/blob/master/src/nvs.rs#L455 , changing the unwrap for an error raise.

In order to test this, you could call set_str with a value like "hi\0", that currently creates a panic.

If you want, I can send a PR for this. I'm not a Rust guru but it seems easy enough (famous last words :D).

punkto:
I have added an example of NVS use that triggers the changed lines.

### DIFF
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -89,7 +89,10 @@ impl NvsCustom {
         partition: &str,
         registrations: &mut alloc::collections::BTreeSet<CString>,
     ) -> Result<Self, EspError> {
-        let c_partition = CString::new(partition).unwrap();
+        let c_partition = match CString::new(partition) {
+            Ok(v) => v,
+            Err(_) => return Err(EspError::from_non_zero(NonZeroI32::new(ESP_FAIL).unwrap())),
+        };
 
         if registrations.contains(c_partition.as_ref()) {
             return Err(EspError::from_infallible::<ESP_ERR_INVALID_STATE>());
@@ -175,7 +178,10 @@ impl<T: NvsPartitionId> EspNvs<T> {
         namespace: &str,
         read_write: bool,
     ) -> Result<Self, EspError> {
-        let c_namespace = CString::new(namespace).unwrap();
+        let c_namespace = match CString::new(namespace) {
+            Ok(v) => v,
+            Err(_) => return Err(EspError::from_non_zero(NonZeroI32::new(ESP_FAIL).unwrap())),
+        };
 
         let mut handle: nvs_handle_t = 0;
 
@@ -214,7 +220,10 @@ impl<T: NvsPartitionId> EspNvs<T> {
     }
 
     pub fn remove(&mut self, name: &str) -> Result<bool, EspError> {
-        let c_key = CString::new(name).unwrap();
+        let c_key = match CString::new(name) {
+            Ok(v) => v,
+            Err(_) => return Err(EspError::from_non_zero(NonZeroI32::new(ESP_FAIL).unwrap())),
+        };
 
         // nvs_erase_key is not scoped by datatype
         let result = unsafe { nvs_erase_key(self.1, c_key.as_ptr()) };
@@ -230,7 +239,10 @@ impl<T: NvsPartitionId> EspNvs<T> {
     }
 
     fn len(&self, name: &str) -> Result<Option<usize>, EspError> {
-        let c_key = CString::new(name).unwrap();
+        let c_key = match CString::new(name) {
+            Ok(v) => v,
+            Err(_) => return Err(EspError::from_non_zero(NonZeroI32::new(ESP_FAIL).unwrap())),
+        };
 
         let mut value: u_int64_t = 0;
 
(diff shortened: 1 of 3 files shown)


src/nvs.rs
```diff
@@ -452,7 +453,10 @@ impl<T: NvsPartitionId> EspNvs<T> {
 
     pub fn set_str(&mut self, name: &str, val: &str) -> Result<(), EspError> {
         let c_key = CString::new(name).unwrap();
-        let c_val = CString::new(val).unwrap();
+        let c_val = match CString::new(val) {
```
ivmarkov, line 456: OK so you've addressed *this* `CString` construction, but you have not addressed the one above (for the key), nor the `CString` constructions elsewhere in this module. I think we should either address all `CString` constructions from user-provided `str`s - at least within this module - or none.

ivmarkov (changes requested):

src/nvs.rs
punkto: I was just addressing the issue and I was not sure if the solution was acceptable for you. No problem, I'll complete the module.

Vollbrecht:
Thank you for trying to work on this problem!
In light that CString is used in a lot of places in esp-idf-svc, in contrast to esp-idf-hal, it might be good to discuss it beforehand, so potential changes can be applied to the complete crate and not only patched on some places, such that overall the behavior is similar across the crate,  either by documenting it better or transforming it.

punkto:
> In light that CString is used in a lot of places in esp-idf-svc, in contrast to esp-idf-hal, it might be good to discuss it beforehand, so potential changes can be applied to the complete crate and not only patched on some places, such that overall the behavior is similar across the crate, either by documenting it better or transforming it.

I agree. The problem I see is that when we use a version of the module that does not use the IDF, then we wouldn't need to worry about this. On the other hand, you could argue that the NVS is defined to be used with C-type strings, so all nulls should be trimmed for strings used with the NVS so it is compatible with C.

Anyway, I'm happy to help. Thanks for your support!

ivmarkov:
> > In light that CString is used in a lot of places in esp-idf-svc, in contrast to esp-idf-hal, it might be good to discuss it beforehand, so potential changes can be applied to the complete crate and not only patched on some places, such that overall the behavior is similar across the crate, either by documenting it better or transforming it.
> 
> I agree. The problem I see is that when we use a version of the module that does not use the IDF, then we wouldn't need to worry about this. On the other hand, you could argue that the NVS is defined to be used with C-type strings, so all nulls should be trimmed for strings used with the NVS so it is compatible with C.

I'm not really sure what you mean by "a version of the module". `esp-idf-svc` is hard-wired to ESP IDF. If you mean "using another implementation of the e.g. `RawStorage` trait from `embedded-svc` - well - it will not error out on string with 0s inside. Big deal.

I've in the meantime went ahead and in fact applied your checks to *all* of `esp-idf-svc`. Everywhere. So thanks for the idea, and sorry that it renders the changes suggested in your PR obsolete (I've used a utility function so that I don't have to repeat the `match` pattern everywhere).

What I still find valuable in your PR however is the example how to use the API. If you could update the PR to only contain the example, and then don't stress so much on the "key/value should not contain 0s" (this is an example - i.e. how to implement the happy path - not a unit test), then I would very much like to merge it.

BTW - you do realize that this module exposes two *separate* set of APIs?
- the get_XXX/set_XXX API (where XXX is u8, str, etc.) - this is only for interop with C code that uses the  C ESP IDF NVS API as well
- the `get_raw`/`set_raw` APIs that take a `&[u8]`. This is the "native" Rust API that implements the `RawStorage` trait from `embedded-svc` and it should be preferred actually, as you can layer on top of it any serde you want.

> 
> Anyway, I'm happy to help. Thanks for your support!

punkto:
> > > In light that CString is used in a lot of places in esp-idf-svc, in contrast to esp-idf-hal, it might be good to discuss it beforehand, so potential changes can be applied to the complete crate and not only patched on some places, such that overall the behavior is similar across the crate, either by documenting it better or transforming it.
> > 
> > 
> > I agree. The problem I see is that when we use a version of the module that does not use the IDF, then we wouldn't need to worry about this. On the other hand, you could argue that the NVS is defined to be used with C-type strings, so all nulls should be trimmed for strings used with the NVS so it is compatible with C.
> 
> I'm not really sure what you mean by "a version of the module". `esp-idf-svc` is hard-wired to ESP IDF. If you mean "using another implementation of the e.g. `RawStorage` trait from `embedded-svc` - well - it will not error out on string with 0s inside. Big deal.
> 

I was really mixing no_IDF with no_std... :D

> I've in the meantime went ahead and in fact applied your checks to _all_ of `esp-idf-svc`. Everywhere. So thanks for the idea, and sorry that it renders the changes suggested in your PR obsolete (I've used a utility function so that I don't have to repeat the `match` pattern everywhere).
> 

Thanks for the effort. I'll read your implementation so I can learn to avoid adding so many `match'.

> What I still find valuable in your PR however is the example how to use the API. If you could update the PR to only contain the example, and then don't stress so much on the "key/value should not contain 0s" (this is an example - i.e. how to implement the happy path - not a unit test), then I would very much like to merge it.
> 

OK. Scheduled for next week.

> BTW - you do realize that this module exposes two _separate_ set of APIs?
> 
>     * the get_XXX/set_XXX API (where XXX is u8, str, etc.) - this is only for interop with C code that uses the  C ESP IDF NVS API as well
> 
>     * the `get_raw`/`set_raw` APIs that take a `&[u8]`. This is the "native" Rust API that implements the `RawStorage` trait from `embedded-svc` and it should be preferred actually, as you can layer on top of it any serde you want.
> 
> 

I learn that on #esp-rs:matrix.org but this is the explanation I needed. I could try to create an example for that too.

ivmarkov:
Closing, as we now have #274

PR #274, opened 2023-07-01
Add example that uses the C type of NVS API

punkto:


ivmarkov:
@punkto If you can please run `cargo fmt` and `cargo clippy` (Look at the CI as to how to run the latter.)

ivmarkov:
Thanks!

ivmarkov:
Hey @punkto, I'm so sorry. I force-pushed by mistake and for a short while, I lost your example. I restored it as-is, but the commit is listed as mine now. Again - sorry about that!

punkto:
Don't worry @ivmarkov . I'll have better luck with the raw+serde example (that I still have plans to write) :P

PR #276, opened 2023-07-18
Add NVS usage example using the native Rust API

punkto:
Please note that I have added a new dependence (postcard) just for the example. I don't know if that is OK for you or it is better to use something already added (json?).

### DIFF
--- a/examples/nvs_get_set_raw_storage.rs
+++ b/examples/nvs_get_set_raw_storage.rs
@@ -1,0 +1,128 @@
+//! Updates tags from the NVS (default partition) using the native Rust API.
+//!
+//! Note that this module exposes two separate set of APIs:
+//!  * the get_XXX/set_XXX API (where XXX is u8, str, etc.) - this is only for interop with C code that uses the C ESP IDF NVS API as well.
+//!  * the `get_raw`/`set_raw` APIs that take a `&[u8]`. This is the "native" Rust API that implements the `RawStorage` trait from `embedded-svc`
+//!     and it should be preferred actually, as you can layer on top of it any serde you want.
+//!
+//! More info regarding NVS:
+//!   https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/storage/nvs_flash.html
+
+use esp_idf_sys::{self as _}; // If using the `binstart` feature of `esp-idf-sys`, always keep this module imported
+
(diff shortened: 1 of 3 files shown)


ivmarkov:
Dependencies which are only relevant for the `[examples]` section should be added in the `[dev-dependencies]` section.

examples/nvs_get_set_raw_storage.rs
```diff
@@ -0,0 +1,128 @@
...
+    {
+        let tag_raw_str_data: &mut [u8] = &mut [0; 100];
+
+        match nvs.get_raw(tag_raw_str, tag_raw_str_data) {
+            Ok(v) => {
+                if let Some(the_str) = v {
+                    info!("{:?} = {:?}", tag_raw_str, from_bytes::<&str>(the_str));
+                }
+            }
+            Err(e) => info!("Couldn't get tag {} because{:?}", tag_raw_str, e),
```
ivmarkov, line 89: Nit: missing space

examples/nvs_get_set_raw_storage.rs
```diff
@@ -0,0 +1,128 @@
...
+    let test_namespace = "test_ns";
+    let mut nvs = match EspNvs::new(nvs_default_partition, test_namespace, true) {
+        Ok(nvs) => {
+            info!("Got namespace {:?} from default partition", test_namespace);
+            nvs
+        }
+        Err(e) => panic!("Could't get namespace {:?}", e),
+    };
+
+    let tag_raw_u8 = "test_raw_u8";
```
ivmarkov, line 43: Rather than `tag`s, can we name these things `key`s?

examples/nvs_get_set_raw_storage.rs
```diff
@@ -0,0 +1,128 @@
...
+            Ok(v) => {
+                if let Some(the_struct) = v {
+                    info!(
+                        "{:?} = {:?}",
+                        tag_raw_str,
+                        from_bytes::<StructToBeStored>(the_struct)
+                    )
+                }
+            }
+            Err(e) => info!("Couldn't get tag {} because{:?}", tag_raw_str, e),
```
ivmarkov, line 123: Ditto - missing space?

ivmarkov (changes requested):
Really just two small nits.

punkto:
done!

PR #288, opened 2023-08-23
Abstraction for synchronous esp-tls

torkleyy:
I think this should be pretty much good as-is, these TODOs remain:

* [x] Figure out how to map `crt_bundle_attach`
* [x] ~~Figure out how to map `ds_data`~~ (not implemented by this PR)
* [x] Is the naming okay / how could async be added in the future?
* [x] Module structure / `#[cfg(esp_idf_comp_esp_tls_enabled)]`
* [x] `non-block` config

Let me know if this PR is headed into the right direction.

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -1,22 +1,26 @@
 //! Type safe abstraction for esp-tls
 
-use std::{
+use core::{
     ffi::CStr,
     fmt::{Debug, Display},
-    io::{self, Read, Write},
+    time::Duration,
 };
 
+use embedded_svc::io;
+use esp_idf_sys::{EspError, ESP_ERR_INVALID_SIZE, ESP_ERR_NO_MEM, ESP_FAIL};
+
+use crate::errors::EspIOError;
+
+/// see https://www.ietf.org/rfc/rfc3280.txt ub-common-name-length
+const MAX_COMMON_NAME_LENGTH: usize = 64;
+
 pub struct EspTls {
-    raw: *mut esp_idf_sys::esp_tls,
+    reader: EspTlsRead,
+    writer: EspTlsWrite,
 }
 
 impl EspTls {
-    pub fn new(host: &str, port: i32, cfg: Config) -> Result<Self> {
-        // TODO: where to put async version? seperate struct?
-        let tls = unsafe { esp_idf_sys::esp_tls_init() };
-        if tls == std::ptr::null_mut() {
-            return Err(Error::BadAlloc);
-        }
+    pub fn new(host: &str, port: u16, cfg: Config) -> Result<Self, EspError> {
         let mut rcfg: esp_idf_sys::esp_tls_cfg = unsafe { std::mem::zeroed() };
 
         if let Some(ca_cert) = cfg.ca_cert {
@@ -40,13 +44,11 @@ impl EspTls {
             rcfg.clientkey_password_len = ckp.to_bytes().len() as u32;
         }
 
-        let mut alpn_protos: Vec<*const i8>;
+        // allow up to 9 protocols
+        let mut alpn_protos: [*const i8; 10];
+        let mut alpn_protos_cbuf = [0u8; 99];
         if let Some(protos) = cfg.alpn_protos {
-            alpn_protos = protos
-                .iter()
-                .map(|p| p.as_ptr())
-                .chain(std::iter::once(std::ptr::null()))
-                .collect();
+            alpn_protos = cstr_arr_from_str_slice(protos, &mut alpn_protos_cbuf)?;
             rcfg.alpn_protos = alpn_protos.as_mut_ptr();
         }
 
@@ -54,7 +56,12 @@ impl EspTls {
         rcfg.use_secure_element = cfg.use_secure_element;
         rcfg.timeout_ms = cfg.timeout_ms as i32;
         rcfg.use_global_ca_store = cfg.use_global_ca_store;
-        rcfg.common_name = cfg.common_name.as_ptr();
+
+        if let Some(common_name) = cfg.common_name {
+            let mut common_name_buf = [0; MAX_COMMON_NAME_LENGTH + 1];
+            rcfg.common_name = cstr_from_str(common_name, &mut common_name_buf).as_ptr();
+        }
+
         rcfg.skip_common_name = cfg.skip_common_name;
 
         let mut raw_kac: esp_idf_sys::tls_keep_alive_cfg;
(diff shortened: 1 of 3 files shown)


src/tls.rs
```diff
@@ -1,6 +1,234 @@
-//! TLS-related helper types
-use core::ffi::{c_char, CStr};
-use core::fmt::Debug;
+//! Type safe abstraction for esp-tls
+
+use std::{
```
ivmarkov, line 3: `use core::`

src/tls.rs
```diff
@@ -1,6 +1,234 @@
...
+    ffi::CStr,
+    fmt::{Debug, Display},
+    io::{self, Read, Write},
```
ivmarkov, line 6: Switch to the `io` traits from `embedded-svc` which re-export `embedded-io`. These are no-std compatible. They also have - behind a feature flag - mapping to `std::io::Read` / `std::io::Write`.

src/tls.rs
```diff
@@ -1,6 +1,234 @@
...
+
+            Err(Error::ConnectionNotEstablished)
+        }
+    }
+
+    pub fn close(self) {}
+}
+
+impl Read for EspTls {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
```
ivmarkov, line 109: * Implement `embedded_svc::io::Read` instead 
* Move `read` as `pub fn read` on `EspTls` and then delegate to this method from here

src/tls.rs
```diff
@@ -1,6 +1,234 @@
...
+        // ESP docs treat 0 as error, but in Rust it's common to return 0 from `Read::read` to indicate eof
+        if ret >= 0 {
+            Ok(ret as usize)
+        } else {
+            Err(io::Error::new(io::ErrorKind::Other, Error::ReadError(ret)))
+        }
+    }
+}
+
+impl Write for EspTls {
```
ivmarkov, line 124: * Implement `embedded_svc::io::Write` instead 
* Move `write` as `pub fn write` on `EspTls` and then delegate to this method from here

Additional topic: is `EspTls` supporting parallel `read`/`write`? I would assume so. This means it should be possible to `EspTls::split(&mut self) -> (EspTlsRead<_>, EspTlsWrite<'_>)`, so that you can handle read separately from write.

src/tls.rs
```diff
@@ -1,6 +1,234 @@
...
+    fmt::{Debug, Display},
+    io::{self, Read, Write},
+};
+
+pub struct EspTls {
+    raw: *mut esp_idf_sys::esp_tls,
+}
+
+impl EspTls {
+    pub fn new(host: &str, port: i32, cfg: Config) -> Result<Self> {
```
ivmarkov, line 14: * Nit: pass `Config` by ref, i.e. `&Config` as we do everywhere else
* Since we want to decouple this from STD, the return type should be `Result<Self, EspError>`. Do not introduce a custom `Error` type. We use everywhere just `EspError`.
* port should be `u16`

src/tls.rs
```diff
@@ -1,6 +1,234 @@
...
+            rcfg.__bindgen_anon_6.clientkey_bytes = client_key.data().len() as u32;
+        }
+
+        if let Some(ckp) = cfg.client_key_password {
+            // TODO: maybe this is not actually required to be NUL-terminated
+            rcfg.clientkey_password = ckp.as_ptr() as *const u8;
+            rcfg.clientkey_password_len = ckp.to_bytes().len() as u32;
+        }
+
+        let mut alpn_protos: Vec<*const i8>;
```
ivmarkov, line 43: Can we get by with `heapless::Vec` or `[*const i8; N]` array? Requiring `alloc` just for this does not seem reasonable.

src/tls.rs
```diff
@@ -1,6 +1,234 @@
...
+impl Drop for EspTls {
+    fn drop(&mut self) {
+        unsafe {
+            esp_idf_sys::esp_tls_conn_destroy(self.raw);
+        }
+    }
+}
+
+pub struct Config<'a> {
+    pub alpn_protos: Option<&'a [&'a CStr]>,
```
ivmarkov, line 152: Ideally - `&'a str` instead of `&'a CStr`. That might require `alloc` though as the conversion from `&str` to `&CStr` needs `CString`.

src/tls.rs
```diff
@@ -1,6 +1,234 @@
...
+    /// Keep-alive packet retry send count
+    pub count: u32,
+}
+
+pub struct PskHintKey<'a> {
+    pub key: &'a [u8],
+    pub hint: &'a CStr,
+}
+
+#[derive(Clone, Debug)]
```
ivmarkov, line 210: As per above, please do not introduce a custom error type. Specifically when implementing the `embedded_svc::io` traits, you can plug in there `EspIOError`, which is a wrapper around `EspError`.

ivmarkov (changes requested):
That's a great idea! 

I've left a bunch of (mostly stylistic / code reshuffling) comments, but the approach is solid.

Maybe the one important topic is whether it would be possible to also expose an API, that "adopts" an already existing socket, and does the TLS negotiation on top of it. But we can survive without this initially, I believe.

ivmarkov:
> * [ ]  Is the naming okay / how could async be added in the future?

For the naming - yes, that's fine.

As for `async`, adding it - as in - `async read` / `async write` should be somehow possible, but I'm not clear on details:
* The easy part: setting `cfg.non_block` to `true` is likely to switch the underlying socket to `O_NONBLOCK` mode post initial TLS negotiation (which itself is either sync or async, but for the async TLS negotiation we can think later). 
* The question is more about how to plug the underlying `sock_fd` into the `smol` reactor (`async-io` + `polling`) or into the upcoming support for tokio's reactor.  If your `EspTls` implementation was capable of "adopting" an already opened  socket, that might have been easier, as we would be using e.g. `Async<std::net::TcpSocket>` (or rather, its `T: OwnedFd` trait) which would be already registered with `smol`'s reactor and in non-blocking mode, so the TLS traffic would just flow on top then?

src/tls.rs
torkleyy: I switched to `embedded-io` but I couldn't find where it implements the std traits for me. It has a `std` flag but I couldn't find a blanket impl.

ivmarkov: [Here](https://github.com/embassy-rs/embedded-io/blob/v0.4.0/src/adapters/std_io.rs). Don't be afraid of `embedded-io` being archived. We are for now stuck with version 0.4, but we'll upgrade soon-ish. In the meantime they migrated it to `embedded-hal`, as a sub-project.  The STD adaptors in the `embedded-hal` workspace are in a separate project, but we'll upgrade to these when we migrate to the newer embedded-io (0.5). No rush.

ivmarkov:
One more thing: you need to replace `use esp_idf_sys::*` with `use crate::sys::*`, as in the meantime, both `esp-idf-hal` and `esp-idf-svc` re-export their `esp-idf-sys` dependency as `sys`. (And `esp-idf-svc` re-exports `esp-idf-hal` as `hal`.)

Long term this means most folks would depend only on `esp-idf-svc` (as they almost always would need a service or two from it anyway), and would use `esp-idf-hal` and `esp-idf-sys` via the `esp_idf_svc::sys` and `esp_idf_svc::hal` re-exports.

src/tls.rs
```diff
@@ -1,6 +1,354 @@
...
+    pub key: &'a [u8],
+    pub hint: &'a CStr,
+}
+
+/// str to cstr, will be truncated if str is larger than buf.len() - 1
+///
+/// # Panics
+///
+/// * Panics if buffer is empty.
+fn cstr_from_str<'a>(rust_str: &str, buf: &'a mut [u8]) -> &'a CStr {
```
ivmarkov, line 306: Can you move these to `private::cstr`?

src/tls.rs
```diff
@@ -1,6 +1,354 @@
...
+    CStr::from_bytes_with_nul(&buf[..truncated_str.len() + 1]).unwrap()
+}
+
+/// Convert slice of rust strs to NULL-terminated fixed size array of c string pointers
+///
+/// # Panics
+///
+/// * Panics if cbuf is empty.
+/// * Panics if N is <= 1
+fn cstr_arr_from_str_slice<const N: usize>(
```
ivmarkov, line 323: Ditto.

src/tls.rs
```diff
@@ -48,3 +396,57 @@ impl<'a> Debug for X509<'a> {
         write!(f, "X509(...)")
     }
 }
+
+#[cfg(test)]
```
ivmarkov, line 400: Ditto.

src/tls.rs
```diff
@@ -1,6 +1,354 @@
...
+    }
+}
+
+pub struct Config<'a> {
+    /// up to 9 ALPNs allowed, with avg 10 bytes for each name
+    pub alpn_protos: Option<&'a [&'a str]>,
+    pub ca_cert: Option<X509<'a>>,
+    pub client_cert: Option<X509<'a>>,
+    pub client_key: Option<X509<'a>>,
+    pub client_key_password: Option<&'a CStr>,
```
ivmarkov, line 245: Shouldn't we have a `&'a str` here as well?

src/tls.rs
```diff
@@ -1,6 +1,354 @@
...
+    ///
+    /// This function establishes a TLS/SSL connection with the specified host in a blocking manner.
+    ///
+    /// # Errors
+    ///
+    /// * `ESP_ERR_INVALID_SIZE` if `cfg.alpn_protos` exceeds 9 elements or avg 10 bytes/ALPN
+    /// * `ESP_ERR_NO_MEM` if TLS context could not be allocated
+    /// * `ESP_FAIL` if connection could not be established
+    pub fn new(host: &str, port: u16, cfg: &Config) -> Result<Self, EspError> {
+        let mut rcfg: esp_idf_sys::esp_tls_cfg = unsafe { std::mem::zeroed() };
```
ivmarkov, line 36: `crate::sys` - here and everywhere below. In fact, I am surprised how the `esp_idf_sys::` code even compiles, given that latest `master` is not even directly depending on `esp-idf-sys` anymore?

ivmarkov: UPDATE: Yeah, it is not, CI failed.

src/tls.rs
```diff
@@ -1,6 +1,354 @@
...
+}
+
+impl EspTlsRead {
+    pub fn read(&mut self, buf: &mut [u8]) -> Result<usize, EspIOError> {
+        // cannot call esp_tls_conn_read bc it's inline
+        //let ret = unsafe { esp_idf_sys::esp_tls_conn_read(self.raw, buf.as_mut_ptr(), buf.len()) };
+        let esp_tls = unsafe { std::ptr::read_unaligned(self.raw) };
+        let read_func = esp_tls.read.unwrap();
+        let ret = unsafe { read_func(self.raw, buf.as_mut_ptr() as *mut i8, buf.len()) };
+        // ESP docs treat 0 as error, but in Rust it's common to return 0 from `Read::read` to indicate eof
```
ivmarkov, line 179: But then you should check for `buf.is_empty()` at the beginning of the function and return Ok(0).

src/tls.rs
```diff
@@ -1,6 +1,354 @@
...
+    raw: *mut esp_idf_sys::esp_tls,
+}
+
+impl EspTlsWrite {
+    pub fn write(&mut self, buf: &[u8]) -> Result<usize, EspIOError> {
+        // cannot call esp_tls_conn_write bc it's inline
+        //let ret = unsafe { esp_idf_sys::esp_tls_conn_write(self.raw, buf.as_ptr(), buf.len()) };
+        let esp_tls = unsafe { std::ptr::read_unaligned(self.raw) };
+        let write_func = esp_tls.write.unwrap();
+        let ret = unsafe { write_func(self.raw, buf.as_ptr() as *const i8, buf.len()) };
```
ivmarkov, line 208: Ditto, if ESP-TLS treats writing a 0-length buffer as an error. That's OK in Rust, and in that case we should return `Ok(0)`, as that's how the `Read`/`Write` traits contract says.

ivmarkov (changes requested):

src/tls.rs
torkleyy: Ah right, there are two cases:

1. The buffer provided is empty
2. The connection was closed (thus eof)

torkleyy (commented):
I think there's one thing remaining, most of the tls module should be guarded by a cfg, but we can't do it for the full module because X509 is exposed unconditionally.

I could split `EspTls` etc into a separate module and then reexport it into `tls`, what do you think?

src/tls.rs
```diff
@@ -1,6 +1,294 @@
...
+
+#[derive(Default)]
+pub struct Config<'a> {
+    /// up to 9 ALPNs allowed, with avg 10 bytes for each name
+    pub alpn_protos: Option<&'a [&'a str]>,
+    pub ca_cert: Option<X509<'a>>,
+    pub client_cert: Option<X509<'a>>,
+    pub client_key: Option<X509<'a>>,
+    pub client_key_password: Option<&'a str>,
+    pub non_block: bool,
```
torkleyy, line 259: If this is `true`, should we call `conn_new_async` instead?

ivmarkov:
@torkleyy Build fails because you are still using aliases from the `std::` module. Switch these to `core::`.

ivmarkov:
> I think there's one thing remaining, most of the tls module should be guarded by a cfg, but we can't do it for the full module because X509 is exposed unconditionally.
> 
> I could split `EspTls` etc into a separate module and then reexport it into `tls`, what do you think?

Yeah, just don't create extra files. Surround your code with 
```rust
[cfg("esp_idf_comp_esp_tls_enabled")]
mod esptls {
...
}
``` 

... and then at the top, do
```rust
[cfg("esp_idf_comp_esp_tls_enabled")]
pub use esptls::*;
```

src/tls.rs
ivmarkov: Not necessarily, but **VERY IMPORTANT**: I was [looking at the code]() and I think I just figured out a way for us to pass an already opened socket!

I think we still should keep the existing `new`, but we should introduce (maybe post-merge) a new STD-only method along the lines of
```rust
#[cfg(feature = "std")]
pub fn wrap(socket: std::net::TcpSocket, cfg: &Config) -> Result<EspError, Self> {
    let tls = esp!(unsafe { esp_tls_init(); })?;
    esp!(unsafe { esp_tls_set_conn_state(tls, esp_tls_conn_state_t_ESP_TLS_CONNECTING) })?;
    esp!(unsafe { esp_tls_set_conn_sockfd(tls, socket.fd)? })?;
    esp!(unsafe { esp_tls_conn_new_sync(b"\0", 0, 0, cfg, tls) })?;

    Ok(...)
}
```

The above ^^^ will not try to open a new socket and will re-use ours, as the state of the connection is `esp_tls_conn_state_t_ESP_TLS_CONNECTING`.

The above approach is important, because it might be a way to implement `async` as well. Again, to do `async` in a way which is independent of the concrete async Reactor, we needed a way to pass-in an already opened socket handle, and the above might be just this.

ivmarkov: To not make the above STD-only, we can even pass to the method a raw file descriptor, where the assumption is that the file descriptor is a socket descriptor, and also it is already opened.

torkleyy:
Build is looking good, finally 😅

ivmarkov:
@torkleyy JFYI I'll be pushing changes to the API in the next hours. Two reasons:
* I found a way to "adopt" existing socket, including for async, so I have an async implementation ready and working (kind of, there is a bug/something not well thought out in the API of `esp_tls_connect` - it returns 0 where it should return ESP_TLS_WANT READ or ESP_TLS_WANT_WRITE)
* `split` is unsound and broken. Turns out, you cannot do overlapping blocking read/write with mbedtls, because the mbedtls SSL context is not thread-safe. The only way to do overlapping read/write with mbedtls is not-blocking (and thus also async). Sounds weird, but this also means the only way to do blocking read/write is to use the async implementation, and then "synchronize" it by using `block_on`

torkleyy:
@ivmarkov Alright, thanks for letting me know!

PR #289, opened 2023-08-26
Add example for ethernet

torkleyy:


torkleyy:
Just noticed that most of the examples are missing `link_patches`, should we add it?

ivmarkov:
Number and type of GPIOs are MCU-specific.
You need to `#[cfg(esp32)]`, `#[cfg(esp32c3)]` and so on. Or you can only do a single `#[cfg(esp32)]` and for the other MCUs panic with an error that the user needs to adjust the GPIOs for their MCUs in the example and then run it.

examples/eth.rs
```diff
@@ -0,0 +1,65 @@
...
+    eventloop::EspSystemEventLoop,
+    hal::{gpio, prelude::Peripherals},
+    log::EspLogger,
+};
+#[cfg(esp32)]
+use log::info;
+
+#[cfg(esp32)]
+fn main() -> anyhow::Result<()> {
+    esp_idf_sys::link_patches();
```
ivmarkov, line 15: `esp_idf_svc::sys::` here and everywhere below.

examples/eth.rs
```diff
@@ -0,0 +1,65 @@
+use esp_idf_sys::{self as _}; // If using the `binstart` feature of `esp-idf-sys`, always keep this module imported
```
ivmarkov, line 1: Not necessary - remove.

examples/eth.rs
```diff
@@ -0,0 +1,65 @@
...
+        pins.gpio23,
+        pins.gpio22,
+        pins.gpio21,
+        pins.gpio19,
+        pins.gpio18,
+        esp_idf_svc::eth::RmiiClockConfig::<gpio::Gpio0, gpio::Gpio16, gpio::Gpio17>::OutputInvertedGpio17(
+            pins.gpio17,
+        ),
+        Some(pins.gpio5),
+        esp_idf_svc::eth::RmiiEthChipset::LAN87XX,
```
ivmarkov, line 37: Out of pure curiosity - is this working? Another user on the matrix room had issues with precisely this chipset, you might want to chime in there for help.

ivmarkov (changes requested):

examples/eth.rs
torkleyy: Yes, this is working on Olimex ESP32-Gateway, but I didn't do anything special other than using these specific pins.

examples/eth.rs
ivmarkov: Hm. How about `use esp_idf_svc::sys::{self as _};`? If same story, I'll need to see what is going on...

PR #290, opened 2023-08-27
Experimental BT support

ivmarkov:
* For now, only BT Classic is supported (via the Bluedroid APIs) - and - only a subset of the profiles (A2DP sink, AVRC controller, HFP client).
* However, the rest can be easily added (it is just a laborious mapping of the various events of the other profiles).
* BLE support (with Bluedroid) is also in the works, but not compilable yet.

As for the approach - similar to what @pyalliet did [here](https://github.com/pyaillet/esp-idf-ble) but with extra type safety on the events' enums and without the "smartness" of auto-registering event callbacks and waiting on these in an async way.

All under the `experimental` feature flag.


ivmarkov:
BTW and if that's unclear - BT Classic unfortunately only works on the original ESP32 chip. All later variants support BLE only. Which rules out - for all other chips - nice stuff like A2DP (high quality audio streaming) and HFP (headset profile).

While experimental, I've tested it and it actually works quite nicely in combination with I2S for streaming audio output - and - say either I2S for digital microphone input, or the built-in ADC with the continuous ADC streaming driver, that also landed recently.

Issue #291, opened 2023-08-27
Wifi scanning fails seemingly due to invalid access point SSID


krigga:
I'm trying to run a program similar to https://github.com/esp-rs/std-training/blob/main/intro/http-client/examples/http_client.rs

After scanning, I see the following:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 11, error_len: Some(1) }', /home/krigga/.cargo/registry/src/index.crates.io-6f17d22bba15001f/esp-idf-svc-0.46.2/src/private/cstr.rs:30:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

abort() was called at PC 0x400fdac6 on core 0
0x400fdac6 - panic_abort::__rust_start_panic::abort
    at /home/krigga/.rustup/toolchains/esp/lib/rustlib/src/rust/library/panic_abort/src/lib.rs:42


Backtrace: 0x40082b12:0x3ffb9b10 0x400887e5:0x3ffb9b30 0x4008ec92:0x3ffb9b50 0x400fdac6:0x3ffb9bc0 0x400fdaba:0x3ffb9be0 0x400e1842:0x3ffb9c00 0x400e17f4:0x3ffb9c50 0x400eed97:0x3ffb9cc0 0x400eec5c:0x3ffb9cf0 0x400e1582:0x3ffb9d10 0x4010a364:0x3ffb9d40 0x4010be7b:0x3ffb9d80 0x400da09c:0x3ffb9de0 0x400da1c2:0x3ffb9e10 0x400dc452:0x3ffb9e40 0x400db965:0x3ffb9e90 0x400dbc49:0x3ffb9f40 0x400dba99:0x3ffb9ff0 0x400d816f:0x3ffba090 0x400d8098:0x3ffba0d0 0x400d8c82:0x3ffba150 0x400d60b6:0x3ffba170 0x400d5afd:0x3ffba500 0x40188bdf:0x3ffba530 0x400d7144:0x3ffba550 0x400ecd3a:0x3ffba570 0x400d7134:0x3ffba5a0 0x400d5b8b:0x3ffba5d0 0x400dee33:0x3ffba5f0 0x4018d0fb:0x3ffba610
0x40082b12 - panic_abort
    at /home/krigga/dev/tmp/esp-rs-test/test1/.embuild/espressif/esp-idf/v4.4.5/components/esp_system/panic.c:408
[...]

PR #292, opened 2023-08-27
Filter out bad wifi APs when scanning

krigga:
Aims to fix https://github.com/esp-rs/esp-idf-svc/issues/291, but probably requires more polishing

### DIFF
--- a/src/private/cstr.rs
+++ b/src/private/cstr.rs
@@ -5,6 +5,7 @@ extern crate alloc;
 pub use alloc::ffi::CString;
 
 pub use core::ffi::{c_char, CStr};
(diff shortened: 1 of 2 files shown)


ivmarkov:
I agree with you it is not ideal, but as a quick fix, it is good enough until we figure out how to properly address it.

BTW it is also a breaking change, as we no longer implement `From<Newtype<&wifi_ap_record_t>>`. Yet, I don't think anybody outside of `esp-idf-svc` is using these trait impls anyway so we can probably pretend this didn't happen (and next release of `esp-idf-svc` is likely to be a non-patch release anyway, so breaking a few things should be fine actually).

Thanks!

PR #294, opened 2023-08-29
Added Esp nvs ecryption

MartinBroers:
This patch implements the encryption API. It may still need some work, like putting stuff behind compiler options (I think if your sdkconfig does not contain the encryption parameters, it will fail compilation). Also, there is quite a lot of duplication, which feels wrong, but I both want to learn what is a good way to prevent it and I want to improve the code.

In order to run it, (ok, as I am writing this, it feels like I should add documentation as well...), build your app with the following keys in your defconfig:

```
CONFIG_SECURE_FLASH_ENC_ENABLED=y
CONFIG_SECURE_FLASH_ENCRYPTION_MODE_DEVELOPMENT=y
CONFIG_SECURE_FLASH_HAS_WRITE_PROTECTION_CACHE=y
CONFIG_SECURE_FLASH_UART_BOOTLOADER_ALLOW_ENC=y
CONFIG_SECURE_FLASH_SKIP_WRITE_PROTECTION_CACHE=y
CONFIG_SECURE_FLASH_CHECK_ENC_EN_IN_APP=y
CONFIG_PARTITION_TABLE_OFFSET=0x10000
CONFIG_NVS_ENCRYPTION=y
CONFIG_FLASH_ENCRYPTION_ENABLED=y
CONFIG_FLASH_ENCRYPTION_INSECURE=y
CONFIG_FLASH_ENCRYPTION_UART_BOOTLOADER_ALLOW_ENCRYPT=y
CONFIG_PARTITION_TABLE_MD5=y
```
Your partition table should look like
```
 # ESP-IDF Partition Table
 # Name,   Type, SubType, Offset,  Size, Flags
 nvs,      data, nvs,     0x11000,  0x4000,
 otadata,  data, ota,     0x15000,  0x2000,
 phy_init, data, phy,     0x17000,  0x1000,
 factory,  app,  factory, ,        0x280000,
 nvs_key,  data, nvs_keys, ,       0x1000, encrypted
 conf,     data, nvs,     ,        0x10000,
```

and even that ain't enough, since there is no complete support yet in the rust tooling, we'll need the partition.bin and bootloader.bin from the idf.py pipeline, I used examples/security sample for that. Just copy your partition table to there and build, copy the artifacts to some sane location, together with the output from `cargo build --save-image` and flash with 
```
esptool.py -p /dev/ttyUSB0 -b 921600 --before default_reset --after no_reset --chip esp32  write_flash --flash_mode dio --flash_size 4MB --flash_freq 40m --encrypt 0x1000 bootloader.bin 0x10000 partition-table.bin 0x20000 ota.bin
```

I am open help working on the last two steps, given some pointers where to start on them...

### DIFF
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -20,6 +20,7 @@ static NONDEFAULT_LOCKED: mutex::Mutex<alloc::collections::BTreeSet<CString>> =
 
 pub type EspDefaultNvsPartition = EspNvsPartition<NvsDefault>;
 pub type EspCustomNvsPartition = EspNvsPartition<NvsCustom>;
+pub type EspCustomEncNvsPartition = EspNvsPartition<NvsCustomEnc>;
 
 pub trait NvsPartitionId {
     fn is_default(&self) -> bool {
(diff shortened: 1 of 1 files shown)


torkleyy (commented):
Should use `log::*` instead of `println`. Also, I think the right cfg to use is probably `esp_idf_nvs_encryption`.

ivmarkov:
Thanks! Will merge as-is and will potentially do some cosmetic changes myself afterwards.

PR #298, opened 2023-09-11
Decode event data in WifiEvent

stabl-gjn:
…-masters-v0.46.0


ivmarkov:
Why did you close it?

stabl-gjn:
@ivmarkov sorry created on wrong repository, should have been on my private one.

ivmarkov:
Np. Would be good to contribute something like this at some point in time though.

PR #300, opened 2023-09-25
🗑️ add deprecate note on 4.3 specific cfg

Vollbrecht:


src/mqtt/client.rs
```diff
@@ -236,7 +236,6 @@ impl<'a> TryFrom<&'a MqttClientConfiguration<'a>>
                 verification: esp_mqtt_client_config_t_broker_t_verification_t {
                     use_global_ca_store: conf.use_global_ca_store,
                     skip_cert_common_name_check: conf.skip_cert_common_name_check,
-                    #[cfg(not(esp_idf_version = "4.3"))]
```
Vollbrecht, line 239: since the impl itself is for 
#[cfg(not(esp_idf_version_major = "4"))] this should be redundant

Vollbrecht, line 239: Since the impl is `#[cfg(not(esp_idf_version_major = "4"))]` this cfg should be redundant

PR #303, opened 2023-09-28
Implement Drop for EspOtaUpdate

kyrias:
This is required to free the resources allocated by `esp_ota_begin`.

### DIFF
--- a/src/ota.rs
+++ b/src/ota.rs
@@ -152,9 +152,14 @@ impl<'a> EspOtaUpdate<'a> {
         self.check_write()?;
 
         esp!(unsafe { esp_ota_end(self.update_handle) })?;
+        let update_partition = self.update_partition;
+
+        // `Drop::drop` must not be called on `EspOtaUpdate` after the OTA handle has been
+        // invalidated.
+        mem::forget(self);
 
         Ok(EspOtaUpdateFinished {
-            update_partition: self.update_partition,
+            update_partition,
             _data: PhantomData,
         })
     }
@@ -165,14 +170,15 @@ impl<'a> EspOtaUpdate<'a> {
         esp!(unsafe { esp_ota_end(self.update_handle) })?;
         esp!(unsafe { esp_ota_set_boot_partition(self.update_partition) })?;
 
+        // `Drop::drop` must not be called on `EspOtaUpdate` after the OTA handle has been
+        // invalidated.
+        mem::forget(self);
+
         Ok(())
     }
 
     pub fn abort(self) -> Result<(), EspError> {
-        self.check_write()?;
-
-        esp!(unsafe { esp_ota_abort(self.update_handle) })?;
-
+        // The OTA update is aborted when `EspOtaUpdate` is dropped.
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


src/ota.rs
```diff
@@ -185,6 +185,19 @@ impl<'a> EspOtaUpdate<'a> {
     }
 }
 
+impl Drop for EspOtaUpdate<'_> {
```
ivmarkov, line 188: I think you need to call `core::mem::forget(self)` at the end of the `abort()` and `complete()` methods, or else you'll end up calling `esp_ota_abort` twice (once from `abort` and once more from the dtor). Or you'll call `esp_ota_abort` post-completion, which should not happen either, if the user happens to invoke `complete` instead.

See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=f72c0b1023f12f0978f812888731f916).

kyrias: Ah, d'oh, of course.  Though, one downside of using `mem::forget` is that it's an easy thing to forget or accidentally remove.

Another option would be to instead make the `update_handle` field `Option<esp_ota_handle_t>`.  All methods that use it can `.take().unwrap()` it since we would never construct an instance without it set, but then the drop implementation can use `if let Some(handle) = self.update_handle.take()` to conditionally run it only if it was not already taken.  What do you think about that?

ivmarkov: I still think putting `forget` at two places is perhaps the simpler approach. Also what I found online as the recommended approach. :)

kyrias: Fair enough.  :+1:

Though if we define the behavior to be abort on drop then I guess the abort method can just be empty, no need to duplicate the abort implementation.

ivmarkov: Also an option, yes. Or just pass on the `abort` method completely, but that might mean rethinking the traits too, and given that a release is imminent, perhaps we can pass on that for now.

src/ota.rs
```diff
@@ -152,9 +152,14 @@ impl<'a> EspOtaUpdate<'a> {
         self.check_write()?;
 
         esp!(unsafe { esp_ota_end(self.update_handle) })?;
+        let update_partition = self.update_partition;
+
+        // `Drop::drop` must not be called on `EspOtaUpdate` after the OTA handle has been
+        // invalidated.
+        std::mem::forget(self);
```
ivmarkov, line 159: Only pls change to `core::mem::forget`, or else the no_std build would fail. Don't ask why we have no_std build... a bit of a hygiene (to make sure we are not overusing std apis where not really necessary), not much else.

src/ota.rs
```diff
@@ -165,14 +170,15 @@ impl<'a> EspOtaUpdate<'a> {
         esp!(unsafe { esp_ota_end(self.update_handle) })?;
         esp!(unsafe { esp_ota_set_boot_partition(self.update_partition) })?;
 
+        // `Drop::drop` must not be called on `EspOtaUpdate` after the OTA handle has been
+        // invalidated.
+        std::mem::forget(self);
+
```
ivmarkov, line 176: Ditto.

ivmarkov (changes requested):
It is OK, only pls change `std::` to `core::`.

PR #308, opened 2023-10-23
Gate uncased alloc feature behind this alloc feature

irriden:
Since uncased is only used in `http`, which itself depends on `alloc`, I thought another possibility could be to set `optional = true` on the `uncased` dependency, and only pull it when the `alloc` feature is raised. But this PR here is more straightforward.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ rust-version = "1.71"
 default = ["std", "native", "binstart"]
 
 std = ["alloc", "log/std", "esp-idf-hal/std", "embedded-svc/std"]
-alloc = ["esp-idf-hal/alloc", "embedded-svc/alloc"]
+alloc = ["esp-idf-hal/alloc", "embedded-svc/alloc", "uncased/alloc"]
 nightly = ["embedded-svc/nightly", "esp-idf-hal/nightly"]
 experimental = ["embedded-svc/experimental"]
 embassy-time-driver = ["embassy-time"]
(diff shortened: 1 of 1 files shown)


ivmarkov:
Yeah, why not.

Disclaimer though: 

Not really sure why you want to use the `no_std` + no-`alloc` mode of `esp-idf-svc`, but keep in mind that:
* Quite a few modules in `esp-idf-svc` do require `alloc` and hence they would be unavailable, so `no_std` + `alloc` is a better bet
* Even if you run without `alloc` for `esp-idf-*`, the ESP IDF C code **itself** would continue to do allocations. There is no "no-alloc" mode for ESP IDF
* Given that ESP IDF is supported directly in the Rust Standard Library, and the Rust Standard Library is **the** way - on top of ESP IDF - to spawn threads, do networking and use blocking synchronization primitives, you probably **do** want not just `alloc`, but also `std` enabled in your crate. (`esp-idf-svc` is only taking care of (a) compiling and linking ESP IDF itself (b) exposing APIs which are specific to ESP IDF and do not have counterparts in the Rust Standard Library (as in various drivers and services) and (c) exposing ALL ESP IDF raw C APIs in the `::sys` module

The `no_std` and `alloc`  `cfg`s and compilation mode in the `esp-idf-*` crates are more for an internal consumption by the crate authors themselves. As in, we don't have STD and allocations enabled by default, so that we can think really hard when/if we need allocations and/or functionality in Rust STD. This is kind of important on a device with < 400K of memory, as overusing allocations and STD types can lead to all sorts of issues.

ivmarkov:
> Thank you for the response @ivmarkov !
> 
> My goal here was to reduce the size of the binary as much as possible, for an app that runs just fine under `no_std + no-alloc`.
> 
> I was hoping that by specifying `no-alloc`, I would kick out any code related to a memory allocator from the binary.

The memory alocator just calls ESP IDF's `malloc`/`free` so no.
Not using individual components from `alloc` if you really don't have to, like `String`, `Vec`, `Arc` etc. will save you stuff.
 
> If I understand you correctly, this may be the case for the rust parts, but not from the esp-idf C parts, as they will continue to do allocations in a `no-alloc` setting.

Yes. Your best bet is:
* Try not to use everything and the kitchen sink from ESP IDF unless you really have to. Though once you pull Wifi and a couple of STD APIs, a lot gets linked in anyway, as they all call each other
* Be careful when pulling 3rd party crates. Though if your code is no-std/no-alloc you probably are, already
* Inspect your binary segments. You might have too much `rodata` because of error strings/logging strings. Forgot the name of the tool though, maybe ask in the forum

Issue #309, opened 2023-10-28
How can I post with chunked data?


ysak-y:
## Overview

Hi, I love this crate, thank you for providing eco-system to implement firmware by Rust.
btw, now I'm worrying how I send POST request with chunked data that is achieved by using `Transfer-Encoding: chunked` header and without `Content-Length` header basically.

## Goal

I want to write data multiple times for the request as follow.

```Rust
    let mut request = client.post(url, &headers).unwrap();
    request.write(b"a").unwrap();
    request.write(b"b").unwrap();
    request.write(b"c").unwrap();
    request.flush().unwrap(); -> Send request with "abc" payload.
```

## Problem

To send chunked data, I want to add `Transfer-Encoding: chunked` to the header and don't want to add `Content-Length` because I can't measure the total size of chunked data while sending requests.
But it seems this crate add `Content-Length` with `0` value if I don't set it to the header.

Example is as follows.
```Rust
    let payload = b"{\"hello\": \"world\"}";

    // Prepare headers and URL
    let headers = [
        ("content-type", "application/json"),
        ("Transfer-Encoding", "chunked"),
    ];
    let url = "http://192.168.86.48";
    let mut client = Client::wrap(EspHttpConnection::new(&Default::default()).unwrap());

    // Send request
    let mut request = client.post(url, &headers).unwrap();
    request.write(payload).unwrap();

    request.flush().unwrap();
    request.submit().unwrap();
```

Above implementation sends follow request (I captured it by `Wireshark`). It has `Content-Length: 0` in its header.
```
POST / HTTP/1.1
User-Agent: ESP32 HTTP Client/1.0
Host: 192.168.86.48
content-type: application/json
Transfer-Encoding: chunked
Content-Length: 0

HTTP/1.1 400 Bad Request
Connection: close

{"hello": "world"}
```

## Solution

Don't add `Content-Length: 0` to the header even if there is no `Content-Length`. Or maybe some other ways to send request with chunked data?

ysak-y:
I guess this line set `Content-Length` to 0.
https://github.com/esp-rs/esp-idf-svc/blob/master/src/http/client.rs#L249

This value is used for `esp_http_client_open` method, maybe I need to investigate about it.
https://github.com/esp-rs/esp-idf-svc/blob/master/src/http/client.rs#L251

Vollbrecht:
[...]

PR #310, opened 2023-10-28
Set negative value to second argument of `esp_http_client_open` method if request header has `Transfer-Encoding: chunked`

ysak-y:
## What

Set nevative value to `request_content_len` property if can't unwrap `content_len` somehow.

## Why

Ref: https://github.com/esp-rs/esp-idf-svc/issues/309

I want to implement streaming http post that sets `Transfer-Encoding: chunked` to the header and doesn't set `Content-Length`. But this library sets `Content-Length: 0` if I don't set `Content-Length` to the header and got 400 error (Please read above issue description).

As I investigate, I can avoid this behavior and can get what I'm looking for if set negative value to the second argument of  `esp_http_client_open` method.
https://github.com/esp-rs/esp-idf-svc/issues/309#issuecomment-1783754746

So I want to set negative value if can't fetch correct value from `content_len`.

## How

Set `request_content_len` type to i64 because I want to set negative value to it, and set `request_content_len` property to -1 (Some negative value) if can't unwrap `content_len`.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -101,7 +101,7 @@ pub struct EspHttpConnection {
     follow_redirects_policy: FollowRedirectsPolicy,
     event_handler: Box<Option<Box<dyn Fn(&esp_http_client_event_t) -> esp_err_t>>>,
     state: State,
-    request_content_len: i32,
+    request_content_len: u64,
     follow_redirects: bool,
     headers: BTreeMap<Uncased<'static>, String>,
     content_len_header: UnsafeCell<Option<Option<String>>>,
@@ -221,7 +221,7 @@ impl EspHttpConnection {
 
         for (name, value) in headers {
             if name.eq_ignore_ascii_case("Content-Length") {
-                if let Ok(len) = value.parse::<i32>() {
+                if let Ok(len) = value.parse::<u64>() {
                     content_len = Some(len);
                 }
             }
@@ -254,7 +254,7 @@ impl EspHttpConnection {
                 if self.is_chunked_streaming_request(headers) {
                     -1
                 } else {
-                    self.request_content_len
+                    self.request_content_len as i32
                 },
             )
         })?;
@@ -408,7 +408,7 @@ impl EspHttpConnection {
                     })?;
                     esp!(unsafe { esp_http_client_set_redirection(self.raw_client) })?;
                     esp!(unsafe {
-                        esp_http_client_open(self.raw_client, self.request_content_len)
+                        esp_http_client_open(self.raw_client, self.request_content_len as i32)
                     })?;
 
                     self.headers.clear();
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -248,7 +248,16 @@ impl EspHttpConnection {
 
         self.request_content_len = content_len.unwrap_or(0);
 
-        esp!(unsafe { esp_http_client_open(self.raw_client, self.request_content_len) })?;
+        esp!(unsafe {
+            esp_http_client_open(
+                self.raw_client,
+                if self.is_chunked_streaming_request(headers) {
+                    -1
+                } else {
+                    self.request_content_len
+                },
+            )
+        })?;
 
         self.state = State::Request;
 
@@ -441,6 +450,18 @@ impl EspHttpConnection {
         if self.state != State::Response {
             panic!("connection is not in response phase");
         }
+    }
+
+    fn is_chunked_streaming_request(&self, headers: &[(&str, &str)]) -> bool {
+        for (name, value) in headers {
+            if name.eq_ignore_ascii_case("Transfer-Encoding")
+                && value.eq_ignore_ascii_case("chunked")
+            {
+                return true;
+            }
+        }
+
+        false
     }
 }
 
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -101,7 +101,7 @@ pub struct EspHttpConnection {
     follow_redirects_policy: FollowRedirectsPolicy,
     event_handler: Box<Option<Box<dyn Fn(&esp_http_client_event_t) -> esp_err_t>>>,
     state: State,
-    request_content_len: u64,
+    request_content_len: i64,
     follow_redirects: bool,
     headers: BTreeMap<Uncased<'static>, String>,
     content_len_header: UnsafeCell<Option<Option<String>>>,
@@ -221,7 +221,7 @@ impl EspHttpConnection {
 
         for (name, value) in headers {
             if name.eq_ignore_ascii_case("Content-Length") {
-                if let Ok(len) = value.parse::<u64>() {
+                if let Ok(len) = value.parse::<i64>() {
                     content_len = Some(len);
                 }
             }
@@ -246,18 +246,9 @@ impl EspHttpConnection {
             _ => false,
         };
 
-        self.request_content_len = content_len.unwrap_or(0);
+        self.request_content_len = content_len.unwrap_or(-1);
 
-        esp!(unsafe {
-            esp_http_client_open(
-                self.raw_client,
-                if self.is_chunked_streaming_request(headers) {
-                    -1
-                } else {
-                    self.request_content_len as i32
-                },
-            )
-        })?;
+        esp!(unsafe { esp_http_client_open(self.raw_client, self.request_content_len as i32) })?;
 
         self.state = State::Request;
 
@@ -450,18 +441,6 @@ impl EspHttpConnection {
         if self.state != State::Response {
             panic!("connection is not in response phase");
         }
-    }
-
-    fn is_chunked_streaming_request(&self, headers: &[(&str, &str)]) -> bool {
-        for (name, value) in headers {
-            if name.eq_ignore_ascii_case("Transfer-Encoding")
-                && value.eq_ignore_ascii_case("chunked")
-            {
-                return true;
-            }
-        }
-
-        false
     }
 }
 
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -101,7 +101,7 @@ pub struct EspHttpConnection {
     follow_redirects_policy: FollowRedirectsPolicy,
     event_handler: Box<Option<Box<dyn Fn(&esp_http_client_event_t) -> esp_err_t>>>,
     state: State,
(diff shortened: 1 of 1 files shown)


ysak-y:
I will test it more and ask someone to review it later.

ivmarkov:
I don't get it. Do you want to implement chunked encoding **yourself**? Because the built-in ESP IDF HTTP client does not support it.

ivmarkov:
Actually I stand corrected. Chunked encoding DOES seems to be supported, after all.

Vollbrecht:
yeah we just need to make sure to then not set the content-length header ourself. Also we might need to check if that is true only for the latest master / 5.1 or also included inside the 4.4.6 client so @ysak-y if you would check if it would work on both versions

ivmarkov:
@ysak-y if you plan to finish this, you need to address the CI build errors.

ysak-y:
Thank you for your feedbacks. I'll try to implement it soon. Please just a while.

> @ysak-y if you plan to finish this, you need to address the CI build errors.

Yes, I know. I'll try to pass the CI later.

ysak-y:
# UPDATE

I investigated `esp-idf-svc` more, and I implemented it as follows finally.

## [a86d5f7](https://github.com/esp-rs/esp-idf-svc/pull/310/commits/a86d5f78404bff8abdd11783e31e446e2502cec0)

Convert type of `request_content_len` property from `u64` to `i32` to accept negative value. In the first place, I think this property should be `i32` type because it will be cast to `i32` finally. It is described in https://github.com/esp-rs/esp-idf-svc/pull/311 .

## [0da0e59](https://github.com/esp-rs/esp-idf-svc/pull/310/commits/0da0e596f0c89cd3e6ccef6971daacf9604655b2)

Implement `is_chunked_streaming_request` method to check whether header has `Transfer-Encoding: chunked` in it. Then pass `-1` if header has it, otherwise, `self.request_content_len` value that is an existing behavior.

I thought `is_chunked_streaming_request` should check `Content-Length` header at first, but it seems we can ignore it if `Transfer-Encoding: chunked` is in its header. So `is_chunked_streaming_request` checks only `Transfer-Encoding` header.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#directives

# Example

I implement example code as follows (This is only HTTP part, not full example), and then got 200 response, and server accepted `Hello World!!` value!!. It works on ESP-IDF `v5.1` and `v4.4.6` also.

```Rust
    let payload = b"Hello";
    let payload2 = b" World!!";

    let headers = vec![
        ("Content-Type", "plain/text"),
        ("Transfer-Encoding", "chunked"),
    ];
    let mut request = clinet.post(URL, &headers).unwrap();

    // Write payload
    request
        .write(format!("{:X}", payload.len()).as_bytes())
        .unwrap();
    request.write(b"\r\n").unwrap();
    request.write(payload).unwrap();
    request.write(b"\r\n").unwrap();

    // Write payload2
    request
        .write(format!("{:X}", payload2.len()).as_bytes())
        .unwrap();
    request.write(b"\r\n").unwrap();
    request.write(payload2).unwrap();
    request.write(b"\r\n").unwrap();

    // Write ending
    request.write(b"0\r\n\r\n").unwrap();

    request.flush().unwrap();

    let response = request.submit().unwrap();
```

Please refer following documents if you don't know how to write chunked data with streaming post.
https://datatracker.ietf.org/doc/html/rfc9112#name-chunked-transfer-coding

ysak-y:
> I suggested there instead of i64 to do Option<u64>. That might or might not be a good idea. Please think about that, and try to come - in the other PR - with a complete solution implementing chunked encoding for POST/PUT requests.
https://github.com/esp-rs/esp-idf-svc/pull/311#issuecomment-1784607622

I think `i32` would be better because `esp_http_client_open` expects `i32` type, and `request_content_len` is only used for it. What do you think? @ivmarkov

ivmarkov:
> > I suggested there instead of i64 to do Option. That might or might not be a good idea. Please think about that, and try to come - in the other PR - with a complete solution implementing chunked encoding for POST/PUT requests.
> > [#311 (comment)](https://github.com/esp-rs/esp-idf-svc/pull/311#issuecomment-1784607622)
> 
> I think `i32` would be better because `esp_http_client_open` expects `i32` type, and `request_content_len` is only used for it. What do you think? @ivmarkov

I don't like it very much as i32 is "too wide". What does it even mean if the user passes -2 as a value?
Also the fact that ESP IDF expects i32 does not mean that our Rust wrapper should expect i32.

ysak-y:
> Also the fact that ESP IDF expects i32 does not mean that our Rust wrapper should expect i32.

Ah, make sense. I agree with your opinion. I'll consider better way.

ysak-y:
@ivmarkov I modify `request_content_len` type to `u64` and cast it to `i32` expressly when pass to `esp_http_client_open` method. Does it make sense to you? If so, I'll align commits of this PR (e.g. We don't need [a86d5f7](https://github.com/esp-rs/esp-idf-svc/pull/310/commits/a86d5f78404bff8abdd11783e31e446e2502cec0) anymore).
[adeaca5](https://github.com/esp-rs/esp-idf-svc/pull/310/commits/adeaca53a1a1d38fd8c0c316fe750f181687fdd4)

ysak-y:
> While the HTTP spec discourage it, to have both set, the client driver wouldn't probably complain.

Ah... nice catch. My idea is just emit a warning or throw error. But, it may be server's responsibility. Actually, in NodeJS, built-in http server throws error when request has both (`Content-Length` and `Transfer-Encoding`) headers but it wouldn't throw error even if request has both when sending http request (https://nodejs.org/api/http.html).

I think basically client doesn't validate header values because there are some specification differences between server. We can notice mistakes if we get 400 BAD REQUEST error from server.

src/http/client.rs
```diff
@@ -442,6 +451,18 @@ impl EspHttpConnection {
             panic!("connection is not in response phase");
         }
     }
+
```
ivmarkov, line 454: I think you don't need this method at all. [This post](https://www.esp32.com/viewtopic.php?t=17314) says, that as long as you pass `-1` to `esp_http_clienbt_open`, the client would assume chunked encoding.

src/http/client.rs
```diff
@@ -399,7 +408,7 @@ impl EspHttpConnection {
                     })?;
                     esp!(unsafe { esp_http_client_set_redirection(self.raw_client) })?;
                     esp!(unsafe {
-                        esp_http_client_open(self.raw_client, self.request_content_len as _)
+                        esp_http_client_open(self.raw_client, self.request_content_len as i32)
```
ivmarkov, line 411: Please just change the type of `EspHttpConnection::request_content_len` to `i64`, as you originally started doing. Sorry for asking you to do other stuff - I did not realize, that `request_content_len` is a private field which is not visible to the user.

src/http/client.rs
```diff
@@ -248,7 +248,16 @@ impl EspHttpConnection {
 
         self.request_content_len = content_len.unwrap_or(0);
```
ivmarkov, line 249: Assuming you'll change `request_content_len` to `i64`, as suggested below, just do here `unwrap_or(-1)`

src/http/client.rs
```diff
@@ -248,7 +248,16 @@ impl EspHttpConnection {
...
 
-        esp!(unsafe { esp_http_client_open(self.raw_client, self.request_content_len as _) })?;
+        esp!(unsafe {
+            esp_http_client_open(
+                self.raw_client,
+                if self.is_chunked_streaming_request(headers) {
```
ivmarkov, line 254: Remove the `if`. Assuming `is_chunked_streaming_request` is removed, and then unwrapping `content-len` just sets `request_content_len` to -1 if `content_len` is `None`, you'll no longer need it.

ivmarkov (changes requested):
Assuming you do the small changes suggested below, I think can merge.

What I actually STILL don't like in this approach is that the chunked encoding **is left to the user**. I get it that the ESP IDF client does not do it by default, but this is still annoying. For one, the ESP IDF **server** automatically decodes/encodes in chunked encoding, which is asymmetric to their client. I think also the ESP IDF client automatically **decodes** chunked encoding, so it is weird that it cannot encode it, no?

src/http/client.rs
ysak-y: Okay, change type to `i64`.
[a7d084c](https://github.com/esp-rs/esp-idf-svc/pull/310/commits/a7d084cb4c792362c1dbeb33b91476cd1ac8492c)

ysak-y:
As you commented, change type of `request_content_len` to `i64` and just pass `-1` if can't unwrap `content_len` variable. Then remove `is_chunked_streaming_request` because it is no longer needed. Does it make sense to you? @ivmarkov
If it seems okay, I'll rebase commits to merge soon.

[a7d084c](https://github.com/esp-rs/esp-idf-svc/pull/310/commits/a7d084cb4c792362c1dbeb33b91476cd1ac8492c)

> I think also the ESP IDF client automatically decodes chunked encoding, so it is weird that it cannot encode it, no?

I agree with it. But as you say, original `ESP-IDF` doesn't implement it now. I think `esp-idf-svc` shouldn't implement original encoding method because it would be complicated if `ESP-IDF` implements it later. `esp-idf-svc` should be just the wrapper library of `ESP-IDF`.

Vollbrecht:
> esp-idf-svc should be just the wrapper library of ESP-IDF.

This maybe your first impression, but we often have an opinionated implementation over the original  raw idf api to make it first and foremost a good rust api. In fact the goal is to have our api stable compared to idf versions like between 4.4 and 5.1 and don't leak the original api out. Though in reality its not always so simple, so good compromises will win in the end.

ivmarkov:
@ysak-y @Vollbrecht The [crate now supports automatic chunked encoding](https://github.com/esp-rs/esp-idf-svc/commit/f966704549f80a4650dd55ad9a1cd35ca0f724f6) (user does not have to do it herself, unless that's explicitly requested using the new `Configuration::raw_request_body` parameter.

Also I had to fix a buglet from this PR whereas if no `Content-Length` was explicitly set by the user, the client was switched to `POST` + chunked encoding *even when the user was doing an innocent GET* and forgot to put a `Content-Length: 0` header.

PR #311, opened 2023-10-30
[nits] Convert request_content_len to i32 because it would be casted to i32 finally

ysak-y:
## What

Make type of `request_content_len` property in `EspHttpConnection` struct from `u64` to `i32`.

## Why

Now I'm working on https://github.com/esp-rs/esp-idf-svc/pull/310 . I noticed the second argument of `esp_http_client_open` method expects `i32` value, not `u64`. But `request_content_len` type is `u64`. I investigated `requset_content_len` property but it seems there is no reason that `request_content_len` type should be `u64`. So I want to modify it to `i32` because it would make more easier to implement https://github.com/esp-rs/esp-idf-svc/pull/310 .

When I remove `as _` from https://github.com/esp-rs/esp-idf-svc/blob/master/src/http/client.rs#L251 , then shows the error as below.
```
      |         esp!(unsafe { esp_http_client_open(self.raw_client, self.request_content_len) })?;
      |                       --------------------                  ^^^^^^^^^^^^^^^^^^^^^^^^ expected `i32`, found `u64`
      |                       |
      |                       arguments to this function are incorrect
```

## How

Replace `u64` type with `i32` of `request_content_len` property. Then remove `as _` syntax for them because now it has expected type.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -101,7 +101,7 @@ pub struct EspHttpConnection {
     follow_redirects_policy: FollowRedirectsPolicy,
     event_handler: Box<Option<Box<dyn Fn(&esp_http_client_event_t) -> esp_err_t>>>,
     state: State,
-    request_content_len: u64,
+    request_content_len: i32,
     follow_redirects: bool,
     headers: BTreeMap<Uncased<'static>, String>,
     content_len_header: UnsafeCell<Option<Option<String>>>,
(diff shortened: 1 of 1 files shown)


Vollbrecht:
To make the process a bit simpler can you make all changes you want to do, in your initial PR with consecutively commits.

ysak-y:
IMO, PRs should be created for a single purpose. Doing so makes it easier to revert or track PRs. However, I will follow your suggestion if it is better.

ivmarkov:
This PR is modifying the type of a **single private variable**. It does not have a purpose on its own, only as part of the other PR. Consolidate your work there please. And also follow my comments on the other PR - for example, I suggested there instead of `i64` to do `Option<u64>`. That might or might not be a good idea. Please think about that, and try to come - in the other PR - with a **complete solution** implementing chunked encoding for POST/PUT requests.

PR #316, opened 2023-11-13
✏️ Change the name of the field esp_idf_svc::http::server::Configurat…

Angelo13C:
Change the name of the field `esp_idf_svc::http::server::Configuration::max_resp_handlers` to `max_resp_headers`, because it's called in this way in [ESP-IDF's documentation](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/protocols/esp_http_server.html#_CPPv4N12httpd_config16max_resp_headersE))

### DIFF
--- a/src/http/server.rs
+++ b/src/http/server.rs
@@ -51,7 +51,7 @@ pub struct Configuration {
     pub stack_size: usize,
     pub max_open_sockets: usize,
     pub max_uri_handlers: usize,
-    pub max_resp_handlers: usize,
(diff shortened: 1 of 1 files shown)


ivmarkov:
It is a breaking change (API change) but given that we'll likely publish a breaking version of `esp-idf-svc` soon anyway, due to the upgrade to `heapless` 0.8, I think I'm fine with that.

Thanks!

PR #319, opened 2023-11-16
Set ping TTL to 64 (fixes #318)

### LINKED ISSUE #318: ping fails when going via a gateway(?)
I have a simple app which connects to wifi, pings an IP address, and lights up an LED based on how fast the response is.

* In the wokwi simulator, the device gets a 10.x.x.x IP address and can ping 1.1.1.1 and get responses just fine
* On an `esp32-c3-devkit-rust-1` (risc-v), as well as a `esp-wroom-32` (xtensa), I can ping hosts on the local network (192.168.1.x), but I can't ping hosts on the internet (1.1.1.1)

Running tcpdump on my access point, I notice something weird.

Here's the tcpdump when my laptop pings 1.1.1.1 (all good):
```
22:06:15.882597 IP 192.168.1.96 > 1.1.1.1: ICMP echo request, id 5273, seq 1, length 64
22:06:15.922553 IP 1.1.1.1 > 192.168.1.96: ICMP echo reply, id 5273, seq 1, length 64
22:06:16.885267 IP 192.168.1.96 > 1.1.1.1: ICMP echo request, id 5273, seq 2, length 64
22:06:16.927101 IP 1.1.1.1 > 192.168.1.96: ICMP echo reply, id 5273, seq 2, length 64
22:06:17.886193 IP 192.168.1.96 > 1.1.1.1: ICMP echo request, id 5273, seq 3, length 64
22:06:18.037990 IP 1.1.1.1 > 192.168.1.96: ICMP echo reply, id 5273, seq 3, length 64
```

And here's the tcpdump when my esp32 pings 1.1.1.1:
```
22:23:05.438439 IP 192.168.1.23 > 1.1.1.1: ICMP echo request, id 19580, seq 256, length 64
22:23:05.438661 IP 192.168.1.1 > 192.168.1.23: ICMP time exceeded in-transit, length 92
22:23:07.946762 IP 192.168.1.23 > 1.1.1.1: ICMP echo request, id 19580, seq 512, length 64
22:23:07.947317 IP 192.168.1.1 > 192.168.1.23: ICMP time exceeded in-transit, length 92
22:23:10.446840 IP 192.168.1.23 > 1.1.1.1: ICMP echo request, id 19580, seq 768, length 64
22:23:10.447511 IP 192.168.1.1 > 192.168.1.23: ICMP time exceeded in-transit, length 92
22:23:12.946633 IP 192.168.1.23 > 1.1.1.1: ICMP echo request, id 19580, seq 1024, length 64
22:23:12.947167 IP 192.168.1.1 > 192.168.1.23: ICMP time exceeded in-transit, length 92
22:23:15.446674 IP 192.168.1.23 > 1.1.1.1: ICMP echo request, id 19580, seq 1280, length 64
22:23:15.447205 IP 192.168.1.1 > 192.168.1.23: ICMP time exceeded in-transit, length 92
```

I see two things: the sequence numbers appear to be shifted by a couple of bits, and the only response is TTL timeout. I wonder if something is shifting the bits in the packet causing the sequence number to be too high and the TTL to be too low?

Here is my code: https://github.com/shish/esp-ping-leds/blob/main/src/main.rs

shish:
The default TTL is 0, ie pinging any host that isn't on the local network will fail - with this, it succeeds

### DIFF
--- a/src/ping.rs
+++ b/src/ping.rs
@@ -77,6 +77,7 @@ impl EspPing {
             task_stack_size: 4096,
             task_prio: 2,
             interface: self.0,
+            ttl: 64,
             ..Default::default()
         };
 


Vollbrecht:
yeah thanks for the PR TTL 0 is not a good default

Vollbrecht (approved):

PR #323, opened 2023-11-26
Add example demonstrating how to configure a static IPv4 address in lieu of DHCP

ssnover:
I had to do quite a bit of spelunking in the docs to figure this out and I think this is a fairly common use case (`esp-wifi` has [an example like this](https://github.com/esp-rs/esp-wifi/blob/main/esp-wifi/examples/static_ip.rs)). 

One question I had was why the API for swapping the network interface configs needs to swap both the STA and AP configs at once when many are not necessarily using both? It'd help clean up this example if that were possible and the code of anyone who isn't running an ESP device as both an AP and STA device.


examples/wifi_static_ip.rs
```diff
@@ -0,0 +1,111 @@
...
+fn main() -> anyhow::Result<()> {
+    esp_idf_svc::sys::link_patches();
+    EspLogger::initialize_default();
+
+    let peripherals = Peripherals::take()?;
+    let sys_loop = EspSystemEventLoop::take()?;
+    let nvs = EspDefaultNvsPartition::take()?;
+
+    let mut wifi = EspWifi::new(peripherals.modem, sys_loop.clone(), Some(nvs))?;
+    configure_wifi(&mut wifi);
```
ivmarkov, line 41: I think it would be much clearer if - rather than configuring `EspWifi` with default settings - and then - immediately swapping out the configured STA (and AP) netifs, you instead just **create** the `EspWifi` instance with your custom STA *from the very beginning*.

I.e. something like this (excuse potential typos):
```rust
let mut wifi = EspWifi::wrap_all(
    WifiDriver::new(peripherals.modem, sys_loop.clone(), Some(nvs))?,
    EspNetif::new_with_conf(&NetifConfiguration {
        ip_configuration: IpConfiguration::Client(IpClientConfiguration::Fixed(IpClientSettings {
            ip: static_ip,
            subnet: Subnet {
                gateway: gateway_addr,
                mask: Mask(netmask),
            },
            // Can also be set to Ipv4Addrs if you need DNS
            dns: None,
            secondary_dns: None,
        })),
        ..NetifStack::Sta::default_configuration()
    }),
    EspNetif::new(NetifStack::Ap),
)?;
```

examples/wifi_static_ip.rs
```diff
@@ -0,0 +1,111 @@
...
+            ip: static_ip,
+            subnet: Subnet {
+                gateway: gateway_addr,
+                mask: Mask(netmask),
+            },
+            // Can also be set to Ipv4Addrs if you need DNS
+            dns: None,
+            secondary_dns: None,
+        })),
+        route_priority: 0,
```
ivmarkov, line 85: Not ideal if you plan to plug-in eth as well, as the `eth` netif stack will *not* take priority. Best is to use the "default" STA/AP/ETH configurations and stacks and just "plug" in those your custom ipv4 configuration, as per above.

ivmarkov (changes requested):

ivmarkov:
> One question I had was why the API for swapping the network interface configs needs to swap both the STA and AP configs at once when many are not necessarily using both? It'd help clean up this example if that were possible and the code of anyone who isn't running an ESP device as both an AP and STA device.

Pure laziness on my side. I don't think there's anything wrong with exposing - e.g. - `swap_netif_sta` and `swap_netif_ap` - in addition to the existing `swap_netif` method.

examples/wifi_static_ip.rs
ssnover: Thank you for pointing me in the right direction, that is much more sensible!

examples/wifi_static_ip.rs
ssnover: Ah yes, while tinkering I defaulted it and did not revisit. Fixed.

PR #331, opened 2023-12-20
Add support for WPS

thetek42:


examples/wps_async.rs
```diff
@@ -0,0 +1,96 @@
+//! Example of using async wifi.
```
torkleyy, line 1:
```suggestion
//! Example of using async wps.
```

src/private/cstr.rs
```diff
@@ -19,6 +19,13 @@ pub fn set_str(buf: &mut [u8], s: &str) -> Result<(), crate::sys::EspError> {
     Ok(())
 }
 
+#[cfg(feature = "alloc")]
+pub fn set_cchar_str(buf: &mut [c_char], s: &str) -> Result<(), crate::sys::EspError> {
+    let buf_u8 = unsafe { buf.as_mut_ptr() as *mut u8 };
+    let slice = unsafe { std::slice::from_raw_parts_mut(buf_u8, buf.len()) };
+    set_str(slice, s)
```
torkleyy, line 26: I don't really understand why we should reuse that function, it seems it does an allocation (in `to_cstring_arg`), which I don't think is necessary.

src/wifi.rs
```diff
@@ -464,6 +482,11 @@ impl<'d> WifiDriver<'d> {
                 WifiEvent::StaConnected => guard.0 = WifiEvent::StaConnected,
                 WifiEvent::StaDisconnected => guard.0 = WifiEvent::StaDisconnected,
                 WifiEvent::ScanDone => guard.0 = WifiEvent::ScanDone,
+                WifiEvent::StaWpsSuccess(_)
+                | WifiEvent::StaWpsFailed
+                | WifiEvent::StaWpsTimeout
+                | WifiEvent::StaWpsPin(_)
+                | WifiEvent::StaWpsPbcOverlap => guard.2 = Some(event.try_into().unwrap()),
```
torkleyy, line 489: Can we destructure the tuple? It's becoming a bit hard too read imo

src/wifi.rs
```diff
@@ -1050,6 +1073,41 @@ impl<'d> WifiDriver<'d> {
...
 
+    /// Enable and start WPS
+    pub fn start_wps(&mut self, config: &WpsConfig) -> Result<(), EspError> {
+        let config = Newtype::<esp_wps_config_t>::try_from(config)?;
+
+        self.set_configuration(&Configuration::None)?;
+        esp!(unsafe { esp_wifi_set_mode(wifi_mode_t_WIFI_MODE_STA) })?;
+
+        if !self.is_started()? {
+            self.start()?;
```
torkleyy, line 1084: Should we start the wifi implicitly? AFAIK we don't do that for other methods like `connect`.

src/wifi.rs
```diff
@@ -1050,6 +1073,41 @@ impl<'d> WifiDriver<'d> {
         esp!(unsafe { esp_wifi_set_mac(interface.into(), mac.as_ptr() as *mut _) })
     }
 
+    /// Enable and start WPS
+    pub fn start_wps(&mut self, config: &WpsConfig) -> Result<(), EspError> {
+        let config = Newtype::<esp_wps_config_t>::try_from(config)?;
+
+        self.set_configuration(&Configuration::None)?;
+        esp!(unsafe { esp_wifi_set_mode(wifi_mode_t_WIFI_MODE_STA) })?;
```
torkleyy, line 1081: This doesn't seem right, that would cancel an active AP when initiating WPS.

src/wifi.rs
```diff
@@ -1050,6 +1073,41 @@ impl<'d> WifiDriver<'d> {
...
+        Ok(())
+    }
+
+    pub fn is_wps_active(&self) -> Result<bool, EspError> {
+        let status = self.status.lock();
+        Ok(matches!(status.2, Some(WpsEvent::Active)))
+    }
+
+    /// Gets the WPS status as a [`WPS Event`] and disables WPS.
+    pub fn take_wps_event(&mut self) -> Result<WpsEvent, EspError> {
```
torkleyy, line 1101: Any value in this being public?

src/wifi.rs
```diff
@@ -1678,10 +1779,10 @@ pub enum WifiEvent {
     StaAuthmodeChanged,
     StaBssRssiLow,
     StaBeaconTimeout,
-    StaWpsSuccess,
+    StaWpsSuccess(Option<Arc<heapless::Vec<WpsCredentials, 3>>>),
```
torkleyy, line 1782: can we use `MAX_WPS_AP_CRED` here?

src/wifi.rs
```diff
@@ -1724,13 +1825,37 @@ impl EspTypedEventDeserializer<WifiEvent> for WifiEvent {
...
         } else if event_id == wifi_event_t_WIFI_EVENT_STA_WPS_ER_SUCCESS {
-            WifiEvent::StaWpsSuccess
+            let payload =
+                unsafe { (data.payload as *const wifi_event_sta_wps_er_success_t).as_ref() };
+            let credentials = payload.map(|payload| {
+                let mut vec = heapless::Vec::new();
+                for i in 0..payload.ap_cred_cnt {
+                    let creds = &payload.ap_cred[i as usize];
+                    let Ok(ssid) = std::str::from_utf8(&creds.ssid) else {
+                        continue;
```
torkleyy, line 1835: Should probably print a warning.

src/wifi.rs
```diff
@@ -1724,13 +1825,37 @@ impl EspTypedEventDeserializer<WifiEvent> for WifiEvent {
...
+            WifiEvent::StaWpsSuccess(credentials.map(Arc::new))
         } else if event_id == wifi_event_t_WIFI_EVENT_STA_WPS_ER_FAILED {
             WifiEvent::StaWpsFailed
         } else if event_id == wifi_event_t_WIFI_EVENT_STA_WPS_ER_TIMEOUT {
             WifiEvent::StaWpsTimeout
         } else if event_id == wifi_event_t_WIFI_EVENT_STA_WPS_ER_PIN {
-            WifiEvent::StaWpsPin
+            let payload = unsafe { (data.payload as *const wifi_event_sta_wps_er_pin_t).as_ref() };
+            let pin = payload.unwrap().pin_code;
+            let pin_str = std::str::from_utf8(&pin).unwrap_or("bad pin");
```
torkleyy, line 1856: This shouldn't panic

src/private/cstr.rs
```diff
@@ -19,6 +19,16 @@ pub fn set_str(buf: &mut [u8], s: &str) -> Result<(), crate::sys::EspError> {
     Ok(())
 }
 
+pub fn set_cchar_slice(buf: &mut [c_char], s: &str) -> Result<(), EspError> {
```
ivmarkov, line 22: I don't think this is correct w.r.t. ASCIIZ strings. You are not procesing the `\0` byte correctly.

Also, rather than having yet another function that sets a `core::str` as an ASCIIZ C string, can you just use `set_str` from above? If necessary, please make it faillble rather than asserting. 

Also and if necessary, please make it working without requiring the `alloc` module.

CHANGELOG.md
```diff
@@ -6,7 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [?.??.?] - ????-??-??
+* #316 - BREAKING CHANGE addressing a typo - `http::server::Configuration::max_resp_handlers` renamed to `http::server::Configuration::max_resp_headers`
```
ivmarkov, line 9: Nit: Seems you merged/rebased wrongly? Most of what pops up as "new" here is really my earlier changes?

src/wifi.rs
```diff
@@ -1050,6 +1086,42 @@ impl<'d> WifiDriver<'d> {
         esp!(unsafe { esp_wifi_set_mac(interface.into(), mac.as_ptr() as *mut _) })
     }
 
+    /// Enable and start WPS
+    pub fn start_wps(&mut self, config: &WpsConfig) -> Result<(), EspError> {
```
ivmarkov, line 1090: `start_wps` / `take_wps_event` is confusing. Why not:
- `start_wps` - calls `esp_wifi_wps_enable` and then `esp_wifi_wps_start` (and `None`-s the `WpsStatus instance kept in the driver)
- `stop_wps` - calls `esp_wifi_wps_disable`
- `get_wps_status(...) -> WpsStatus` (WpsStatus is your renamed WpsEvent) - returns the current WPS status (including the received WPS credentials, if any)

src/wifi.rs
```diff
@@ -1050,6 +1086,42 @@ impl<'d> WifiDriver<'d> {
...
+        let config = Newtype::<esp_wps_config_t>::try_from(config)?;
+
+        match self.get_configuration()? {
```
ivmarkov, line 1093: I don't think we should be doing this on behalf of the user. If the chip is not in STA or APSTA mode, the code should just fail.

src/wifi.rs
```diff
@@ -1050,6 +1086,42 @@ impl<'d> WifiDriver<'d> {
...
+
+        esp!(unsafe { esp_wifi_wps_enable(&config.0 as *const _) })?;
+        esp!(unsafe { esp_wifi_wps_start(0) })?;
+
+        self.status.lock().wps = Some(WpsEvent::Active);
+
+        Ok(())
+    }
+
+    pub fn is_wps_active(&self) -> Result<bool, EspError> {
```
ivmarkov, line 1109: Remove

src/wifi.rs
```diff
@@ -1050,6 +1086,42 @@ impl<'d> WifiDriver<'d> {
...
+        let status = self.status.lock();
+        Ok(matches!(status.wps, Some(WpsEvent::Active)))
+    }
+
+    /// Gets the WPS status as a [`WPS Event`] and disables WPS.
+    fn take_wps_event(&mut self) -> Result<WpsEvent, EspError> {
```
ivmarkov, line 1115: See above.

src/wifi.rs
ivmarkov: Optionally, we can also merge `get_wps_status` with `stop_wps`

ivmarkov (changes requested):
Overall, looks good. Some small polishing suggested in terms of API as well as internal Cstr processing.

CHANGELOG.md
thetek42: yup, i somehow messed up with rebasing somewhere. dunno why git still considers this as an addition though.

src/wifi.rs
```diff
@@ -2244,3 +2422,103 @@ where
...
+                result
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum WpsEvent {
+    Active,
+    Success(Option<Arc<heapless::Vec<WpsCredentials, MAX_WPS_AP_CRED_USIZE>>>),
```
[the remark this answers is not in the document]
thetek42: Without the `Arc`, clippy will obviously complain because the enum variant will be huge compared to the other ones:

```
warning: large size difference between variants
    --> src\wifi.rs:1777:1
     |
1777 | / pub enum WifiEvent {
1778 | |     Ready,
1779 | |
1780 | |     ScanDone,
...    |
1789 | |     StaWpsSuccess(heapless::Vec<WpsCredentials, MAX_WPS_AP_CRED_USIZE>),
     | |     ------------------------------------------------------------------- the largest variant contains at least 316 bytes
...    |
1792 | |     StaWpsPin(Option<u32>),
     | |     ---------------------- the second-largest variant contains at least 8 bytes
...    |
1803 | |     RocDone,
1804 | | }
     | |_^ the entire enum is at least 320 bytes
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant
     = note: `#[warn(clippy::large_enum_variant)]` on by default
help: consider boxing the large fields to reduce the total size of the enum
     |
1789 |     StaWpsSuccess(Box<heapless::Vec<WpsCredentials, MAX_WPS_AP_CRED_USIZE>>),
     |                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The reason I chose `Arc` over `Box` is because it avoids copying the entire vector when the `WifiEvent` is converted to the `WpsStatus`.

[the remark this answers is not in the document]
thetek42: Also, I decided to split the `WpsStatus::Success` variant up into `SuccessConnected` and `SuccessMultipleAccessPoints(Vec<...>)` since having an empty vector for "connected successfully" would make the API kind of awkward to use.

src/wifi.rs
thetek42: The only problem I see here is that the user will need to call `.set_configuration()` with a ClientConfig in order to be able to start WPS. However, setting a ClientConfig requieres entering WiFi credentials, which are not known at this point. Thus, the user would need to enter some dummy values for SSID and password.

src/wifi.rs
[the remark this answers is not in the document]
ivmarkov: You can always do `#[allow(clippy::large_enum_variant)]`.

With that said, and since you are anyway using allocations, I don't see the point of using `heapless::Vec` in the first place. Just use a regular `alloc::vec::Vec` then.

This would also make `Vec::clone` cheaper, of course.

src/wifi.rs
ivmarkov: Hmmmmm. OK, leave it like this for now.

src/private/cstr.rs
ivmarkov: I don't think this is resolved. How is this whole function you introduced any different than `set_str`? Please fix `set_str` instead if it needs fixing (as in, make it not assert but return an error instead).

ivmarkov: Sorry, I did not realize that you are now calling `set_str`. Still, why is your function necessary? If it is just to re-cast `&mut [u8]` to `&mut [c_char]` can you please do it in the calling code instead?

ivmarkov:
@thetek42 Once you remove the pesky `set_cchar_slice` the PR should be good to go. Rest is OK.

PR #334, opened 2023-12-21
Add missing mutability qualifiers for nvs

torkleyy:
I think this is an oversight, `set_str` and `set_blob` require `&mut self`. OTOH, it seems the nvs functions perform internal locking, indicating that they might be thread safe and none of the functions require mutability.

### DIFF
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -581,7 +581,7 @@ impl<T: NvsPartitionId> EspNvs<T> {
         }
     }
 
-    pub fn set_u8(&self, name: &str, val: u8) -> Result<(), EspError> {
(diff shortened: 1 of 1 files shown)


Vollbrecht:
you say that they don't require it but now want to enforce it? Can you elaborate a bit more?

torkleyy:
@Vollbrecht haha yeah I'm not sure which one is correct, I think one could argue to make them mutable even though it would be safe without (to avoid data races, because there are multiple individually locked calls).

torkleyy:
If that's the case I will revert my change and just add some docs explaining the rationale. It just seemed weird at first sight.

PR #335, opened 2023-12-21
Issue 333 panic when scanning wifi with unknown auth method

empirephoenix:
related https://github.com/esp-rs/embedded-svc/pull/65

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -196,16 +196,17 @@ impl TryFrom<&ClientConfiguration> for Newtype<wifi_sta_config_t> {
 
 impl From<Newtype<wifi_sta_config_t>> for ClientConfiguration {
     fn from(conf: Newtype<wifi_sta_config_t>) -> Self {
-        let auth_method: Option<AuthMethod> = Newtype(conf.0.threshold.authmode).into();
-
         Self {
-            ssid: from_cstr(&conf.0.ssid).into(),
+            ssid: from_cstr(&conf.0.ssid).try_into().unwrap(),
             bssid: if conf.0.bssid_set {
                 Some(conf.0.bssid)
             } else {
                 None
             },
-            auth_method: auth_method.unwrap(),
+            auth_method: <Newtype<u32> as std::convert::Into<Option<AuthMethod>>>::into(Newtype(
+                conf.0.threshold.authmode,
+            ))
+            .unwrap(),
             password: from_cstr(&conf.0.password).try_into().unwrap(),
             channel: if conf.0.channel != 0 {
                 Some(conf.0.channel)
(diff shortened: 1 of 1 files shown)


src/wifi.rs
```diff
@@ -194,22 +194,26 @@ impl TryFrom<&ClientConfiguration> for Newtype<wifi_sta_config_t> {
     }
 }
 
-impl From<Newtype<wifi_sta_config_t>> for ClientConfiguration {
+impl From<Newtype<wifi_sta_config_t>> for Option<ClientConfiguration> {
```
ivmarkov, line 197: Why don't you keep the original code as-is, and only change
```rust
auth_method: Newtype(conf.0.threshold.authmode).into()
```

to
```rust
auth_method: Newtype(conf.0.threshold.authmode).into().unwrap()
```

?

src/wifi.rs
```diff
@@ -237,25 +241,32 @@ impl TryFrom<&AccessPointConfiguration> for Newtype<wifi_ap_config_t> {
     }
 }
 
-impl From<Newtype<wifi_ap_config_t>> for AccessPointConfiguration {
+impl From<Newtype<wifi_ap_config_t>> for Option<AccessPointConfiguration> {
```
ivmarkov, line 241: Ditto.

src/wifi.rs
```diff
@@ -1127,8 +1138,12 @@ impl<'d> WifiDriver<'d> {
         let mut wifi_config: wifi_config_t = Default::default();
         esp!(unsafe { esp_wifi_get_config(wifi_interface_t_WIFI_IF_STA, &mut wifi_config) })?;
 
-        let result: ClientConfiguration = unsafe { Newtype(wifi_config.sta).into() };
-
+        let client_config: Option<ClientConfiguration>;
```
ivmarkov, line 1133: This would then stay as-is.

src/wifi.rs
```diff
@@ -1158,8 +1173,12 @@ impl<'d> WifiDriver<'d> {
...
         let mut wifi_config: wifi_config_t = Default::default();
         esp!(unsafe { esp_wifi_get_config(wifi_interface_t_WIFI_IF_AP, &mut wifi_config) })?;
-
-        let result: AccessPointConfiguration = unsafe { Newtype(wifi_config.ap).into() };
+        let access_point_configuration: Option<AccessPointConfiguration>;
+        unsafe {
+            access_point_configuration = Newtype(wifi_config.ap).into();
+        }
+        //unwrap should be safe here, as it should be impossible to configure a accesspoint, that cannot be represented
+        let result: AccessPointConfiguration = unsafe { access_point_configuration.unwrap() };
```
ivmarkov: Why the `unsafe` here? There is nothing `unsafe` in unwrapping an `Option`. It might panic, but that's all.

ivmarkov: Anyway, here also this change would be unnecessary, once you restore the original `From` definition for `AccessPointConfiguration`.

ivmarkov (changes requested):

src/wifi.rs
empirephoenix: Hm actually, yeah good point..., let me adjust stuff

ivmarkov:
Thanks!
CI fails ATM due to unrelated issues in Rust STD. hopefully to be addressed beginning of next week.

PR #336, opened 2023-12-29
Fix EspWebSocketClient issues and add websocket_client example

mykmelez:
While developing an application that uses the EspWebSocketClient API, I ran into several bugs in the implementation. This branch includes fixes for them along with one minor enhancement and a websocket_client example that demonstrates how to use the API.

I've separated the changes into individual commits, except for the last one, which contains several fixes to enable and document building the websocket_client example with ESP-IDF 5.*. See the individual commit messages for detailed descriptions of each fix/enhancement. (If it'd be easier to review, I can also separate these changes into individual PRs.)

### DIFF
--- a/src/ws/client.rs
+++ b/src/ws/client.rs
@@ -294,8 +294,8 @@ impl<'a> TryFrom<&'a EspWebSocketClientConfig<'a>> for (esp_websocket_client_con
 /// X509 takes a CStr, which must contain a null terminator, and X509.data() returns a &[u8]
 /// including that null.
 ///
-/// But RawCstrs.as_nptr() takes an Into<Vec<u8>> and gives it to CString::new(), whose argument
-/// cannot contain a null character.
+/// But RawCstrs.as_nptr() takes an Into<alloc::vec::Vec<u8>> and gives it to CString::new(),
+/// whose argument cannot contain a null character.
 ///
 /// So this function returns the subset of the slice without the terminator.
 fn x509_data<'a>(x509: &'a X509) -> &'a [u8] {
--- a/src/ws/client.rs
+++ b/src/ws/client.rs
@@ -18,6 +18,7 @@ use crate::io::EspIOError;
 use crate::private::common::Newtype;
 use crate::private::cstr::RawCstrs;
 use crate::private::mutex::{Condvar, Mutex};
+use crate::tls::{X509Encoding, X509};
 
 #[derive(Copy, Clone, Eq, PartialEq, Debug)]
 pub enum EspWebSocketTransport {
@@ -173,32 +174,6 @@ impl<'a> WebSocketEventType<'a> {
     }
 }
 
-/// An encoding of an X.509 certificate or key.
-pub enum EspWebSocketX509Encoding<'a> {
-    PEM(&'a str),
-    DER(&'a str),
-}
-
-impl EspWebSocketX509Encoding<'_> {
-    fn len(&self) -> usize {
-        match self {
-            // Special case: when specified in esp_websocket_client_config_t, the PEM-encoded
-            // length is expected to be `0`, regardless of the actual length of the value.
-            // If the actual length is specified, then the underlying implementation will treat
-            // the value as being DER-encoded rather than PEM-encoded.
-            EspWebSocketX509Encoding::PEM(_) => 0,
-            EspWebSocketX509Encoding::DER(val) => val.len(),
-        }
-    }
-
-    fn val(&self) -> &str {
-        match self {
-            EspWebSocketX509Encoding::PEM(val) => val,
-            EspWebSocketX509Encoding::DER(val) => val,
-        }
-    }
-}
-
 #[derive(Default)]
 pub struct EspWebSocketClientConfig<'a> {
     pub username: Option<&'a str>,
@@ -224,9 +199,9 @@ pub struct EspWebSocketClientConfig<'a> {
     pub ping_interval_sec: time::Duration,
     #[cfg(esp_idf_version = "4.4")]
     pub if_name: Option<&'a str>,
-    pub server_cert: Option<EspWebSocketX509Encoding<'a>>,
-    pub client_cert: Option<EspWebSocketX509Encoding<'a>>,
-    pub client_key: Option<EspWebSocketX509Encoding<'a>>,
+    pub server_cert: Option<X509<'a>>,
+    pub client_cert: Option<X509<'a>>,
+    pub client_key: Option<X509<'a>>,
 }
 
 impl<'a> TryFrom<&'a EspWebSocketClientConfig<'a>> for (esp_websocket_client_config_t, RawCstrs) {
@@ -260,12 +235,12 @@ impl<'a> TryFrom<&'a EspWebSocketClientConfig<'a>> for (esp_websocket_client_con
 
             ping_interval_sec: conf.ping_interval_sec.as_secs() as _,
 
-            cert_pem: cstrs.as_nptr(conf.server_cert.as_ref().map(|c| c.val()))?,
-            cert_len: conf.server_cert.as_ref().map(|c| c.len()).unwrap_or(0) as _,
-            client_cert: cstrs.as_nptr(conf.client_cert.as_ref().map(|c| c.val()))?,
-            client_cert_len: conf.client_cert.as_ref().map(|c| c.len()).unwrap_or(0) as _,
-            client_key: cstrs.as_nptr(conf.client_key.as_ref().map(|c| c.val()))?,
-            client_key_len: conf.client_key.as_ref().map(|c| c.len()).unwrap_or(0) as _,
+            cert_pem: cstrs.as_nptr(conf.server_cert.as_ref().map(x509_data))?,
+            cert_len: conf.server_cert.as_ref().map(x509_len).unwrap_or(0) as _,
+            client_cert: cstrs.as_nptr(conf.client_cert.as_ref().map(x509_data))?,
+            client_cert_len: conf.client_cert.as_ref().map(x509_len).unwrap_or(0) as _,
+            client_key: cstrs.as_nptr(conf.client_key.as_ref().map(x509_data))?,
+            client_key_len: conf.client_key.as_ref().map(x509_len).unwrap_or(0) as _,
 
             // NOTE: default keep_alive_* values are set below, so they are not explicitly listed
             // here
@@ -311,6 +286,37 @@ impl<'a> TryFrom<&'a EspWebSocketClientConfig<'a>> for (esp_websocket_client_con
         }
 
         Ok((c_conf, cstrs))
+    }
+}
+
+/// Return the X509 data without the null terminator.
+///
+/// X509 takes a CStr, which must contain a null terminator, and X509.data() returns a &[u8]
+/// including that null.
+///
+/// But RawCstrs.as_nptr() takes an Into<Vec<u8>> and gives it to CString::new(), whose argument
+/// cannot contain a null character.
+///
+/// So this function returns the subset of the slice without the terminator.
+fn x509_data<'a>(x509: &'a X509) -> &'a [u8] {
+    &x509.data()[..x509.data().len().saturating_sub(1)]
+}
+
+/// Return the X509 data len to specify in esp_websocket_client_config_t fields for the given
+/// X509 encoding.
+///
+/// If the X509 data is PEM-encoded, then its length must be specified as `0`
+/// in the esp_websocket_client_config_t fields cert_len, client_cert_len, and client_key_len.
+///
+/// If the X509 data is DER-encoded, then its length should be specified as the actual length
+/// of the data (not including the null terminator).
+///
+/// So this function returns the appropriate length for the given encoding.
+fn x509_len(x509: &X509) -> usize {
+    if x509.encoding() == X509Encoding::PEM {
+        0
+    } else {
+        x509.data().len() - 1
     }
 }
 
--- a/src/ws/client.rs
+++ b/src/ws/client.rs
@@ -170,6 +170,32 @@ impl<'a> WebSocketEventType<'a> {
     }
 }
 
+/// An encoding of an X.509 certificate or key.
+pub enum EspWebSocketX509Encoding<'a> {
+    PEM(&'a str),
+    DER(&'a str),
+}
+
+impl EspWebSocketX509Encoding<'_> {
+    fn len(&self) -> usize {
+        match self {
+            // Special case: when specified in esp_websocket_client_config_t, the PEM-encoded
+            // length is expected to be `0`, regardless of the actual length of the value.
+            // If the actual length is specified, then the underlying implementation will treat
+            // the value as being DER-encoded rather than PEM-encoded.
+            EspWebSocketX509Encoding::PEM(_) => 0,
+            EspWebSocketX509Encoding::DER(val) => val.len(),
+        }
+    }
+
+    fn val(&self) -> &str {
+        match self {
+            EspWebSocketX509Encoding::PEM(val) => val,
+            EspWebSocketX509Encoding::DER(val) => val,
+        }
+    }
+}
+
 #[derive(Default)]
 pub struct EspWebSocketClientConfig<'a> {
     pub username: Option<&'a str>,
@@ -195,9 +221,9 @@ pub struct EspWebSocketClientConfig<'a> {
     pub ping_interval_sec: time::Duration,
     #[cfg(esp_idf_version = "4.4")]
     pub if_name: Option<&'a str>,
-    pub cert_pem: Option<&'a str>,
-    pub client_cert: Option<&'a str>,
-    pub client_key: Option<&'a str>,
+    pub server_cert: Option<EspWebSocketX509Encoding<'a>>,
+    pub client_cert: Option<EspWebSocketX509Encoding<'a>>,
+    pub client_key: Option<EspWebSocketX509Encoding<'a>>,
 }
 
 impl<'a> TryFrom<&'a EspWebSocketClientConfig<'a>> for (esp_websocket_client_config_t, RawCstrs) {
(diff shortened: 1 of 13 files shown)


ivmarkov:
@mykmelez Thanks for the PR and sorry for the delay!
I think overall it looks good - the one thing I need to look at tmr, is the `EspWebSocketX509Encoding` thing. I vaguely remember we already had something similar in the crate, need to look if we can unite the two.

src/ws/client.rs
```diff
@@ -165,11 +166,39 @@ impl<'a> WebSocketEventType<'a> {
...
             esp_websocket_event_id_t_WEBSOCKET_EVENT_CLOSED => Ok(Self::Closed),
+            #[cfg(esp_idf_version_major = "5")]
+            esp_websocket_event_id_t_WEBSOCKET_EVENT_BEFORE_CONNECT => Ok(Self::BeforeConnect),
             _ => Err(EspError::from_infallible::<ESP_ERR_INVALID_ARG>().into()),
         }
     }
 }
 
+/// An encoding of an X.509 certificate or key.
+pub enum EspWebSocketX509Encoding<'a> {
```
ivmarkov, line 177: Look [here](https://github.com/esp-rs/esp-idf-svc/blob/master/src/tls.rs#L72). How about reusing it, rather than coming up with a similar struct yet slightly different?

ivmarkov (changes requested):

src/ws/client.rs
mykmelez: > Look [here](https://github.com/esp-rs/esp-idf-svc/blob/master/src/tls.rs#L72). How about reusing it, rather than coming up with a similar struct yet slightly different?

Thanks for the reference! I'll try using that type.

mykmelez: Ok, I've reused the X509 type in 59926e3.

There were some complications.

First I had to extend X509 to track and report its encoding, since EspWebSocketClient must know the encoding to determine what to specify in the esp_websocket_client_config_t fields cert_len, client_cert_len, and client_key_len.

Then I had to generalize _RawCstrs.as_nptr()/as_ptr()_ and _to_cstring_arg()_ to accept an `Into<Vec<u8>>`, since _X509.data()_ returns a `&[u8]`; which also required me to update some _to_cstring_arg()_ callers that were passing `&&str` to pass `&str` instead.

And because _X509.pem()_ takes a CStr, I had to add a null terminator to the SERVER_CERT in the example app.

But because _X509.data()_ includes that null terminator, and _to_cstring_arg()_ creates a CString, whose input cannot contain a null terminator, I had to remove the terminator from the byte slice before passing it to _RawCstrs.as_nptr()_.

src/ws/client.rs
```diff
@@ -285,6 +289,37 @@ impl<'a> TryFrom<&'a EspWebSocketClientConfig<'a>> for (esp_websocket_client_con
...
+///
+/// So this function returns the subset of the slice without the terminator.
+fn x509_data<'a>(x509: &'a X509) -> &'a [u8] {
+    &x509.data()[..x509.data().len().saturating_sub(1)]
+}
+
+/// Return the X509 data len to specify in esp_websocket_client_config_t fields for the given
+/// X509 encoding.
+///
+/// If the X509 data is PEM-encoded, then its length must be specified as `0`
```
ivmarkov, line 308: Are you sure this is the case?
The way I read the documentation in the client [here](https://github.com/espressif/esp-idf/blob/release/v4.4/components/esp_websocket_client/include/esp_websocket_client.h#L86), it says "**may** be 0 for for null-terminated pem".

... so it is not mandatory to pass 0. You can also pass the full length, including the "0".

Why am I saying this? The `X509` struct is used already in the HTTP client [here](https://github.com/esp-rs/esp-idf-svc/blob/master/src/http/client.rs#L84), HTTP server [here](https://github.com/esp-rs/esp-idf-svc/blob/master/src/http/server.rs#L90), and the MQTT client [here](https://github.com/esp-rs/esp-idf-svc/blob/master/src/mqtt/client.rs#L80).

The key is that all of these components share the same TLS layer, and I would assume the same layer is used by the WS client.

In **none** of the other cases, we had to pass 0 for an ASCIIZ pem certificate. We just unconditionally pass `as_esp_idf_raw_len` and live happily thereafter. I would suggest to just copy-paste the code from one of the other places and check if that works.

src/ws/client.rs
```diff
@@ -231,12 +235,12 @@ impl<'a> TryFrom<&'a EspWebSocketClientConfig<'a>> for (esp_websocket_client_con
 
             ping_interval_sec: conf.ping_interval_sec.as_secs() as _,
 
-            cert_pem: cstrs.as_nptr(conf.cert_pem)?,
-            cert_len: conf.cert_pem.map(|c| c.len()).unwrap_or(0) as _,
-            client_cert: cstrs.as_nptr(conf.client_cert)?,
-            client_cert_len: conf.client_cert.map(|c| c.len()).unwrap_or(0) as _,
-            client_key: cstrs.as_nptr(conf.client_key)?,
-            client_key_len: conf.client_key.map(|c| c.len()).unwrap_or(0) as _,
+            cert_pem: cstrs.as_nptr(conf.server_cert.as_ref().map(x509_data))?,
```
ivmarkov, line 238: This is doing an expensive (in terms of RAM usage) copy of the der/pem data which is referenced by the X509 certificate struct into c-strings that are kept internally in the `RawCstrs` struct which is kept around while the client is created, but is then **dropped** by the current code, which is wrong.

I don't think dropping the `RawCstrs` struct is save, as the TLS layer under the WS client *does not do its own copy of the PEM/DER data*. So you have to keep around the `RawCstrs` structure until the client is alive, as its member.
 
I think however it is much better to take another approach: change the types from `X509<'a>` to `X509<'static>` as we do everywhere else. This would enforce the user to pass statically allocated pems and ders, but will remove the RAM consumption and the copying.

And... pems and ders tend to be inbcluded into the flash binary's `.rodata` section using `include_bytes!` / `include_str!` anyway (as you did in your `main` function) so why copy them?

ivmarkov (changes requested):
Sorry for the late reply.

I think most of the extensions to string handling and X509 are not needed and removing those would simplify the code considerable. Please check my comments and if you can try out my suggestions, that would be appreciated!

ivmarkov:
@mykmelez I would like to branch a new release of the `esp-idf-*` crates by end of this week, however I would really like to also pull your WS-related fixes and the WS client example too. If you won't have time to work on that in the next couple of days, I can address the remaining X509-related concerns? Just give me a sign.

src/ws/client.rs
ivmarkov: I think you were fooled [by this code line](https://github.com/espressif/esp-idf/blob/release/v4.4/components/esp_websocket_client/esp_websocket_client.c#L368)...

... but then when digging further, you'll eventually notice two things:
* Most of the TLS code only uses the PEM buffer that is set e.g. by the above call and disregards the DER one, which looks very weird...
* ... until you see [this](https://github.com/espressif/esp-idf/blob/release/v4.4/components/esp-tls/esp_tls.h#L117) which explains it all, and which - I hope - proves that *we do not need to differentiate between DER and PEM*.
:)

ivmarkov:
@mykmelez See [this](https://github.com/esp-rs/esp-idf-svc/pull/355). I've applied here all fixes from this PR. Let me run the example to double check it works fine, and if all is OK, I'll push.

BTW this new PR is based on your branch, so you would still be listed as a contributor.

ivmarkov:
@mykmelez Good news! `X509` works as-is for the WS client as well.
I therefore merged the other PR.

ivmarkov:
I think we can close this, thanks again.

src/ws/client.rs
mykmelez: > I think you were fooled [by this code line](https://github.com/espressif/esp-idf/blob/release/v4.4/components/esp_websocket_client/esp_websocket_client.c#L368)...

I think I was rather fooled by my expectation that a high-level Rust API like EspWebSocketClient that takes a `&str` certificate wouldn't require its value to include a C-style null terminator.

As well, because I observed this error when setting EspWebSocketClientConfig.cert_pem to that value:

> E (11701) esp-tls-mbedtls: mbedtls_x509_crt_parse of CA cert returned -0x2180
> E (11701) esp-tls-mbedtls: Failed to set client configurations, returned [0x8015] (ESP_ERR_MBEDTLS_X509_CRT_PARSE_FAILED)
> E (11711) esp-tls: create_ssl_handle failed
> E (11711) esp-tls: Failed to open new connection
> E (11721) transport_base: Failed to open a new connection
> E (11731) transport_ws: Error connecting to host example.websocket.org:443
> E (11741) websocket_client: esp_transport_connect() failed with -1, transport_error=ESP_ERR_MBEDTLS_X509_CRT_PARSE_FAILED, tls_error_code=8576, tls_flags=0, esp_ws_handshake_status_code=0, errno=119

And I found that the error is resolved by setting esp_websocket_client_config_t.cert_len to `0` when the cert is PEM-encoded, which causes _esp_websocket_client_create_transport()_ to call _esp_transport_ssl_set_cert_data()_ rather than _esp_transport_ssl_set_cert_data_der()_ (as you've noted):

```c
            if (!client->config->cert_len) {
                esp_transport_ssl_set_cert_data(ssl, client->config->cert, strlen(client->config->cert));
            } else {
                esp_transport_ssl_set_cert_data_der(ssl, client->config->cert, client->config->cert_len);
            }
```

(https://github.com/junelife/esp-protocols/blob/efa26b7062623d6eaa33b09da737d4c87f3e20ea/components/esp_websocket_client/esp_websocket_client.c#L503-L507).

> ... but then when digging further, you'll eventually notice two things:
> 
>     * Most of the TLS code only uses the PEM buffer that is set e.g. by the above call and disregards the DER one, which looks very weird...
> 
>     * ... until you see [this](https://github.com/espressif/esp-idf/blob/release/v4.4/components/esp-tls/esp_tls.h#L117) which explains it all, and which - I hope - proves that _we do not need to differentiate between DER and PEM_.

Yes, _esp_transport_ssl_set_cert_data()_ and _esp_transport_ssl_set_cert_data_der()_ both write the cert to the same buffer, since cacert_pem_buf is an alias for cacert_buf. And they both write the cert length to the same field, since cacert_pem_bytes is an alias for cacert_bytes. But they write a **different length** to that field:

```c
void esp_transport_ssl_set_cert_data(esp_transport_handle_t t, const char *data, int len)
{
    GET_SSL_FROM_TRANSPORT_OR_RETURN(ssl, t);
    ssl->cfg.cacert_pem_buf = (void *)data;
    ssl->cfg.cacert_pem_bytes = len + 1;
}

void esp_transport_ssl_set_cert_data_der(esp_transport_handle_t t, const char *data, int len)
{
    GET_SSL_FROM_TRANSPORT_OR_RETURN(ssl, t);
    ssl->cfg.cacert_buf = (void *)data;
    ssl->cfg.cacert_bytes = len;
}
```

(https://github.com/espressif/esp-idf/blob/release/v5.1/components/tcp_transport/transport_ssl.c#L352-L364)

Because _esp_transport_ssl_set_cert_data()_ adds `1` to the length, _mbedtls_x509_crt_parse()_ sees its null terminator when searching the buffer, which is one of the conditions for it to treat the buffer as PEM-encoded:

```c
    if (buflen != 0 && buf[buflen - 1] == '\0' &&
        strstr((const char *) buf, "-----BEGIN CERTIFICATE-----") != NULL) {
        buf_format = MBEDTLS_X509_FORMAT_PEM;
    }
```

(https://github.com/Mbed-TLS/mbedtls/blob/47c74a477378ec3f0d1ba80547db836e078fa3a0/library/x509_crt.c#L1412-L1415)

The current implementation works, but I don't think it uses the underlying APIs as they were intended to be used. And personally I'd prefer a Rust API that takes a `&str` certificate not to require it to contain an explicit C-style null terminator. But it's your project, your decision.

src/ws/client.rs
mykmelez: The requirement for an explicit null terminator actually makes it challenging to use _include_bytes!()_ and _include_str!()_, as @torkleyy noted when inlining a certificate into the blocking TLS example in 26057471e4698f4ee4162ccfbeb73d3a67a169b7. (I should've added a similar comment to the websocket client example.)

But it isn't impossible, as noted in https://stackoverflow.com/questions/75064278/include-str-for-null-terminated-string. And we can at the very least inline the certs into the source and avoid copying them (a worthy goal, to be sure), as we've done here, by explicitly adding a null terminator.

src/ws/client.rs
ivmarkov: > The current implementation works, but I don't think it uses the underlying APIs as they were intended to be used. 

Can you clarify? As mentioned above, I don't think the API actually says or mandates that you should pass 0 as length for the DER certificate. The word "**may**" being key - either you pass 0, or if you pass the actual length, it *should* contain the 0. When you pas "0", what happens is that the API simply converts the length to the *other* accepted format, which is `asciiz_length_of_the_string` + 1

> And personally I'd prefer a Rust API that takes a `&str` certificate not to require it to contain an explicit C-style null terminator. But it's your project, your decision.

My understanding is that Rust `&str` is completely agnostic w.r.t. `\0`. Meaning, it can contain multiple `\0`s inside it, and everything should work regardless. It might be a bit surprising as we are used to ASCIIZ, but I think that's true. Please correct me if I'm wrong.

> But it's your project, your decision.

I'm open to changing the existing API, but I hope you would agree that when changing we should preserve these invariants:
* Code should behave the same and use the same `X509` types *everywhere*. As in WS client, HTTP (and WS) server, MQTT etc. Even if the API is weird, as long as its consistent everywhere I think that's a win, compared to - say - "WS client has a slightly different way of doing things w.r.t. TLS"
* Could should not allocate and keep around large `CStrings` only in the name of user friendliness.

Thanks!

src/ws/client.rs
Vollbrecht: its trivial to add a null termination to a .pem file you get out of for example openssl. For example you can use `dd if=/dev/zero bs=1 count=1 >> your_cert.pem` or a more "safe" way is to use `truncate -s +1 your_cert.pem` and then just include it via the macro.

src/ws/client.rs
mykmelez: > > The current implementation works, but I don't think it uses the underlying APIs as they were intended to be used.
> 
> Can you clarify? As mentioned above, I don't think the API actually says or mandates that you should pass 0 as length for the DER certificate. The word "**may**" being key - either you pass 0, or if you pass the actual length, it _should_ contain the 0. When you pas "0", what happens is that the API simply converts the length to the _other_ accepted format, which is `asciiz_length_of_the_string` + 1

My interpretation of "Length of the buffer pointed to by cert_pem. May be 0 for null-terminated pem" is that the value of cert_len may or may not be `0`, and whether it is or not depends on the encoding of the cert in cert_pem, i.e. `0` if cert_pem is PEM-encoded and the length of cert_pem if it's DER-encoded.

It doesn't suggest to me that it was intended for you to pass the length including the null terminator for a PEM-encoded cert (whether or not it's technically possible to do so).

Nor is there any such suggestion in the implementation, since the only thing esp_websocket_client does with cert_pem is pass it into either [esp_transport_ssl_set_cert_data()](https://github.com/espressif/esp-idf/blob/a5b261f699808efdacd287adbded5b718dffd14e/components/tcp_transport/include/esp_transport_ssl.h#L25-L34) or [esp_transport_ssl_set_cert_data_der()](https://github.com/espressif/esp-idf/blob/a5b261f699808efdacd287adbded5b718dffd14e/components/tcp_transport/include/esp_transport_ssl.h#L36-L45), which say they "Set SSL certificate data (as PEM format)" and "Set SSL certificate data (as DER format)", respectively.

As for examples, the one in the esp_websocket_client component doesn't specify a cert,  but the configuration example in https://docs.espressif.com/projects/esp-idf/en/v4.4.6/esp32/api-reference/protocols/esp_websocket_client.html#tls shows a PEM-encoded cert being specified without a length.

Same with the esp_websocket_client consumers in https://github.com/bleskomat/bleskomat-bills-firmware/blob/efeca74006b11c289323f7d5553e7151eaf1440b/src/platform.cpp#L139-L144, https://github.com/abobija/esp-discord/blob/606e04fc021ceac4a6776c06c2d37889162fb83a/src/discord/private/_gateway.c#L196-L204, and https://github.com/systronix/heimdall/blob/b34c295313c58d124691e40d493ebfb8624526d7/backend/common/network.c#L262-L267 (found in a GitHub search).

Because the data ends up in the same buffer, and Mbed TLS uses a heuristic to determine the encoding in that buffer, it's possible to specify a PEM-encoded cert using a non-zero length that winds up working, despite it triggering a call into an API that expects a DER-encoded cert. So you "may" do it in the sense that it's technically possible. But I see nothing in the docs, the implementation, the underlying APIs, nor examples to suggest that it was intended to specify a PEM-encoded cert in this way.

> > And personally I'd prefer a Rust API that takes a `&str` certificate not to require it to contain an explicit C-style null terminator. But it's your project, your decision.
> 
> My understanding is that Rust `&str` is completely agnostic w.r.t. `\0`. Meaning, it can contain multiple `\0`s inside it, and everything should work regardless. It might be a bit surprising as we are used to ASCIIZ, but I think that's true. Please correct me if I'm wrong.

You're correct, and when I say that I'd personally prefer not to have to append a null terminator to my PEM-encoded certificate, I mean that it feels non-canonical and a leaky abstraction, given that the terminator is not part of the encoding of the cert, nor a requirement of the Rust language (which as you note treats it as just another character), but rather a requirement of the underlying C implementation that the Rust API is abstracting, or at least of the particular way that the Rust API abstracts it (without any copying that would enable it to add this terminator itself to meet the requirements of the underlying implementation).

> I'm open to changing the existing API, but I hope you would agree that when changing we should preserve these invariants:
> 
>     * Code should behave the same and use the same `X509` types _everywhere_. As in WS client, HTTP (and WS) server, MQTT etc. Even if the API is weird, as long as its consistent everywhere I think that's a win, compared to - say - "WS client has a slightly different way of doing things w.r.t. TLS"

Yes, absolutely, which is why I rewrote my branch to use the same X509 type once you pointed it out to me.

>     * Could should not allocate and keep around large `CStrings` only in the name of user friendliness.

I disagree and am willing to spend memory to achieve ergonomics, although I do agree that it's valuable to minimize allocations, and I understand why you want to avoid doing so here.

PR #338, opened 2023-12-30
fixes `*mut esp_netif_obj` not Sync compiler error

### LINKED ISSUE #56: esp_netif_obj isn't thread safe
I was trying to access the `wifi` object within a thread but it is throwing this error:
```
202 |     thread::Builder::new().stack_size(32768).spawn(move || {
    |                                              ^^^^^ `*mut esp_netif_obj` cannot be sent between threads safely
    |
    = help: within `EspWifi`, the trait `Send` is not implemented for `*mut esp_netif_obj`
    = note: required because it appears within the type `EspNetif`
    = note: required because it appears within the type `std::option::Option<EspNetif>`
    = note: required because it appears within the type `EspWifi`
    = note: required because of the requirements on the impl of `Sync` for `std::sync::Mutex<EspWifi>`
    = note: required because of the requirements on the impl of `Send` for `Arc<std::sync::Mutex<EspWifi>>`
    = note: required because it appears within the type `[closure@src/main.rs:202:52: 204:6]`
note: required by a bound in `Builder::spawn`
   --> /Users/username/.rustup/toolchains/esp/lib/rustlib/src/rust/library/std/src/thread/mod.rs:379:12
    |
379 |         F: Send + 'static,
    |            ^^^^ required by this bound in `Builder::spawn`

```

This happens after I wrapped wifi object inside a mutex and arc.
```
    let wifi = EspWifi::new(netif_stack, sys_loop_stack, default_nvs).expect("Need wifi");
    let wifi_arc = Arc::new(Mutex::new(wifi));

    thread::Builder::new().stack_size(32768).spawn(move || {
        let status = wifi_arc.clone().lock().unwrap().get_status();
    });

```

cs-clarence:
Fixes #56 (again).
Implements a new type wrapper `SendSyncPtr` for arbitrary pointer types and `impl`s `Send` and `Sync` on that wrapper. This is just an assertion for the compiler and assumes that the underlying apis from the sdk are relatively thread safe.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /target
 /Cargo.lock
 **/*.rs.bk
+/rust-toolchain.toml
--- a/src/eventloop.rs
+++ b/src/eventloop.rs
@@ -19,6 +19,7 @@ use crate::hal::cpu::Core;
 use crate::hal::delay::TickType;
 use crate::hal::interrupt;
 
+use crate::private::common::SendSyncPtr;
 use crate::sys::*;
 
 use crate::handle::RawHandle;
@@ -215,8 +216,8 @@ struct UnsafeCallback<'a>(*mut Box<dyn FnMut(&EspEventFetchData) + Send + 'a>);
 
 impl<'a> UnsafeCallback<'a> {
     #[allow(clippy::type_complexity)]
-    fn from(boxed: &mut Box<Box<dyn FnMut(&EspEventFetchData) + Send + 'a>>) -> Self {
-        Self(boxed.as_mut())
+    fn from(boxed: &mut Box<dyn FnMut(&EspEventFetchData) + Send + 'a>) -> Self {
+        Self(boxed)
     }
 
     unsafe fn from_ptr(ptr: *mut ffi::c_void) -> Self {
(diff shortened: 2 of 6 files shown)


ivmarkov:
> Fixes #56 (again). Implements a new type wrapper `SendSyncPtr` for arbitrary pointer types and `impl`s `Send` and `Sync` on that wrapper. This is just an assertion for the compiler and assumes that the underlying apis from the sdk are relatively thread safe.

There is no such thing as "relatively" thread safe. Either they are thread safe (i.e. you can call methods on the pointer concurrently, from any thread - so you can unsafely "proclaim" the raw C API as `Send` + `Sync`) or not. If it is not, and you still mark it as `Send` + `Sync` you'll be ending up with Rust UB = weird crashes/misbehaviors in your program, which is against the whole point of using rust in the first place.

`esp_netif_t` is **not** thread-safe, contrary to what the ESP IDF documentation says.  Did you read [this](https://github.com/esp-rs/esp-idf-svc/issues/56#issuecomment-1859070030)? It is right there, in #56.

What you are doing with this PR is a bit to blindly apply `Send` + `Sync` to a bunch of raw C APIs. This is very dangerous. Each API has to be reviewed **in detail** as to if it can be proclaimed `Send` + `Sync` (i.e. "thread safe").

The event loop *might* be thread safe. The Netif API, as per above - is just not (data races between reading the Netif state and another thread modifying the Netif state).

I'll address the `await` + `tokio::sync::Mutex` topic in the issue, a bit later.

PR #347, opened 2024-01-18
Docstrings for the httpd module

IvanSanchez:
Documents some main functionality of the `httpd` module.

Not 100% complete, but it should be enough to users like me who want to program servers.

### DIFF
--- a/src/http/server.rs
+++ b/src/http/server.rs
@@ -1,4 +1,31 @@
 //! HTTP server
+//!
+//! Provides a HTTP(S) server in `EspHttpServer`, plus all related structs.
+//!
+//! Typical usage of `EspHttpServer` involves creating a function (or closure)
+//! for every URI+method that the server is meant to handle. A minimal server that
+//! only handles HTTP GET requests to `index.html` looks like this:
+//!
+//! ```
+//! use esp_idf_svc::http::server::{Configuration, EspHttpServer};
+//!
+//! let mut server = EspHttpServer::new(&Configuration::default())?;
+//!
+//! server.fn_handler("/index.html", Method::Get, |request| {
+//! 	request.into_ok_response()?
+//! 		.write_all("<html><body>Hello world!</body></html>".as_bytes())?;
+//!
+//! 	Ok(());
+//! })?;
+//! ```
+//!
+//! Note that the server is automatically started when instantiated.
+//!
+//! You can find an example of handling GET/POST requests at [`json_post_handler.rs`](https://github.com/esp-rs/esp-idf-svc/blob/master/examples/json_post_handler.rs).
+//!
+//! You can find an example of HTTP+Websockets at [`examples/ws_guessing_game.js`](https://github.com/esp-rs/esp-idf-svc/blob/master/examples/ws_guessing_game.rs).
+
+
 use core::cell::UnsafeCell;
 use core::fmt::Debug;
 use core::marker::PhantomData;
@@ -257,6 +284,7 @@ impl EspHttpServer<'static> {
     }
 }
 
+/// HTTP server
 impl<'a> EspHttpServer<'a> {
     /// # Safety
     ///
@@ -341,6 +369,7 @@ impl<'a> EspHttpServer<'a> {
         Ok(server)
     }
 
+    /// Unregisters a URI.
     fn unregister(&mut self, uri: CString, conf: httpd_uri_t) -> Result<(), EspIOError> {
         unsafe {
             esp!(httpd_unregister_uri_handler(
@@ -361,6 +390,7 @@ impl<'a> EspHttpServer<'a> {
         Ok(())
     }
 
+    /// Stops the server.
     fn stop(&mut self) -> Result<(), EspIOError> {
         if !self.sd.is_null() {
             while let Some((uri, registration)) = self.registrations.pop() {
@@ -401,6 +431,7 @@ impl<'a> EspHttpServer<'a> {
         Ok(self)
     }
 
+    // Registers a `Handler` for a URI and a method (GET, POST, etc).
     pub fn handler<H>(
         &mut self,
         uri: &str,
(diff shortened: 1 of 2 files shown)


ivmarkov:
Run `cargo fmt`

src/http/server.rs
```diff
@@ -1,4 +1,31 @@
 //! HTTP server
+//!
+//! Provides a HTTP(S) server in `EspHttpServer`, plus all related structs.
```
ivmarkov, line 3: Nit: "an" not "a"

src/http/server.rs
```diff
@@ -1,4 +1,31 @@
...
+//! for every URI+method that the server is meant to handle. A minimal server that
+//! only handles HTTP GET requests to `index.html` looks like this:
+//!
+//! ```
+//! use esp_idf_svc::http::server::{Configuration, EspHttpServer};
+//!
+//! let mut server = EspHttpServer::new(&Configuration::default())?;
+//!
+//! server.fn_handler("/index.html", Method::Get, |request| {
+//! 	request.into_ok_response()?
```
ivmarkov, line 15: This example would no longer compile since today. You either need to annotate the exact error type you are returning using a turbofish syntax on the `Ok(())` or you should change this code to just be 
```
request.into_ok_response()?
    .write_all("<html><body>Hello world!</body></html>".as_bytes())
```

... without the final `Ok`
```

File: src/http/server.rs [hunk 3]
```diff
@@ -434,6 +465,11 @@ impl<'a> EspHttpServer<'a> {
         Ok(self)
     }
 
+    /// Registers a function as the handler for the given URI and HTTP method (GET, POST, etc).
+    ///
+    /// The function will be called every time a HTTP client requests that URI
```
ivmarkov, line 475: Also here: "an"

src/http/server.rs
```diff
@@ -498,6 +534,7 @@ impl<'a> EspHttpServer<'a> {
 }
 
 impl<'a> Drop for EspHttpServer<'a> {
+    /// Stops and drops the server.
```
ivmarkov, line 542: I'm not sure that it is a common practice to document trait impls. Perhaps rather than doing this here, maybe in the top-level comment (which is indeed very useful) you can mention that the user has to keep around the http server, because dropping it would stop it and destroy everything.

src/http/server.rs
```diff
@@ -511,6 +548,9 @@ impl<'a> RawHandle for EspHttpServer<'a> {
     }
 }
 
+/// Wraps the given function into a `FnHandler`.
```
ivmarkov, line 556: an

src/http/server.rs
```diff
@@ -787,6 +849,7 @@ impl<'a> EspHttpConnection<'a> {
         Ok(buf.len())
     }
 
+    // As per `EspHttpConnection::write`.
```
ivmarkov, line 857: ?

src/http/server.rs
```diff
@@ -1130,12 +1199,14 @@ pub mod ws {
     }
 
     impl Sender for EspHttpWsConnection {
+        /// As per [`crate::http::EspHttpWsConnection::send`](EspHttpWsConnection::send)
```
ivmarkov, line 1207: Ditto here and also for `Receiver` - not sure trait impls need documentation. It is another topic that the traits themselves need documentation, but this is for `embedded-svc` :)

src/http/server.rs
```diff
@@ -1297,6 +1368,14 @@ pub mod ws {
     }
 
     impl<'a> EspHttpServer<'a> {
+        /// Registers a function as the handler for a Websockets URI.
+        ///
+        /// The function will be called every time a Websockets connection is
+        /// made to that URI, receiving a different `EspHttpWsConnection` each
+        /// call.
+        ///
+        /// Note that Websockets functionality is gated behind a SDK flag.
```
ivmarkov, line 1382: an

ivmarkov (changes requested):
Very nice, thank you! If you can address a few nits, and then I'll merge.

src/http/server.rs
IvanSanchez: The `write` and `write_all` methods behave the same way, except for the return value. I would have expected `write_all` to flush the buffers or the like.

From my point of view, having both methods is a bit confusing.

src/http/server.rs
```diff
@@ -1,4 +1,37 @@
...
+//!
+//! server.fn_handler("/index.html", Method::Get, |request| {
+//! 	request.into_ok_response()?
+//! 		.write_all("<html><body>Hello world!</body></html>".as_bytes());
+//!
+//! 	Ok(());
+//! })?;
+//! ```
+//!
+//! Note that the server is automatically started when instantiated, and stopped
```
ivmarkov, line 22: Maybe nitpicking again, but 
* remove "Due to this, a common practice is to run the HTTP server in its own thread.". This is not a common practice, and I don't see the point of doing this. You just need to have the server instance around, that's all
* "and when its references are dropped" -> "and stopped when dropped"

... or I can do that too post-commit.

ivmarkov:
@IvanSanchez Do you plan to finish this PR? CI fails and there are still two outstanding issues.

src/http/server.rs
IvanSanchez: Changed to "If you want to keep the server running indefinitely then make sure it's not dropped - you may add an infinite loop after the server is created, use `core::mem::forget`, or keep around a reference to it somehow.".

(Doesn't show as a change since it's a couple of lines down)

src/http/server.rs
IvanSanchez: Changed to "If you want to keep the server running indefinitely then make sure it's not dropped - you may add an infinite loop after the server is created, use core::mem::forget, or keep around a reference to it somehow.".

(Doesn't show as a change since it's a couple of lines down)

src/http/server.rs
IvanSanchez: Changed. (Doesn't show as a change since it's a couple of lines down)

IvanSanchez:
Ugh, `clippy` is complaining about spaces/tabs, will fix.

src/http/server.rs
ivmarkov: Argh, this is now even worse and not what I suggested. Np, I'll fix post-commit.

ivmarkov:
Thanks!

PR #352, opened 2024-01-22
add missing wifi event WIFI_EVENT_HOME_CHANNEL_CHANGE

1HPorange:
Hi,

Somehow this event was missing for me. It seems to be [here](https://github.com/espressif/esp-idf/blob/b3f7e2c8a4d354df8ef8558ea7caddc07283a57b/components/esp_wifi/include/esp_wifi_types.h#L906), and my home WIFI setup triggers it consistently with the basic connect example, so I need this patch to get around the panic.

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -1824,6 +1824,7 @@ pub enum WifiEvent {
     FtmReport,
     ActionTxStatus,
     RocDone,
+    HomeChannelChange,
 }
 
 impl EspTypedEventSource for WifiEvent {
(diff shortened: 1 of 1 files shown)


1HPorange:
Oh, seems i need to version gate this somehow, how do you prefer that to be done?

Vollbrecht:
What version of esp-idf are you using? I can't imagine that esp is currently triggering it, as you see it was added [here](https://github.com/espressif/esp-idf/commit/40d7cd8b683657e1a7787a0611876241642c1988) relatively recently and this commit also touches all the proprietary wifi blobs, so it is not included in any released esp-idf version. 

I mean yes your router can try force a channel switch non-stop, but that should be independent of that event.

 Anyway if that lands on one of the release versions its fine to futureproof, in general we import via conditional cfg's with respect to the taged semver versions of the esp-idf. Example [here](https://github.com/esp-rs/esp-idf-svc/blob/924022ba3439d2969d7b3499e275cccdd7db64a1/src/wifi.rs#L533) if worst come to worst this way, otherwise if there is a clean point where they only have it for newer versions and dont backport it, it might just one cfg point to include it.

1HPorange:
I am using a the espressif idf dev container `espressif/idf:latest`. The reported version is `ESP-IDF v5.3-dev-1353-gb3f7e2c8a4`.

I might have gone a bit too "bleeding edge" :) After reading your thoughts it's probably best to wait before including this event, which might not even land.

PR #355, opened 2024-01-26
Ws client fixes

ivmarkov:


.github/workflows/ci.yml
```diff
@@ -51,28 +51,48 @@ jobs:
         env:
           ESP_IDF_VERSION: ${{ matrix.idf-version }}
           ESP_IDF_SDKCONFIG_DEFAULTS: "${{ github.workspace }}/.github/configs/sdkconfig.defaults"
+<<<<<<< HEAD
           RUSTFLAGS: "${{ matrix.idf-version == 'v5.1.2' && '--cfg espidf_time64' || ''}}"
+=======
+          ESP_IDF_COMPONENT_MANAGER: "${{ matrix.idf-version == 'v5.1.1' && 'y' || 'n'}}"
+          RUSTFLAGS: "${{ matrix.idf-version == 'v5.1.1' && '--cfg espidf_time64' || ''}}"
```
Vollbrecht, line 58: version 5.1.1 still in use here

.github/workflows/ci.yml
```diff
@@ -51,28 +51,48 @@ jobs:
...
         run: cargo clippy --features nightly,experimental --no-deps --target ${{ matrix.target }} -Zbuild-std=std,panic_abort -Zbuild-std-features=panic_immediate_abort -- -Dwarnings
 
       - name: Build | Compile
         env:
           ESP_IDF_VERSION: ${{ matrix.idf-version }}
           ESP_IDF_SDKCONFIG_DEFAULTS: "${{ github.workspace }}/.github/configs/sdkconfig.defaults"
+<<<<<<< HEAD
           RUSTFLAGS: "${{ matrix.idf-version == 'v5.1.2' && '--cfg espidf_time64' || ''}}"
+=======
+          ESP_IDF_COMPONENT_MANAGER: "${{ matrix.idf-version == 'v5.1.1' && 'y' || 'n'}}"
```
Vollbrecht, line 69: dito

components_esp32s3.lock
```diff
@@ -0,0 +1,15 @@
+dependencies:
```
Vollbrecht, line 1: we should add the .lock files produced by the component manager .gitignore?

Vollbrecht (changes requested):
Added comments

.github/workflows/ci.yml
ivmarkov: The code in the CI yaml will all be reverted. This example is prepared to only compile for ESP IDF 4. There are comments in the example if the user wants to build it for ESP IDF 5. But we cannot alter the `Cargo.toml` file of `esp-idf-svc` in a way where it becomes compatible with ESP IDF 5 only (by including the websocket component). The user will have to do this in their own `Cargo.toml.

components_esp32s3.lock
ivmarkov: There will be no lock files produced by `esp-idf-svc`, as per above.

ivmarkov:
> Added comments

Thanks!

PR #356, opened 2024-01-29
Change payload of EspEventPostData from [u32] to [u8]

IvanSanchez:
I'm implementing variable-sized events (via `serde`+`postcard`). The output of `postcard` is a `heapless::Vec<u8>`, which is exceedingly difficult to fit into a `[u32]`.

Since I cannot see any reason why the payload of a variable-sized event has to be alined to 4-byte words, I propose changing that to `[u8]`.

I have tested this change with https://gitlab.com/IvanSanchez/solanera/-/commit/94a974b70f47d75a6123580041fe0780a4b8e411 and everything seems to work allright.


ivmarkov:
No that's a typing mistake - thanks

PR #358, opened 2024-02-04
make cargo espflash work by adding a partition.csv

Vollbrecht:
For running the examples we are using `cargo espflash`. By default `cargo espflash` will be using the partition.bin created inside esp-idf-sys ( `espflash` by default will use its own. The problem is that the default esp-idf-sys partition has only a 1MB big AppPartition size. All sdkconfig options for using different sized partitions only lead to a max AppPartition to 1.5MB and this is still to less for an wifi example with debug infos. So the simplest / cleanest solution is to simply provide one with the configuration file-

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,10 @@ $ MCU=esp32c3 cargo espflash flash --target riscv32imc-esp-espidf --example wifi
 | esp32c3| riscv32imc-esp-espidf |
 | esp32c6| riscv32imac-esp-espidf |
 | esp32h2 | riscv32imac-esp-espidf |
+| esp32p4 | riscv32imafc-esp-espidf |
 | esp32 | xtensa-esp32-espidf |
 | esp32s2 | xtensa-esp32s2-espidf |
-| esp32s3 | xtemsa-esp32s3-espidf |
+| esp32s3 | xtensa-esp32s3-espidf |
 
 
 ## Setting up a "Hello, world!" binary crate with ESP IDF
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -24,15 +24,21 @@ Follow the [Prerequisites](https://github.com/esp-rs/esp-idf-template#prerequisi
...
-```
+| MCU | "--target" |
+| --- | ------ |
+| esp32c2 | riscv32imc-esp-espidf |
+| esp32c3| riscv32imc-esp-espidf |
+| esp32c6| riscv32imac-esp-espidf |
+| esp32h2 | riscv32imac-esp-espidf |
+| esp32 | xtensa-esp32-espidf |
+| esp32s2 | xtensa-esp32s2-espidf |
+| esp32s3 | xtemsa-esp32s3-espidf |
```
ivmarkov, line 40: Typo: `xtemsa` instead of `xtensa`.

ivmarkov: Could you provide mappings for esp32c5 (I know it's not out yet, but support of it is ready) as well as for esp32p4, which already has the `riscv32imafc-esp-espidf` target upstreamed?

ivmarkov (changes requested):
Just a small typo and if you can list two additional al mcus

ivmarkov:
I think we need this for esp-idf-hal as well, no? And for esp-idf-sys?

README.md
Vollbrecht: yeah will do

Vollbrecht:
i

> I think we need this for esp-idf-hal as well, no? And for esp-idf-sys?

yeah already pushed

README.md
Vollbrecht: though we don't know the target of c5. it could be anything at this point, but i will add p4 as it is on the horizion

ivmarkov:
Thanks!

PR #365, opened 2024-02-16
Support home channel change

MartinBroers:
Commit to support WIFI_EVENT_HOME_CHANNEL_CHANGE, introduced in https://github.com/espressif/esp-idf/commit/1427e14b724292ee8ceede6523d98cdee7c06037. I have opted to 'just' support the event and not propagate the values which come into the struct in the C library, but I could add those if required.

Also, I guarded it with a config directive (`#[cfg(all(esp_idf_version_major = "5", esp_idf_version_minor = "2",))]`, so not to break backwards builds. Maybe this is too restrictive, and we need to adjust for future builds (anything greater than 5.2), so  I am open to suggestions here. Or is the idea the next release works for 5.2 and not for older versions?

Also, I took the liberty to fix a few compilation errors and added 5.2 to the github workflow, but I am not sure this all is correct.


src/bt/a2dp.rs
```diff
@@ -134,6 +134,13 @@ pub enum ConnectionStatus {
 #[repr(u32)]
 pub enum AudioStatus {
     SuspendedByRemote = esp_a2d_audio_state_t_ESP_A2D_AUDIO_STATE_REMOTE_SUSPEND,
+    #[cfg(any(
+        esp_idf_version_major = "4",
+        all(esp_idf_version_major = "5", esp_idf_version_minor = "0"),
```
ivmarkov, line 139: Nit: simplify this by replacing with `all(esp_idf_version_major = "5", any(esp_idf_version_minor = "0", esp_idf_version_minor = "1"))`

src/bt/gap.rs
```diff
@@ -3,6 +3,13 @@
 
 use core::cmp::min;
 use core::convert::TryInto;
+#[cfg(any(
+    esp_idf_version_major = "4",
+    all(esp_idf_version_major = "5", esp_idf_version_minor = "0"),
```
ivmarkov, line 8: Ditto.

src/eth.rs
```diff
@@ -462,6 +462,8 @@ where
                     spi_host_id: host,
                     spi_devcfg: &spi_devcfg as *const _ as *mut _,
                     int_gpio_num: int,
+                    #[cfg(all(esp_idf_version_major = "5", esp_idf_version_minor = "2",))]
```
ivmarkov, line 465: This is not OK, as we'll have to touch it again once 5.3 or 6.0 is out. Rather, replace it with `not(<the upper condition>)`.

src/eth.rs
```diff
@@ -490,6 +492,8 @@ where
...
```
ivmarkov, line 495: Ditto.

src/eth.rs
```diff
@@ -518,6 +522,8 @@ where
...
```
ivmarkov, line 525: Ditto.

src/wifi.rs
```diff
@@ -1888,67 +1924,65 @@ impl<'a> EspEventDeserializer for WifiEvent<'a> {
...
+            wifi_event_t_WIFI_EVENT_AP_STOP => WifiEvent::ApStopped,
+            wifi_event_t_WIFI_EVENT_AP_STACONNECTED => WifiEvent::ApStaConnected,
+            wifi_event_t_WIFI_EVENT_AP_STADISCONNECTED => WifiEvent::ApStaDisconnected,
+            wifi_event_t_WIFI_EVENT_AP_PROBEREQRECVED => WifiEvent::ApProbeRequestReceived,
+            wifi_event_t_WIFI_EVENT_FTM_REPORT => WifiEvent::FtmReport,
+            wifi_event_t_WIFI_EVENT_STA_BSS_RSSI_LOW => WifiEvent::StaBssRssiLow,
+            wifi_event_t_WIFI_EVENT_ACTION_TX_STATUS => WifiEvent::ActionTxStatus,
+            wifi_event_t_WIFI_EVENT_STA_BEACON_TIMEOUT => WifiEvent::StaBeaconTimeout,
+            wifi_event_t_WIFI_EVENT_ROC_DONE => WifiEvent::RocDone,
+            #[cfg(all(esp_idf_version_major = "5", esp_idf_version_minor = "2",))]
```
ivmarkov, line 1970: Ditto - see above.

src/wifi.rs
```diff
@@ -1800,6 +1800,39 @@ impl<'d> NetifStatus for EspWifi<'d> {
 #[repr(transparent)]
 pub struct WpsCredentialsRef(wifi_event_sta_wps_er_success_t__bindgen_ty_1);
 
+#[cfg(all(esp_idf_version_major = "5", esp_idf_version_minor = "2",))]
```
ivmarkov, line 1803: Ditto - see above.

src/wifi.rs
```diff
@@ -1800,6 +1800,39 @@ impl<'d> NetifStatus for EspWifi<'d> {
...
+#[derive(Copy, Clone, Debug)]
+#[repr(C)]
+pub struct HomeChannelChange {
+    old_chan: u8,
+    old_snd: WifiSecondChan,
+    new_chan: u8,
+    new_snd: WifiSecondChan,
+}
+
+#[cfg(all(esp_idf_version_major = "5", esp_idf_version_minor = "2",))]
```
ivmarkov, line 1813: Ditto - see above.

src/wifi.rs
```diff
@@ -1800,6 +1800,39 @@ impl<'d> NetifStatus for EspWifi<'d> {
...
+#[derive(Copy, Clone, Debug)]
+enum WifiSecondChan {
+    None = 0,
+    Above,
+    Below,
+}
+
+#[cfg(all(esp_idf_version_major = "5", esp_idf_version_minor = "2",))]
```
ivmarkov, line 1821: Ditto - see above.

src/wifi.rs
```diff
@@ -1873,6 +1906,9 @@ pub enum WifiEvent<'a> {
     FtmReport,
     ActionTxStatus,
     RocDone,
+
+    #[cfg(all(esp_idf_version_major = "5", esp_idf_version_minor = "2",))]
```
ivmarkov, line 1910: Ditto - see above.

ivmarkov (changes requested):
I think it is in a very good shape. However, if you could fix the `#cfg`s - see my other comments.

MartinBroers:
Thanks @ivmarkov , some good suggestions.

MartinBroers:
Updated MR to 
- Compile again. Something in rust-nightly has changed causing clippy errors. Fixed those in the last commit
- Not use BT in the sample sdkconfig, so we don't need the update on esp-idf-sys.

src/http/client.rs
```diff
@@ -9,10 +9,9 @@ use core::cell::UnsafeCell;
 use core::fmt::Write as _;
 
 extern crate alloc;
-use alloc::boxed::Box;
 use alloc::collections::BTreeMap;
-use alloc::string::String;
-use alloc::string::ToString;
+#[cfg(not(feature = "default"))]
```
ivmarkov, line 13: Should be `feature = "std"`, NOT `feature = "default"`.

src/http/server.rs
```diff
@@ -40,13 +40,11 @@ use core::time::*;
 use core::{ffi, ptr};
 
 extern crate alloc;
-use alloc::borrow::ToOwned;
-use alloc::boxed::Box;
+#[cfg(not(feature = "default"))]
```
ivmarkov, line 43: Ditto.

src/mdns.rs
```diff
@@ -3,9 +3,8 @@
 use core::time::Duration;
 
 extern crate alloc;
-use alloc::string::String;
-use alloc::string::ToString;
-use alloc::vec::Vec;
+#[cfg(not(feature = "default"))]
+use alloc::{string::String, string::ToString, vec::Vec};
```
ivmarkov, line 7: Ditto.

src/mqtt/client.rs
```diff
@@ -1,10 +1,10 @@
 //! MQTT protocol client
-use core::convert::{TryFrom, TryInto};
 use core::ffi::c_void;
 use core::fmt::Debug;
 use core::{slice, time};
 
 extern crate alloc;
+#[cfg(not(feature = "default"))]
```
ivmarkov, line 7: Ditto

src/timer.rs
```diff
@@ -10,11 +10,11 @@
 //! microsecond time resolution, and 52-bit range.
 
 use core::num::NonZeroU32;
-use core::result::Result;
 use core::time::Duration;
 use core::{ffi, ptr};
 
 extern crate alloc;
+#[cfg(not(feature = "default"))]
```
ivmarkov, line 17: Ditto

src/wifi.rs
```diff
@@ -1,11 +1,11 @@
 //! WiFi support
-use core::convert::{TryFrom, TryInto};
 use core::marker::PhantomData;
 use core::str::Utf8Error;
 use core::time::Duration;
 use core::{cmp, ffi, fmt};
 
 extern crate alloc;
+#[cfg(not(feature = "default"))]
```
ivmarkov, line 8: Ditto.

src/ws/client.rs
```diff
@@ -1,10 +1,13 @@
 //! WebSocket client
 
-use core::convert::{TryFrom, TryInto};
 use core::{ffi, time};
+#[cfg(not(feature = "default"))]
```
ivmarkov, line 4: Ditto.

.github/configs/sdkconfig.defaults
```diff
@@ -18,17 +18,7 @@ CONFIG_ETH_SPI_ETHERNET_W5500=y
...
-CONFIG_BT_CLASSIC_ENABLED=y
-CONFIG_BTDM_CTRL_MODE_BLE_ONLY=n
-CONFIG_BTDM_CTRL_MODE_BR_EDR_ONLY=y
-CONFIG_BTDM_CTRL_MODE_BTDM=n
-CONFIG_BT_A2DP_ENABLE=y
-CONFIG_BT_HFP_ENABLE=y
-CONFIG_BT_HFP_CLIENT_ENABLE=y
-CONFIG_BT_HFP_AUDIO_DATA_PATH_HCI=y
-CONFIG_BT_BLE_42_FEATURES_SUPPORTED=y
+CONFIG_BT_ENABLED=n
```
ivmarkov, line 21: Please restore the BT settings, but commented out, and put a short comment on top: "Uncomment to enable Classic BT"

src/espnow.rs
```diff
@@ -10,6 +10,7 @@ use core::marker::PhantomData;
 
 use ::log::info;
 
+#[cfg(not(feature = "default"))]
```
ivmarkov, line 13: See below.

src/eth.rs
```diff
@@ -6,6 +6,7 @@ use core::{ffi, ptr};
 use ::log::*;
 
 extern crate alloc;
+#[cfg(not(feature = "default"))]
```
ivmarkov, line 9: See below.

src/eventloop.rs
```diff
@@ -1,13 +1,12 @@
...
 
-use core::convert::{TryFrom, TryInto};
 use core::fmt::Debug;
 use core::marker::PhantomData;
-use core::result::Result;
 use core::time::Duration;
 use core::{ffi, mem, ptr, slice};
 
 extern crate alloc;
+#[cfg(not(feature = "default"))]
```
ivmarkov, line 9: See below.

ivmarkov (changes requested):

ivmarkov:
@MartinBroers W.r.t. the clippy errors... I'm thinking we need a more generic fix. As in - for example - not importing the `std` prelude - ever - even if the `std` feature is enabled. Otherwise we need to sprincle the whole code with conditional imports, which is not nice at all.

PR #371, opened 2024-02-21
Invert the logic of Channel and Receiver

embediver:
Previously Receiver held a Weak<Channel<T>>,
now Channel holds a Weak<Receiver<T>>.
The Receiver now holds the Condvar and State<T>.
Also Receiver<T> becomes a Arc<Receiver<T>>.

By this the Channel can be dropped while the
Receiver is waiting.


embediver:
@ivmarkov Do you have other opinions or should this PR be closed?

PR #376, opened 2024-03-03
Support disabling SoftAP

DavidVentura:
I think `wrap_all` itself should be gated by the `cfg` but wasn't sure if it was desirable or not

### DIFF
--- a/src/netif.rs
+++ b/src/netif.rs
@@ -26,7 +26,7 @@ use crate::private::mutex;
 pub enum NetifStack {
     /// Station mode (WiFi client)
     Sta,
-    #[cfg(esp_idf_softap_support)]
+    #[cfg(esp_idf_esp_wifi_softap_support)]
     /// Access point mode (WiFi router)
     Ap,
     /// Ethernet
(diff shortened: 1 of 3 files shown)


src/netif.rs
```diff
@@ -26,6 +26,7 @@ use crate::private::mutex;
 pub enum NetifStack {
     /// Station mode (WiFi client)
     Sta,
+    #[cfg(esp_idf_softap_support)]
```
ivmarkov, line 29: Where is this defined? The PR is currently breaking the compilation of the examples, because this `cfg` is actually not defined by default?

DavidVentura: Not sure it is defined by default, but it is used in the C side of the sdk. How shall I address the default value?

DavidVentura: CONFIG_ESP_WIFI_SOFTAP_SUPPORT
https://docs.espressif.com/projects/esp-idf/en/v5.1.3/esp32/api-reference/kconfig.html?highlight=softap_support#config-esp-wifi-softap-support

ivmarkov: Then the `cfg` on the Rust side needs to be named `esp_idf_esp_wifi_softap_support`.

DavidVentura: I'll update the PR, though unexpectedly this works in my local branch; enabling/disabling has the expected effect

ivmarkov: > I'll update the PR, though unexpectedly this works in my local branch; enabling/disabling has the expected effect

I do not understand. 
`esp_idf_softap_support` definitely does **not** work, and is incorrect, because `CONFIG_XYZ` ESP IDF configs are translated by `esp-idf-sys` into `cfg(esp_idf_<the-esp-idf-config-lowercased-and-with-stripped-CONFIG_-prefix>)`.

Or else the Wifi examples would've compiled, because `CONFIG_ESP_WIFI_SOFTAP_SUPPORT` - as per the docu - is enabled by default.

ivmarkov: Btw, you also need to update the failing examples, not to be compiled when this config is not defined.

ivmarkov:
Looks fine, thanks!

PR #379, opened 2024-03-05
Fix WPS regression and cred conversion

### LINKED ISSUE #337: Enabling `esp_idf_svc::ws::client`, documentation needed
I want to use [esp_idf_svc::ws::client](https://esp-rs.github.io/esp-idf-svc/esp_idf_svc/ws/client/index.html). Based on the [source code](https://esp-rs.github.io/esp-idf-svc/src/esp_idf_svc/ws.rs.html#5-14) that pulls in the client module

```
#[cfg(all(
    feature = "alloc",
    esp_idf_comp_tcp_transport_enabled,
    esp_idf_comp_esp_tls_enabled,
    any(
        all(esp_idf_version_major = "4", esp_idf_ws_transport),
        esp_idf_comp_espressif__esp_websocket_client_enabled
    )
))]
pub mod client;
```

and guessing that I'm meant to remove the `esp_idf` prefix, capitalize, prepend `CONFIG`, and add it to `sdkconfig.defaults`, I added the following to my `sdkconfig.defaults`:

```
CONFIG_COMP_TCP_TRANSPORT_ENABLED=y
CONFIG_COMP_ESP_TLS_ENABLED=y
CONFIG_COMP_ESPRESSIF__ESP_WEBSOCKET_CLIENT_ENABLED=y
```

However, `cargo doc --open` still doesn't show the `client` module in `esp_idf_svc::ws`. I found an (exhaustive?) list of variables in `./target/xtensa-esp32s3-espidf/debug/build/esp-idf-sys-3a66d5767bd3b7f0/out/sdkconfig` but the above variables don't appear even after changing my `sdkconfig.defaults`.

Much of esp_idf_svc contains no documentation at all. As a newcomer to the ESP32 platform, getting a footing is difficult. Adding good documentation is a lot of work of course. I think outlining how to enable or disable each (forgive the abuse of terminology) feature is low-hanging fruit that would have a lot of value.

torkleyy:
Fixes #337
Also fixes the conversion of ssid/password in the config to no longer assume 0 termination - this lead to problems when routers send a 64 byte access token to us instead of a password

### DIFF
--- a/src/private/cstr.rs
+++ b/src/private/cstr.rs
@@ -41,6 +41,14 @@ pub fn from_cstr(buf: &[u8]) -> &str {
     from_cstr_fallible(buf).unwrap()
 }
 
+pub fn array_to_heapless_string_failible<const N: usize>(arr: [u8; N]) -> heapless::String<N> {
+    heapless::String::from_utf8(heapless::Vec::from_slice(&arr).unwrap())
+}
+
+pub fn array_to_heapless_string<const N: usize>(arr: [u8; N]) -> heapless::String<N> {
+    array_to_heapless_string_failible(arr).unwrap()
+}
+
 #[cfg(feature = "alloc")]
 pub struct RawCstrs(alloc::vec::Vec<CString>);
 
--- a/src/private/cstr.rs
+++ b/src/private/cstr.rs
@@ -20,6 +20,19 @@ pub fn set_str(buf: &mut [u8], s: &str) -> Result<(), EspError> {
     Ok(())
 }
 
+pub fn set_str_no_termination_requirement(buf: &mut [u8], s: &str) -> Result<(), EspError> {
+    if s.len() > buf.len() {
+        return Err(EspError::from_infallible::<ESP_ERR_INVALID_SIZE>());
+    }
+
+    buf[..s.len()].copy_from_slice(s.as_bytes());
+    if buf.len() != s.len() {
+        buf[s.len()] = 0;
+    }
+
+    Ok(())
+}
+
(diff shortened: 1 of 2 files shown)


ivmarkov:
> Fixes #337 Also fixes the conversion of ssid/password in the config to no longer assume 0 termination - this lead to problems when routers send a 64 byte access token to us instead of a password

@torkleyy 
For my understanding could you elaborate on the access token case you've mentioned?

torkleyy:
I'm not quite sure how it works, but some routers will send 64-byte access tokens instead of the passphrase. Until now, the code assumed that ssid / password are always 0-terminated, but it's actually possible for them to use the maximum length of 32/64.

I had no success actually connecting to the Wifi using that token on the esp (it does work on other devices). I ended up setting [`CONFIG_ESP_WIFI_WPS_PASSPHRASE`](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/kconfig.html#config-esp-wifi-wps-passphrase) to receive the passphrase instead.

ivmarkov:
I've delayed merging this for way too long, and now it has a conflict. :(
@torkleyy if you don't have time or will to resolve the conflicts in this PR, I'll understand and will do over the weekend.

(BTW: I've finally checked the (under-documented) ESP IDF Wifi API, and I confirm you are likely correct in the weird behavior w.r.t. SSID (and likely password)  in that is is a null-terminated string except when it is full length).

ivmarkov:
BTW, I'm worried that we might have other places to fix: specifically, the _reverse_ conversion of native SSID into `heapless::String` SSID. But these we can of course fix later.

torkleyy:
@ivmarkov rebased.

ivmarkov:
Thank you!

PR #380, opened 2024-03-05
Revert event wait functions back to immutable

torkleyy:
Also adds the new mutable versions as `_mut`

### DIFF
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -1247,17 +1247,6 @@ where
 
         wait.wait_while(|| matcher(self), timeout).await
     }
-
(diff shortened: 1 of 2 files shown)


ivmarkov (changes requested):
Let's removbe the `_mut` methods. I'm not sure they give that much of a value other than the fact that the user code that waits can also modify the state of the driver while waiting. But why should it do that...?

torkleyy:
I don't know why, but the functions were changed so I assumed that somebody needed that functionality...

ivmarkov:
@torkleyy Sorry for the delay in replying here: the examples don't really build after that change?

PR #381, opened 2024-03-06
Reflect changes in embedded-svc wifi ClientConfiguration

embediver:
Reflect the changes to `ClientConfiguration` in `embedded-svc` which will be introduced by https://github.com/esp-rs/embedded-svc/pull/75.

Before marking it ready, https://github.com/esp-rs/embedded-svc/pull/75 has to be merged and the dependency to `embedded-svc` has to be changed back in `Cargo.toml`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,11 +43,13 @@ enumset = { version = "1", default-features = false }
 log = { version = "0.4", default-features = false }
 uncased = { version = "0.9.7", default-features = false }
 embedded-hal-async = { version = "1", default-features = false }
-# embedded-svc = { version = "0.27", default-features = false }
-embedded-svc = { git = "https://github.com/embediver/embedded-svc.git", default-features = false }
+embedded-svc = { version = "0.27", default-features = false }
 esp-idf-hal = { version = "0.43", default-features = false }
 embassy-time-driver = { version = "0.1", optional = true, features = ["tick-hz-1_000_000"] }
 embassy-futures = "0.1"
+
+[patch.crates-io]
+embedded-svc = { git = "https://github.com/embediver/embedded-svc.git" }
 
 [build-dependencies]
 embuild = "0.31.3"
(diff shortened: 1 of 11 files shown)


Cargo.toml
```diff
@@ -43,7 +43,8 @@ enumset = { version = "1", default-features = false }
 log = { version = "0.4", default-features = false }
 uncased = { version = "0.9.7", default-features = false }
 embedded-hal-async = { version = "1", default-features = false }
-embedded-svc = { version = "0.27", default-features = false }
+# embedded-svc = { version = "0.27", default-features = false }
+embedded-svc = { git = "https://github.com/embediver/embedded-svc.git", default-features = false }
```
ivmarkov, line 47: Once this PR is ready for merge _and_ the corresponding one in `embedded-svc` is merged, we should restore the original dependency, and then put a `[patch.crates-io]` entry that points to the `embedded-svc` version from GIT.

ivmarkov (approved):
Looks OK. Once you are ready with both PRs (i.e. un-draft these), I'll merge. Just make sure to first fix the `embedded-svc` dependnency, so that we won't break the CI.

Cargo.toml
```diff
@@ -48,6 +48,9 @@ esp-idf-hal = { version = "0.43", default-features = false }
 embassy-time-driver = { version = "0.1", optional = true, features = ["tick-hz-1_000_000"] }
 embassy-futures = "0.1"
 
+[patch.crates-io]
```
ivmarkov, line 51: I've merged the PR in `embedded-svc` now. Would you mind changing this to `https://github.com/esp-rs/embedded-svc.git`?

ivmarkov (changes requested):

ivmarkov:
@embediver Sorry, just noticed that the CI fails - can you fix the failing example? Thanks!

embediver:
> @embediver Sorry, just noticed that the CI fails - can you fix the failing example? Thanks!

Sorry didn't notice that one example was feature-gated and forgot to check it.

But its fixed now.

ivmarkov:
Thanks!

Issue #384, opened 2024-03-12
EspEncryptedNVS cannot find partition by name


MartinBroers:
When opening an encrypted partition via src/nvs.rs, it tries to find the partition by name. However, the code seems not to work. I have created a minimal viable sample in https://github.com/MartinBroers/rust_c_ffi (don't mind the naming there; it is all about the code). The way the Option<&str> is parsed into an CString object somewhere goes awry, maybe because references are dropped in the meantime?

Run the code in the linked repo by running `cargo run` and the expected output is `found it`, But it prints garbage in the `puts` call as well. Can imagine the `strcmp` call also fails.... The rust code converts a &str to CString which is used in calling 2 separate C functions, just as esp-idf-svc does when calling `esp_partition_find_first`.

`NvsEncrypted::init` called via `new` takes an `key_partition: Option<&str>,`  and converts this into a CString or `core::ptr::null` when the option is None. The code works when the key_partition name is not provided, but that hinges on only having a single key partition, which may not cover all use cases.

PR #387, opened 2024-03-15
Fix: NVS encrypted partition name lookup

MartinBroers:
The partition name is wrapped in an option, in the case the option has a value, the pointer to the underlying value is not valid when calling esp_partition_find_first. By keeping the reference valid while calling C, it remains valid.

Fixes https://github.com/esp-rs/esp-idf-svc/issues/384

### DIFF
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -157,15 +157,14 @@ impl NvsEncrypted {
             None
         };
 
-        let c_key_partition = c_key_partition
-            .map(|p| p.as_ptr())
-            .unwrap_or(core::ptr::null());
-
         let keys_partition_ptr = unsafe {
             esp_partition_find_first(
                 esp_partition_type_t_ESP_PARTITION_TYPE_DATA,
                 esp_partition_subtype_t_ESP_PARTITION_SUBTYPE_DATA_NVS_KEYS,
-                c_key_partition,
+                match c_key_partition {
+                    Some(ref v) => v.as_ptr(),
+                    None => core::ptr::null(),
+                },
             )
         };
 


Vollbrecht (approved):
Thanks! LGTM @ivmarkov

src/nvs.rs
```diff
@@ -157,15 +157,14 @@ impl NvsEncrypted {
             None
         };
 
-        let c_key_partition = c_key_partition
```
ivmarkov, line 160: Good catch!

Mentioning below jfyi, your solution would work as well:
Don'we just need one extra `.as_ref()` here, right after `c_key_partition`?
Unless I don't understand Rust, the validity of the original, un-shadowed `c_key_partition` should be all the way down to the end of the function? It is just that when `map`-ing it, I accidentally unwrap an `Option<CString>` instead of an `Option<&CString>` thus destroying the value.

The reason wjhy Rust did not err on that that is because `p.as_ptr()` returns a raw pointer and not really a reference, so Rust does not give a damn that this pointer immediately becomes invalid.

MartinBroers: Yes,  I think if we `ref` the value, we could've left the code as it was.
Both seem to work as you describe.

Thanks!
```
 let c_key_partition = match c_key_partition { =>
     Some(ref v) => v.as_ptr(), => &CString
     None => core::ptr::null(),
 };
 unsafe {
       esp_partition_find_first(.., .., c_key_partition); 
 }
```

```
let c_key_partition = c_key_partition => *const i8, Option<CString>
    .as_ref() => Option<&CString>
    .map(|p| p.as_ptr()) => Option<*const i8>, &CString
    .unwrap_or(core::ptr::null()); <- (default)
unsafe {
       esp_partition_find_first(.., .., c_key_partition); 
}
```

PR #393, opened 2024-03-20
ota: handle partition errors gracefully

armandas:
When trying out the OTA functionality, my code was crashing when calling `get_boot_slot()` function. The function returns a Result, so crashing on error is rather counter-intuitive.

This PR adds graceful handling of calls `esp_ota_get_*_partition()` functions.


Vollbrecht (approved):
Thanks! LGTM

Issue #389, opened 2024-03-17
No server verification option set in esp_tls_cfg_t structure


weiying-chen:
1. I generated a project with [esp-idf-template](https://github.com/esp-rs/esp-idf-template).
2. I copied the code from [here](https://github.com/esp-rs/esp-idf-svc/blob/master/src/wifi.rs).
3. And tried to make a simple HTTP GET request:

```
fn get_request(client: &mut HttpClient<EspHttpConnection>) -> anyhow::Result<()> {
    let url = "https://httpbin.org/get";
    let request = client.get(url)?;
    info!("-> GET {}", url);
    let response = request.submit()?;
    info!("<- {}", response.status());
    std::thread::sleep(core::time::Duration::from_secs(5));
    Ok(())
}
```

I got this error:

```
E (4918) esp-tls-mbedtls: No server verification option set in esp_tls_cfg_t structure. Check esp_tls API reference
E (4918) esp-tls-mbedtls: Failed to set client configurations, returned [0x8017] (ESP_ERR_MBEDTLS_SSL_SETUP_FAILED)
E (4928) esp-tls: create_ssl_handle failed
E (4928) esp-tls: Failed to open new connection
E (4938) transport_base: Failed to open a new connection
E (4948) HTTP_CLIENT: Connection failed, sock < 0
```

In this repository (esp-idf-svc), there is mention of "esp_tls" in the following files: src/tls.rs, ws.rs, src/mqtt/client.rs

What am I supposed to add to my project to stop getting that error?

Note: [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=af31e697da921bf3c98bcf3296f35485) is what my entire file looks like.

ivmarkov:
[This](https://github.com/ivmarkov/rust-esp32-std-demo/blob/58b9ea6680d282dcb26c0a77b8108f4ae2494b48/src/main.rs#L673). The example is for mqtt, but https client wqorks the same.

weiying-chen:
I guess it should be something like this, but what should I put in `??`? 

```
    let config = ?? {
        crt_bundle_attach: Some(esp_idf_svc::sys::esp_crt_bundle_attach),
        ..Default::default()
    };

    let mut client = HttpClient::wrap(EspHttpConnection::new(&config)?);
```

weiying-chen:
Thanks a lot! This worked.

```
[...]

PR #394, opened 2024-03-21
Add comment with HTTPS example

weiying-chen:
This PR will hopefully prevent [this](https://github.com/esp-rs/esp-idf-svc/issues/389) from happening again.

At first, I thought about creating a separate `https_client.rs`, but that would create a lot of duplicate code (also it'd look strange if `http_client.rs` had an HTTPS version and `http_server.rs` didn't). I feel the comment will get the message across with no code duplication.

Note: There's aliasing here: 

    use esp_idf_svc::http::client::{Configuration as HttpConfiguration, EspHttpConnection};

To prevent conflict with this:

```
use embedded_svc::{
    // More imports
    wifi::{AuthMethod, ClientConfiguration, Configuration},
};
```

### DIFF
--- a/examples/http_client.rs
+++ b/examples/http_client.rs
@@ -37,7 +37,20 @@ fn main() -> anyhow::Result<()> {
 
     connect_wifi(&mut wifi)?;
 
-    // Create HTTP(S) client
+    // Create HTTP client
+    //
+    // Note: To send a request to an HTTPS server, you can do:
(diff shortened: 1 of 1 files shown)


Vollbrecht (approved):
Yeah having it clarified in the http client example should be enough, so no need for a separate https example. 

Thanks for the PR!

PR #396, opened 2024-03-21
Exposes WifiEvent#ApStaConnected and WifiEvent#ApStaDisconnected event info

NickUfer:
This PR exposes the [wifi_event_ap_staconnected_t](https://docs.espressif.com/projects/esp-idf/en/v4.3.7/esp32/api-reference/network/esp_wifi.html#_CPPv428wifi_event_ap_staconnected_t)  and [wifi_event_ap_stadisconnected_t](https://docs.espressif.com/projects/esp-idf/en/v4.3.7/esp32/api-reference/network/esp_wifi.html#_CPPv431wifi_event_ap_stadisconnected_t) structs for the WifiEvents `ApStaConnected` and `ApStaDisconnected` respectively.

I tested it so far with an ESP-WROOM-32 development board.

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -1945,15 +1945,31 @@ impl ApStaDisconnectedRef {
     pub fn is_mesh_child(&self) -> bool {
         self.0.is_mesh_child
     }
+
+    /// reason of disconnection
+    #[cfg(not(any(
+        esp_idf_version_major = "4",
+        all(esp_idf_version_major = "5", esp_idf_version_minor = "0"),
+    )))]
+    pub fn reason(&self) -> u8 {
+        self.0.reason
+    }
 }
 
 impl fmt::Debug for ApStaDisconnectedRef {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("ApStaDisconnectedRef")
-            .field("mac", &self.mac())
+        let mut ds = f.debug_struct("ApStaDisconnectedRef");
+        ds.field("mac", &self.mac())
             .field("aid", &self.aid())
-            .field("is_mesh_child", &self.is_mesh_child())
-            .finish()
+            .field("is_mesh_child", &self.is_mesh_child());
+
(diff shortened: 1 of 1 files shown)


src/wifi.rs
```diff
@@ -1895,6 +1903,60 @@ impl TryFrom<&WpsCredentialsRef> for WpsCredentials {
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("ApStaConnectedRef")
+            .field("mac", &self.mac())
+            .field("aid", &self.aid())
+            .field("is_mesh_child", &self.is_mesh_child())
+            .finish()
+    }
+}
+
+impl ApStaDisconnectedRef {
```
Vollbrecht, line 1933: does the C disconnect struct also support an extra field "reason" if we look [here](https://github.com/espressif/esp-idf/blob/4f3cd0deb9c79c8282da4938a29d265705a57564/components/esp_wifi/include/esp_wifi_types_generic.h#L857)? 

Since otherwise the structs would be pretty much identical and we could have unified them.

NickUfer: Hmm, interesting. I was only looking into the official 4.3.7 documentation and it does not include this reason field, but newer versions do include it. It seems to have been added in 5.1.3

NickUfer: Okay, I managed to add it via cfg attributes.
As the exact type of this reason is not specified either in the bindings nor in the documentation, I would assume it is a subset of [wifi_err_reason_t](https://github.com/espressif/esp-idf/blob/4f3cd0deb9c79c8282da4938a29d265705a57564/components/esp_wifi/include/esp_wifi_types_generic.h#L78-L140).
This is probably supposed to be mapped to an enum, but it seems to me that these types of structs and enums are maintained in https://github.com/esp-rs/embedded-svc and a From/TryFrom implementation for Newtype<Binding> is created in the esp-idf-svc project.

I should probably create another PR in the embedded-svc project to add the enum for `wifi_err_reason_t` right?

PR #398, opened 2024-03-25
Add EspNetif::set_ip_info

### LINKED ISSUE #332: Cannot change network interface after creation
Many methods of `EspNetif` are not exposed, such as setters and dhcp control. Without this, the ip configuration of an established connection cannot be changed.

torkleyy:
Fixes #332


ivmarkov:
Limited access to internet right now. Will follow up when I have an opportunity

torkleyy:
Sure, there's no rush.

Vollbrecht:
from the issue conversation ivmarkov thought that it may be problematic to update such values when the netif is in the up state, But you thought based on the docu that it still should work. So you can confirm that calling esp_netif_set_ip_info does not produce sideeffects / e.g its save to be callable in any state?

Vollbrecht:
@ivmarkov @torkleyy bumping anything you have to say here to bring this forward?

torkleyy:
I haven't tested this much as we follow a different approach atm. In any case there is no way to reenable DHCP with just this method added.

torkleyy:
FYI, we've been using this in production for many months now, we noticed that for wifi, you have to disconnect and reconnect for it to take effect - this is not the case for Ethernet.

PR #404, opened 2024-04-02
Flush responses to avoid repeated request failures

### LINKED ISSUE #399: Only a few post requests are sent before program panics
I'm sending post requests every 5 seconds inside a loop that will break when the iterations reach 10:

```
let mut iterations = 0;

loop {
    if scale.is_ready() {
        log::info!("Iteration number {}", iterations);
        
        // logic to set payload
     
        post_request(&mut client, payload_bytes)?;

        log::info!("Post request sent!");
    }

    FreeRtos::delay_ms(5000u32);

    iterations += 1;

    if iterations >= 10 {
        break;
    }
}
```

This is the `post_request` function:

```
fn post_request(client: &mut HttpClient<EspHttpConnection>, payload: &[u8]) -> anyhow::Result<()> {
    // set the key, url, and headers 

    let mut request = client.post(supabase_url, &headers)?;

    request.write_all(payload)?;
    request.flush()?;
    info!("-> POST {}", supabase_url);

    match request.submit() {
        Ok(response) => {
            let status = response.status();
            info!("<- HTTP Status: {}", status);
        }
        Err(e) => {
            info!("Error sending POST request: {:?}", e);
        }
    }

    Ok(())
}
```

The first POST requests will be sent. But then some warning and errors will show up and the program will panic and exit:

```
I (21162) rust_esp32c3_hx711: Iteration: 1
I (21502) esp-x509-crt-bundle: Certificate validated
I (22222) rust_esp32c3_hx711: -> POST https://pratqgdulutgohggfwfo.supabase.co/rest/v1/messages
I (22812) rust_esp32c3_hx711: <- HTTP Status: 201
I (22812) rust_esp32c3_hx711: Post request sent!
I (27812) rust_esp32c3_hx711: Iteration: 2
I (33152) rust_esp32c3_hx711: -> POST https://pratqgdulutgohggfwfo.supabase.co/rest/v1/messages
W (38472) HTTP_CLIENT: Connection timed out before data was ready!
I (38472) rust_esp32c3_hx711: Error sending POST request: EspIOError(EspError(-28679))
I (38472) rust_esp32c3_hx711: Post request sent!
I (43482) rust_esp32c3_hx711: Iteration: 3
thread 'main' panicked at /home/weiying-chen/.cargo/registry/src/index.crates.io-6f17d22bba15001f/esp-idf-svc-0.48.1/src/http/client.rs:477:13:
connection is not in initial phase
````

What could be the issue here?

Notes: 
- It's getting stuck here: `match request.submit()`. And producing the timeout.
- I [tested](https://gist.github.com/weiying-chen/af32ce18a2848ca3df8f763c83337cda) running the same loop but without `esp-idf-svc`. The loop finishes successfully and all the post requests are sent.

### PR DESCRIPTION
**Pull request by weiying-chen**
Fixes #399.

I tested by using a loop that sends many HTTPS requests and by removing the part that reads and drains the data in the `http_client.rs` example. The pattern was the same: `ESP_FAIL` without `flush_response()`. No `ESP_FAIL` with `flush_response()`.

Some considerations:

1. Will flushing the response data in `initiate_request()` cause issues that cannot be observed in the `http_client.rs` example?
2. If the response data is [flushed](https://github.com/weiying-chen/esp-idf-svc/blob/feature/flush-response/src/http/client.rs#L213) in `initiate_request()`, will [flushing](https://github.com/weiying-chen/esp-idf-svc/blob/feature/flush-response/src/http/client.rs#L452) the response data in `fetch_headers()` be redundant?

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -208,6 +208,11 @@ impl EspHttpConnection {
         uri: &'a str,
         headers: &'a [(&'a str, &'a str)],
     ) -> Result<(), EspError> {
+        // If response data from the previous request remains, subsequent requests may fail
+        if self.is_response_initiated() {
+            self.flush_response()?;
+        }
+
         self.assert_initial();
 
         let c_uri = to_cstring_arg(uri)?;
@@ -258,6 +263,13 @@ impl EspHttpConnection {
         esp!(unsafe { esp_http_client_open(self.raw_client, self.request_content_len as i32) })?;
 
         self.state = State::Request;
+
+        Ok(())
+    }
+
+    pub fn flush_response(&mut self) -> Result<(), EspError> {
+        let mut len = 0_i32;
+        esp!(unsafe { esp_http_client_flush_response(self.raw_client, &mut len) })?;
 
         Ok(())
     }
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -267,13 +267,6 @@ impl EspHttpConnection {
         Ok(())
     }
 
-    pub fn flush_response(&mut self) -> Result<(), EspError> {
-        let mut len = 0_i32;
-        esp!(unsafe { esp_http_client_flush_response(self.raw_client, &mut len) })?;
-
-        Ok(())
-    }
-
     pub fn is_request_initiated(&self) -> bool {
         self.state == State::Request
     }
@@ -365,6 +358,13 @@ impl EspHttpConnection {
         } else {
             Ok(())
         }
+    }
+
+    fn flush_response(&mut self) -> Result<(), EspError> {
+        let mut len = 0_i32;
+        esp!(unsafe { esp_http_client_flush_response(self.raw_client, &mut len) })?;
+
+        Ok(())
     }
 
     fn raw_write(&mut self, buf: &[u8]) -> Result<usize, EspError> {

### CODE REVIEW

**Comment by Vollbrecht** [comment]
Thanks for the PR!



> Some considerations:
> 
>     1. Will flushing the response data in `initiate_request()` cause issues that cannot be observed in the `http_client.rs` example?
> 
>     2. If the response data is [flushed](https://github.com/weiying-chen/esp-idf-svc/blob/feature/flush-response/src/http/client.rs#L213) in `initiate_request()`, will [flushing](https://github.com/weiying-chen/esp-idf-svc/blob/feature/flush-response/src/http/client.rs#L452) the response data in `fetch_headers()` be redundant?

Yeah that are some good questions. Potentially we have 3 states (new, request, response). Since you made the additional method public we need to know any site-effects on calling it in any of the 3 states, e.g it dosn't brake anything when its called randomly. 

Also what is happening if you call flush_response()  repeatedly? Is it save to do?

Does flushing len 0 in this context works as expeted, 0 meaning flush everything you have ?

**Comment by weiying-chen** [comment]
> Yeah that are some good questions. Potentially we have 3 states (new, request, response). Since you made the additional method public we need to know any site-effects on calling it in any of the 3 states, e.g it dosn't brake anything when its called randomly.
> 
> Also what is happening if you call flush_response() repeatedly? Is it save to do?
> 
> Does flushing len 0 in this context works as expeted, 0 meaning flush everything you have ?

Those are good questions too. I didn't think of them. I'll try to test each of those three items and get back to you!

**Comment by weiying-chen** [comment]
I'm back.

1. No `ESP_FAIL` if you run `flush_response()` [repeatedly](https://github.com/weiying-chen/esp-idf-svc/blob/test/flush-response-repeated/src/http/client.rs#L213).

2. No `ESP_FAIL` if you run `flush_response()` in a [public context](https://github.com/weiying-chen/esp-idf-svc/blob/test/flush-response-public/examples/http_client.rs#L99) (e.g. in `http_client.rs`).

3. No `ESP_FAIL` if you run `flush_response()` [when](https://github.com/weiying-chen/esp-idf-svc/blob/test/flush-response-states/src/http/client.rs#L211) the `State` is `New` and `Response`.  [There](https://github.com/weiying-chen/esp-idf-svc/blob/test/flush-response-states/src/http/client.rs#L270) is `ESP_FAIL` if `State` is `Request`. But I guess no one is going to run `flush_response` in the last situation? Or we should add a warning or if statement?

File: src/http/client.rs [hunk 1]
```diff
@@ -262,6 +267,13 @@ impl EspHttpConnection {
         Ok(())
     }
 
+    pub fn flush_response(&mut self) -> Result<(), EspError> {
```
Vollbrecht, line 270: if we call it everytime its needed in initate_request, is there any reason to have this methods still publicly available. E.g in what other scenario a user would need to call it?

Vollbrecht (changes requested):
I think if we can clarify and fix the last two bits i asked down below we are good two go !

weiying-chen:
> I think if we can clarify and fix the last two bits i asked down below we are good two go !

Okay, let me take a look and come back to you!

Vollbrecht:
> > I think if we can clarify and fix the last two bits i asked down below we are good two go !
> 
> Okay, let me take a look and come back to you!

One last thing can you make sure that the http_client,rs example also still works with the changes !

src/http/client.rs
weiying-chen: At first, I thought it'd be nice if the user could flush the response with `client.connection().flush_request()`. But then, like you, I thought: why use it if the response is going to be flushed anyway in `initiate_request`? Also, using `flush_request()` publicly could introduce issues unexpected issues. (Also, maybe we shouldn't introduce features that users aren't requesting.) 

So yeah, I think it's not worth making it public (I changed the code).

Vollbrecht:
Thanks for working on this! 

The http_client example still worked for me with the changes.

weiying-chen:
> Thanks for working on this!

I use this crate a lot. So I'm happy to help!

Issue #401, opened 2024-03-30
Possible Memory Leak in `EthDriver.set_rx_callback`


owenthewizard:
Please take this with a massive grain of salt...
It appears that `buf` is never getting freed here:
https://docs.esp-rs.org/esp-idf-svc/src/esp_idf_svc/eth.rs.html#754-764
Minimal example:
```rust
use esp_idf_svc::eth::{EthDriver, RmiiEthChipset};
use esp_idf_svc::eventloop::EspSystemEventLoop;
use esp_idf_svc::hal::gpio;
use esp_idf_svc::hal::prelude::Peripherals;
use esp_idf_svc::sys::EspError;

fn main() -> Result<(), EspError> {
    // It is necessary to call this function once. Otherwise some patches to the runtime
    // implemented by esp-idf-sys might not link properly. See https://github.com/esp-rs/esp-idf-template/issues/71
    esp_idf_svc::sys::link_patches();

    // Bind the log crate to the ESP Logging facilities
    esp_idf_svc::log::EspLogger::initialize_default();

    let peripherals = Peripherals::take()?;
    let sysloop = EspSystemEventLoop::take()?;
    let pins = peripherals.pins;

    let mut eth = EthDriver::new_rmii(
        peripherals.mac,
        pins.gpio25, // RMII RDX0
        pins.gpio26, // RMII RDX1
        pins.gpio27, // RMII CRS DV
        pins.gpio23, // ETH01 SMI MDC
        pins.gpio22, // EMII TXD1
        pins.gpio21, // RMII TX EN
        pins.gpio19, // RMII TXD0
        pins.gpio18, // WT32-ETH01 SMI MDIO
        esp_idf_svc::eth::RmiiClockConfig::<gpio::Gpio0, gpio::Gpio16, gpio::Gpio17>::Input(
            pins.gpio0, // WT32-ETH01 external clock
        ),
        Some(pins.gpio16), // WT32-ETH01 PHY reset
        RmiiEthChipset::LAN87XX,
        Some(1), // WT32-ETH01 PHY address
        sysloop,
    )?;

    log::warn!("Setting eth promiscuous...");
    unsafe {
        use esp_idf_svc::handle::RawHandle;
        use esp_idf_svc::sys::{esp_eth_io_cmd_t_ETH_CMD_S_PROMISCUOUS, esp_eth_ioctl};
        let handle = eth.handle();
        let res = esp_eth_ioctl(
            handle,
[...]

PR #406, opened 2024-04-08
Owned network callback parameters

owenthewizard:
Related: #401 

I think it would make sense for the *Driver callbacks to use owned parameters rather than references (slices). This way they can be e.g. sent via `mpsc::channel` without another allocation, e.g. `to_vec()`.

I'm not the most experienced with Rust so I'm very open to feedback and further guidance.

### DIFF
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -757,8 +757,7 @@ impl<'d, T> EthDriver<'d, T> {
         len: u32,
         event_handler_arg: *mut ffi::c_void,
     ) -> esp_err_t {
-        UnsafeCallback::from_ptr(event_handler_arg as *mut _)
-            .call(EthFrame::new(buf, len));
+        UnsafeCallback::from_ptr(event_handler_arg as *mut _).call(EthFrame::new(buf, len));
 
         ESP_OK
     }
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -1,7 +1,7 @@
 use core::fmt::Debug;
 use core::marker::PhantomData;
 use core::time::Duration;
-use core::{ffi, ptr, ops};
+use core::{ffi, ops, ptr};
 
 use ::log::*;
 
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -117,7 +117,7 @@ pub enum SpiEthChipset {
     KSZ8851SNL,
 }
 
-type RawCallback<'a> = Box<dyn FnMut(&[u8]) + Send + 'a>;
+type RawCallback<'a> = Box<dyn FnMut(EthFrame) + Send + 'a>;
 
 struct UnsafeCallback<'a>(*mut RawCallback<'a>);
 
@@ -135,7 +135,7 @@ impl<'a> UnsafeCallback<'a> {
         self.0.cast()
     }
 
-    unsafe fn call(&self, data: &[u8]) {
+    unsafe fn call(&self, data: EthFrame) {
         let reference = self.0.as_mut().unwrap();
 
         (reference)(data);
@@ -689,7 +689,7 @@ impl<'d, T> EthDriver<'d, T> {
 
     pub fn set_rx_callback<F>(&mut self, callback: F) -> Result<(), EspError>
     where
-        F: FnMut(&[u8]) + Send + 'static,
+        F: FnMut(EthFrame) + Send + 'static,
     {
         self.internal_set_rx_callback(callback)
     }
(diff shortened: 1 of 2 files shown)


src/eth.rs
```diff
@@ -857,6 +856,33 @@ impl<'d, T> RawHandle for EthDriver<'d, T> {
...
+    }
+
+    pub fn as_mut_slice(&mut self) -> &mut [u8] {
+        unsafe { core::slice::from_raw_parts_mut(self.buf, self.len as _) }
+    }
+}
+
+impl Drop for EthFrame {
+    fn drop(&mut self) {
+        unsafe { free(self.buf.cast()) };
```
owenthewizard, line 882: [Memory of buf is allocated in the Layer2, make sure it get free after process. ](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/network/esp_eth.html#_CPPv4N13esp_eth_mac_s7receiveE)

src/eth.rs
```diff
@@ -857,6 +856,33 @@ impl<'d, T> RawHandle for EthDriver<'d, T> {
...
 }
 
+pub struct EthFrame {
+    buf: *mut u8,
+    len: u32,
+}
+
+unsafe impl Send for EthFrame {}
+
+impl EthFrame {
```
ivmarkov, line 866: Not strictly necessary, but how about implementing `Deref` on `EthFrame`?

src/eth.rs
```diff
@@ -857,6 +856,33 @@ impl<'d, T> RawHandle for EthDriver<'d, T> {
...
+    const unsafe fn new(buf: *mut u8, len: u32) -> Self {
+        Self { buf, len }
+    }
+
+    pub const fn as_slice(&self) -> &[u8] {
+        unsafe { core::slice::from_raw_parts(self.buf, self.len as _) }
+    }
+
+    pub fn as_mut_slice(&mut self) -> &mut [u8] {
```
ivmarkov, line 875: I don't think this is a good idea. Why do you need a mutable access to the buffer?

src/wifi.rs
```diff
@@ -1396,6 +1391,34 @@ impl<'d> Wifi for WifiDriver<'d> {
...
+impl WifiFrame {
+    const unsafe fn new(buf: *mut u8, len: u16, eb: *mut ffi::c_void) -> Self {
+        Self { buf, len, eb }
+    }
+
+    pub const fn as_slice(&self) -> &[u8] {
+        unsafe { core::slice::from_raw_parts(self.buf, self.len as _) }
+    }
+
+    pub fn as_mut_slice(&mut self) -> &mut [u8] {
```
ivmarkov, line 1411: As per above, probably not a good idea. Also as per above, maybe implement `Deref`?

ivmarkov (changes requested):

src/eth.rs
owenthewizard: In my use case, per the sta2eth example in esp-idf I want to change the MAC address and some other data before sending it on its way. Otherwise I would have to do another allocation (e.g. into a Vec) beforehand.
Why don't you think it's a good idea? What are the downsides? It seems to be used mutably in esp-idf.

ivmarkov: Yeah. You can only change the data as long as the new data is not overrunning the frame size that is passed to you, while the original buffer might be larger, yet you don't know how large it actually is. Anyway. Leave it mutable but then if you implement `Deref` also implement `DerefMut`.

owenthewizard: That shouldn't be an issue since you can't change the length of a slice, right?

owenthewizard:
I haven't implemented the same changes for the Wi-Fi TX callback since I don't know where that memory is allocated/freed.

src/eth.rs
owenthewizard: `Deref` to `&[u8]` right? I feel like `as_slice()` is more clear than `*EthFrame`.

ivmarkov: You can make the same argument for everything that implements `Deref` then. Like`Bec`. Also:
* Deref should be in addition, not instead of `as_slice`
* It is `&EthFrame`, not `*EthFrame`

src/eth.rs
ivmarkov: What shouldn't be an issue?

ivmarkov: You don't know the original length of the whole buffer, just the length of the frame I side it. So if your transformation s result in a longer ethernet frame, you can't do these in-place, is what I'm saying.

owenthewizard: > if your transformations result in a longer ethernet frame
How would one do this? Isn't slice length immutable? I guess you could get a `*mut u8` from the `&mut [u8]` but that's `unsafe`.

ivmarkov:
Thanks!

PR #410, opened 2024-04-11
clippy lints

owenthewizard:
Many many lints. Notably:
- Remove unneeded `Result` wrapping some function return types.
- `if ... { panic! }` to `assert!`
- Replaced `if let` with `map_or` and similar.
- [Make `fn` `const` if possible.](https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn)
- [`ptr::addr_of!` instead of `&T as *const T` ](https://rust-lang.github.io/rust-clippy/master/index.html#borrow_as_ptr)
- [`.cast() instead of `*const T as *const U`](https://rust-lang.github.io/rust-clippy/master/index.html#ptr_as_ptr)
- [`.cast_mut()` instead of `*const T as *mut T`](https://rust-lang.github.io/rust-clippy/master/index.html#ptr_cast_constness)
- [`U::from(T)` instead of `T as U`](https://rust-lang.github.io/rust-clippy/master/index.html#cast_lossless)
- `"{}". foo` -> `"{foo}"`
- [Use `Self`](https://rust-lang.github.io/rust-clippy/master/index.html#use_self)
- ...

As many of these are subjective, feel free to modify/request modifications.

### DIFF
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -451,7 +451,7 @@ where
 
                 #[cfg(esp_idf_version_major = "4")]
                 let dm9051_cfg = eth_dm9051_config_t {
-                    spi_hdl: spi_handle.unwrap().cast_mut(),
+                    spi_hdl: spi_handle.unwrap().cast(),
                     int_gpio_num: int,
                 };
 
@@ -487,7 +487,7 @@ where
 
                 #[cfg(esp_idf_version_major = "4")]
                 let w5500_cfg = eth_w5500_config_t {
-                    spi_hdl: spi_handle.unwrap().cast_mut(),
+                    spi_hdl: spi_handle.unwrap().cast(),
                     int_gpio_num: int,
                 };
 
@@ -523,7 +523,7 @@ where
 
                 #[cfg(esp_idf_version_major = "4")]
                 let ksz8851snl_cfg = eth_ksz8851snl_config_t {
-                    spi_hdl: spi_handle.unwrap().cast_mut(),
+                    spi_hdl: spi_handle.unwrap().cast(),
                     int_gpio_num: int,
                 };
 
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -78,7 +78,7 @@ where
     GPIO16: Peripheral<P = gpio::Gpio16>,
     GPIO17: Peripheral<P = gpio::Gpio17>,
 {
-    fn eth_mac_clock_config(&self) -> eth_mac_clock_config_t {
+    const fn eth_mac_clock_config(&self) -> eth_mac_clock_config_t {
         let rmii = match self {
             Self::Input(_) => eth_mac_clock_config_t__bindgen_ty_2 {
                 clock_mode: emac_rmii_clock_mode_t_EMAC_CLK_EXT_IN,
@@ -246,7 +246,7 @@ impl<'d> EthDriver<'d, RmiiEth> {
 
         let eth = Self::init(
             Self::rmii_mac(rmii_mdc.pin(), rmii_mdio.pin(), &rmii_ref_clk_config),
-            Self::rmii_phy(chipset, rst, phy_addr)?,
+            Self::rmii_phy(chipset, rst, phy_addr),
             None,
             RmiiEth {},
             sysloop,
@@ -259,10 +259,10 @@ impl<'d> EthDriver<'d, RmiiEth> {
         chipset: RmiiEthChipset,
         reset: Option<i32>,
         phy_addr: Option<u32>,
-    ) -> Result<*mut esp_eth_phy_t, EspError> {
+    ) -> *mut esp_eth_phy_t {
         let phy_cfg = Self::eth_phy_default_config(reset, phy_addr);
 
-        let phy = match chipset {
+        match chipset {
             RmiiEthChipset::IP101 => unsafe { esp_eth_phy_new_ip101(&phy_cfg) },
             RmiiEthChipset::RTL8201 => unsafe { esp_eth_phy_new_rtl8201(&phy_cfg) },
             #[cfg(not(esp_idf_version = "4.3"))]
@@ -279,9 +279,7 @@ impl<'d> EthDriver<'d, RmiiEth> {
             RmiiEthChipset::KSZ8081 => unsafe { esp_eth_phy_new_ksz8081(&phy_cfg) },
             #[cfg(not(esp_idf_version_major = "4"))]
             RmiiEthChipset::KSZ80XX => unsafe { esp_eth_phy_new_ksz80xx(&phy_cfg) },
-        };
-
-        Ok(phy)
+        }
     }
 
     fn rmii_mac(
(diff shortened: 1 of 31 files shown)


ivmarkov:
> Many many lints. Notably:
> 
> * Remove unneeded `Result` wrapping some function return types.

I would accept this only for non-public API. Can you check and make sure you've done this change for private and crate-private methods only? If Clippy complains in public stuff, we should suppress its warnings there.

> * `if ... { panic! }` to `assert!`
> * Replaced `if let` with `map_or` and similar.
> * [Make `fn` `const` if possible.](https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn)
> * [`ptr::addr_of!` instead of `&T as *const T` ](https://rust-lang.github.io/rust-clippy/master/index.html#borrow_as_ptr)
> * [`.cast() instead of `*const T as *const U`](https://rust-lang.github.io/rust-clippy/master/index.html#ptr_as_ptr)
> * [`.cast_mut()` instead of `*const T as *mut T`](https://rust-lang.github.io/rust-clippy/master/index.html#ptr_cast_constness)
> * [`U::from(T)` instead of `T as U`](https://rust-lang.github.io/rust-clippy/master/index.html#cast_lossless)
> * `"{}". foo` -> `"{foo}"`
> * [Use `Self`](https://rust-lang.github.io/rust-clippy/master/index.html#use_self)
> * ...
> 
> As many of these are subjective, feel free to modify/request modifications.

The rest is fine. I don't have strong feelings for doing it one way or the other. These are stylistic changes and as you say, sometimes subjective.
As long as Clippy stays silent, I'm happy.

ivmarkov:
Actually one more thing: None of your improvements seem to address the actual clippy complaints which are currently failing the CI? All of the failures in the CI seem to be related to Clippy demanding that we use explicit turbofish typecasts when using `core::mem::transmute`. I think these should be addressed with priority.

ivmarkov:
And... your PR has come flicts. You need to rebase your PR on top of master. As in `git checkout <your-branch>`. Then `git pull`. Then `git rebase master --interactive`. ... and resolve all conflicts and upon resolving `git rebase --continue` until no conflicts are left and all your commits are on top of master.

ivmarkov:
... where `upstream/master` is the upstream master branch of esp-idf-svc. You can create the upstream remote repo with `git remote add https://github.com/esp-rs/esp-idf-svc`.

ivmarkov:
Ok conflicts seem to be fixed now, but clippy is still not happy. See the failing CI. I think given that this PR addresses Clippy lints, the CI should -as a minimum - be passing.

src/eth.rs
```diff
@@ -259,10 +259,10 @@ impl<'d> EthDriver<'d, RmiiEth> {
         chipset: RmiiEthChipset,
         reset: Option<i32>,
         phy_addr: Option<u32>,
-    ) -> Result<*mut esp_eth_phy_t, EspError> {
```
owenthewizard, line 262: Removed uneeded `Result` (private API).

src/eth.rs
```diff
@@ -717,14 +715,14 @@ impl<'d, T> EthDriver<'d, T> {
     ///
     /// This "local borrowing" will only be possible to express in a safe way once/if `!Leak` types
     /// are introduced to Rust (i.e. the impossibility to "forget" a type and thus not call its destructor).
-    pub unsafe fn set_nonstatic_rx_callback<F>(&mut self, callback: F) -> Result<(), EspError>
```
owenthewizard, line 720: Same as above (public API).

src/eth.rs
```diff
@@ -717,14 +715,14 @@ impl<'d, T> EthDriver<'d, T> {
...
+    pub unsafe fn set_nonstatic_rx_callback<F>(&mut self, callback: F)
     where
         F: FnMut(EthFrame) + Send + 'd,
     {
-        self.internal_set_rx_callback(callback)
+        self.internal_set_rx_callback(callback);
     }
 
-    fn internal_set_rx_callback<F>(&mut self, callback: F) -> Result<(), EspError>
```
owenthewizard, line 727: Private API, but needs to be changed to match public.

src/wifi.rs
```diff
@@ -394,7 +394,7 @@ pub trait NonBlocking {
 
     fn stop_wps(&mut self) -> Result<WpsStatus, EspError>;
 
-    fn is_wps_finished(&self) -> Result<bool, EspError>;
```
owenthewizard, line 397: Remove unneeded `Result` (private API).

src/wifi.rs
```diff
@@ -436,7 +436,7 @@ where
         (**self).stop_wps()
     }
 
-    fn is_wps_finished(&self) -> Result<bool, EspError> {
```
owenthewizard, line 439: As above.

src/wifi.rs
```diff
@@ -1175,12 +1180,12 @@ impl<'d> WifiDriver<'d> {
         }
     }
 
-    fn is_wps_finished(&self) -> Result<bool, EspError> {
```
owenthewizard, line 1178: As above.

src/wifi.rs
```diff
@@ -1711,7 +1716,7 @@ impl<'d> EspWifi<'d> {
         self.driver_mut().stop_wps()
     }
 
-    pub fn is_wps_finished(&self) -> Result<bool, EspError> {
```
owenthewizard, line 1714: As above.

src/wifi.rs
```diff
@@ -1805,7 +1806,7 @@ impl<'d> NonBlocking for EspWifi<'d> {
         EspWifi::stop_wps(self)
     }
 
-    fn is_wps_finished(&self) -> Result<bool, EspError> {
```
owenthewizard, line 1808: As above.

src/eth.rs
```diff
@@ -687,11 +685,11 @@ impl<'d, T> EthDriver<'d, T> {
         Ok(())
     }
 
-    pub fn set_rx_callback<F>(&mut self, callback: F) -> Result<(), EspError>
```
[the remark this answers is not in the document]
ivmarkov: Yes please.

src/eth.rs
ivmarkov: Revert.

ivmarkov (changes requested):
As mentioned, please revert the signatures of all public methods. The rest I'll accept and merge, though not sure what it gives us, as I think Clippy was only triggering on the `transmute` calls?

Last but not least and very important: I really do hope you've implemented these changes automatically with `cargo clippy --fix` or similar, as so.e of these are not so trivial, and if these were done manually, they could I trounce bugs, as I don't th8nk you've tested all changed code paths. Can you confirm that you've done these changes in an automatic fashion, and how exactly?

owenthewizard:
> Can you confirm that you've done these changes in an automatic fashion, and how exactly?

Unfortunately not as `clippy --fix` only works with a small subset of lints. Additionally it can't reason about `as *const _ as *mut _` -> `.cast().cast_mut()` without type hints. If you don't want to merge the changes due to this I completely understand. If you like I could submit multiple smaller PRs so the changes can be more easily reviewed.

Would you consider removing the unnecessary `Result` on the `EthDriver` callbacks in the next major/API update?

ivmarkov:
> > Can you confirm that you've done these changes in an automatic fashion, and how exactly?
> 
> Unfortunately not as `clippy --fix` only works with a small subset of lints. Additionally it can't reason about `as *const _ as *mut _` -> `.cast().cast_mut()` without type hints. If you don't want to merge the changes due to this I completely understand. If you like I could submit multiple smaller PRs so the changes can be more easily reviewed.

Splitting into multiple smaller PRs would help then. 

> Would you consider removing the unnecessary `Result` on the `EthDriver` callbacks in the next major/API update?

Not sure. Why do you need that so much?

owenthewizard:
> Not sure. Why do you need that so much?

I don't need it per se, I just think it's bloat since it always returns `Ok(())`.

ivmarkov:
> > Not sure. Why do you need that so much?
> 
> I don't need it per se, I just think it's bloat since it always returns `Ok(())`.

I think this is insignificant in the grant scheme of things. And it is not worth it to break backwards compatibility - even for a new API - because of this. Not to mention that in future we might need a result-returning API after all.

owenthewizard:
Closing this, will split into smaller PRs.

PR #414, opened 2024-04-16
Fix missing error return

owenthewizard:
This looks like an oversight.
Too bad the C bindings aren't `#[must_use]` :sweat_smile: .


ivmarkov:
@owenthewizard if you haven't started yet with your casts rework, how about we fix the `transmute` problem only? Or at least fix it first, as it is the most pressing, as it breaks CI. The other cleanup is nice to have, but not really urgent, I guess.

PR #415, opened 2024-04-16
clippy: Add type annotations to `transmute` calls

owenthewizard:
As requested. The lint is currently only in nightly clippy.


ivmarkov:
Thanks! Do you plan to, ahem, fix the same issue in `esp-idf-hal`? :)

owenthewizard:
> Thanks! Do you plan to, ahem, fix the same issue in `esp-idf-hal`? :)

https://github.com/esp-rs/esp-idf-hal/pull/406

PR #416, opened 2024-04-18
ESP-IDF 5.2

owenthewizard:
ESP-IDF 5.2 adds a new field to the SPI ethernet configurations:

> Added option to use SPI Ethernet modules in poll mode without interrupt. [#12682](https://github.com/espressif/esp-idf/issues/12682) ([34ec96e](https://github.com/espressif/esp-idf/commit/34ec96ef23f701a9f72bb988b9195e959e97804d))

So far I have simply added the field and it passes `cargo check`, but there might be some further changes required?

### DIFF
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -470,6 +470,11 @@ where
                         ),
                     )))]
                     custom_spi_driver: eth_spi_custom_driver_config_t::default(),
+                    #[cfg(all(
+                        esp_idf_version_major = "5",
+                        not(any(esp_idf_version_minor = "0", esp_idf_version_major = "1"))
+                    ))]
+                    poll_period_ms: 0,
                 };
 
                 let mac = unsafe { esp_eth_mac_new_dm9051(&dm9051_cfg, &mac_cfg) };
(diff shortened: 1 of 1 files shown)


ivmarkov:
CI fails though?

ivmarkov (changes requested):

owenthewizard:
> CI fails though?

It seems to me this is because esp-idf-hal needs to be updated as well.

ivmarkov:
> > CI fails though?
> 
> It seems to me this is because esp-idf-hal needs to be updated as well.

No. The problem is in `eth.rs` and seems to come from your changes.

owenthewizard:
> > > CI fails though?
> > 
> > 
> > It seems to me this is because esp-idf-hal needs to be updated as well.
> 
> No. The problem is in `eth.rs` and seems to come from your changes.

I can't test the RISC-V targets:

```rust
error[E0463]: can't find crate for `core`
  |
  = note: the `riscv32imc-esp-espidf` target may not be installed
  = help: consider downloading the target with `rustup target add riscv32imc-esp-espidf`
```

ivmarkov:
>   = help: consider downloading the target with `rustup target add riscv32imc-esp-espidf`

Did you try what the compiler suggested?

owenthewizard:
> `rustup target add riscv32imc-esp-espidf`

Of course...

> error: toolchain 'esp' does not support components
> error: toolchain 'nightly-x86_64-unknown-linux-gnu' does not support target 'riscv32imc-esp-espidf'

ivmarkov:
I think you are NOT really compiling with `cargo build -Zbuild-std ...`, as you should

Also can you do:
```sh
$ rustc --print target-list | grep espidf
```

It should print something like:
```
riscv32imac-esp-espidf
riscv32imafc-esp-espidf
riscv32imc-esp-espidf
```

ivmarkov:
This is now addressed in `master`.

PR #418, opened 2024-04-22
Allow disabling IPv6

DavidVentura:
This change allows disabling IPv6 - the type is defined as a union only when IPv4 and IPv6 are enabled [here](https://github.com/espressif/esp-lwip/blob/a45be9e438f6cf9c54ec150581819c3b95d5af6b/src/include/lwip/ip_addr.h#L69)

I'm not sure if it should be patched in this repo, or at bindgen time, let me know

### DIFF
--- a/src/ping.rs
+++ b/src/ping.rs
@@ -99,6 +99,16 @@ impl EspPing {
         conf: &Configuration,
         tracker: &mut Tracker<F>,
     ) -> Result<(), EspError> {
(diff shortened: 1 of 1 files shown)


ivmarkov:
CI fails though?

DavidVentura:
My bad, forgot a cfg guard

ivmarkov:
Thanks!

Issue #420, opened 2024-04-22
Feature request : add wrapper for SD/SDIO/SDMMC driver and vFS 


AlixANNERAUD:

Vollbrecht:
Thanks @AlixANNERAUD for not only requesting it but also delivering it!

PR #422, opened 2024-04-23
Add wrapper for SD/SDIO/SDMMC driver and vFS 

AlixANNERAUD:
Hi,
According to issue #420 .
Am I heading in the right direction ?


Vollbrecht (changes requested):
Thanks for the PR! 

> Am I heading in the right direction ?

You did a great start we surly will get there!

For the SDMMC support we need to introduce some gated compilation. To my understanding, currently -  SDMMC host hardware is only present in esp32 / esp32s3. Other devices such as a esp32c6 uses the SDIO api. The esp-idf docu is a bit confusing on this point. For a first hind you can look at the failed CI runs, that complain that for the specific risc targets the API is not there.

We could feature gate by esp variants directly or use some of the down handed information from esp-idf like SOC_SDMMC_HOST_SUPPORTED flag. 

If you first only want to work on the sdmmc part is fine, but it would be extra nice if you could use the sdspi_host api as a fallback for targets that does not support sdmmc. In concrete i talk about this [sdmmc api ](https://github.com/espressif/esp-idf/blob/v5.2.1/components/driver/sdmmc/include/driver/sdmmc_host.h)vs this [sdspi api](https://github.com/espressif/esp-idf/blob/v5.2.1/components/driver/spi/include/driver/sdspi_host.h)

In either case we would gate the sdmmc api in the sd part so we dont brake targets that dosn't have it.

src/sd/spi.rs
```diff
@@ -0,0 +1,174 @@
...
+            gpio_cs: gpio_num_t_GPIO_NUM_13,
+            gpio_cd: gpio_num_t_GPIO_NUM_NC,
+            gpio_wp: gpio_num_t_GPIO_NUM_NC,
+            gpio_int: gpio_num_t_GPIO_NUM_NC,
+        })
+    }
+}
+
+impl SpiDeviceBuilder {
+    pub fn set_spi<T: spi::Spi>(&mut self, _: &mut T) -> &mut Self {
```
Vollbrecht, line 28: i think we can make this more explicit to be usable with SPI2 ( and SPI3 where supported) by replacing spi::SPI with the [SpiAnyPins](https://github.com/esp-rs/esp-idf-hal/blob/master/src/spi.rs#L73) trait.

A second addition would be to also take spi peripheral as an argument from our Peripherals struct e.g ` _spi: impl Peripheral<P = SPI> + 'd` lile we do in the spi driver, so a user cannot instantiate it two times.

AlixANNERAUD: Yes, that makes sense. I've attempted to fix the issue, but I'm not sure if it's the best solution. I will check tomorrow for gated compilation. Thank your advice.

Vollbrecht:
can you have a look at the `clippy` output from the CI runs. If you have a particular question around something feel free to ask !

AlixANNERAUD:
> can you have a look at the `clippy` output from the CI runs. If you have a particular question around something feel free to ask !

Well, I'm having trouble finding documentation on how the cfg flags of target capabilities are defined. It seems they are exported from ESP IDF `#define` macros, but I can't figure out which one corresponds to the Rust equivalent of [SOC_SDMMC_HOST_SUPPORTED](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/soc_caps.html) (like `esp_idf_bt_classic_enabled` is the equivalent of `SOC_BT_CLASSIC_SUPPORTED`).

AlixANNERAUD:
What is going on with the CI ? There's an issue with ldproxy installation.

Vollbrecht:
> What is going on with the CI ? There's an issue with ldproxy installation.

wired unrelated install error / i restarted the workflow

Vollbrecht:
> > can you have a look at the `clippy` output from the CI runs. If you have a particular question around something feel free to ask !
> 
> Well, I'm having trouble finding documentation on how the cfg flags of target capabilities are defined. It seems they are exported from ESP IDF `#define` macros, but I can't figure out which one corresponds to the Rust equivalent of [SOC_SDMMC_HOST_SUPPORTED](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/soc_caps.html) (like `esp_idf_bt_classic_enabled` is the equivalent of `SOC_BT_CLASSIC_SUPPORTED`). I'm currently using `esp_idf_sdmmc_host_enabled`, but I'm not convinced it's the correct one.

The correct one should be something like  `esp_idf_soc_sdmmc_host_supported` . You can always double check with running  `cargo rustc -- --print cfg`. This will print out all flags that are available on your current used target.

Regarding the other CI error with respect to "use alloc::ffi::CString;" : We also use alloc flags for stuff that uses heap allocation - this is done so that the hal can technically build on no_std. Its mostly used as a dev helper to mark things explicitly for us where heap allocs occure on our site. You would also need gates here.

src/sd/spi.rs
```diff
@@ -0,0 +1,179 @@
...
+
+impl Default for SpiDeviceBuilder {
+    fn default() -> Self {
+        Self(sdspi_device_config_t {
+            host_id: spi_host_device_t_SPI2_HOST,
+            gpio_cs: gpio_num_t_GPIO_NUM_13,
+            gpio_cd: gpio_num_t_GPIO_NUM_NC,
+            gpio_wp: gpio_num_t_GPIO_NUM_NC,
+            gpio_int: gpio_num_t_GPIO_NUM_NC,
+            #[cfg(esp_idf_version = "5.2")]
```
Vollbrecht, line 23: this would only work for a specific 5.2 version. So a 5.3 and onwards would break it. 

One way is to negate the logic and say every version that is not 4 & 5.0 & 5.1 will work, that way its future proof. 

```rust
 #[cfg(not(any(
    esp_idf_version_major = "4",
    all(esp_idf_version_major = "5", esp_idf_version_minor = "0"),
    all(esp_idf_version_major = "5", esp_idf_version_minor = "1"),
 )))]
 
 ```
 
Since we only can use any(), all(), or not() and no greater or less that's the best we got ;D

**Comment by ivmarkov** [comment]
@Vollbrecht @AlixANNERAUD Once you feel this PR is mostly complete, would you let me know? I would also like to do a review w.r.t. naming, code conventions followed in the crate etc.

Please don't expect any large redesign requests, but stuff like:
* Renaming structs
* Putting `const new` here and there
* Putting `Debug`, `Send` etc. here and there

File: src/sd/host.rs [hunk 3]
```diff
@@ -0,0 +1,102 @@
...
+            set_card_clk: Some(sdspi_host_set_card_clk),
+            set_cclk_always_on: None,
+            do_transaction: Some(sdspi_host_do_transaction),
+            __bindgen_anon_1: sdmmc_host_t__bindgen_ty_1 {
+                deinit_p: Some(sdspi_host_remove_device),
+            },
+            io_int_enable: Some(sdspi_host_io_int_enable),
+            io_int_wait: Some(sdspi_host_io_int_wait),
+            get_real_freq: Some(sdspi_host_get_real_freq),
+            #[cfg(esp_idf_version = "5.2")]
```
Vollbrecht, line 42: future proof by negating logic, i will give an example down below.

ivmarkov:
One thing which is a tad more obtrusive and we might have to change is the whole `Builder` pattern. We just don't use it. What we have instead (primarily in `esp-idf-hal`) is the notion of `Config` (which IS actually a builder, but does not contain peripherals) + all peripherals directly passed to a `new` constructor at driver construction time.

But maybe let's solve the other issues first and leave the renames / consistency with the other code for the end. Up to you actually how you would like to tackle this.

AlixANNERAUD:
> @Vollbrecht @AlixANNERAUD Once you feel this PR is mostly complete, would you let me know? I would also like to do a review w.r.t. naming, code conventions followed in the crate etc.
> 
> Please don't expect any large redesign requests, but stuff like:
> 
>     * Renaming structs
> 
>     * Putting `const new` here and there
> 
>     * Putting `Debug`, `Send` etc. here and there

Ok, I'm currently fixing the CI

src/sd/spi.rs
AlixANNERAUD: Alright, it is somewhat bulky indeed, but thank you nonetheless ^^.

AlixANNERAUD:
> What we have instead (primarily in `esp-idf-hal`) is the notion of `Config` (which IS actually a builder, but does not contain peripherals) + all peripherals directly passed to a `new` constructor at driver construction time.
> 
> But maybe let's solve the other issues first and leave the renames / consistency with the other code for the end. Up to you actually how you would like to tackle this.

Okay, that makes sense for reducing the memory footprint. Do you have an example of that? It seems feasible to me.

AlixANNERAUD:
> @Vollbrecht @AlixANNERAUD Once you feel this PR is mostly complete, would you let me know? I would also like to do a review w.r.t. naming, code conventions followed in the crate etc.
> 
> Please don't expect any large redesign requests, but stuff like:
> 
>     * Renaming structs
> 
>     * Putting `const new` here and there
> 
>     * Putting `Debug`, `Send` etc. here and there

I think you can start the review.

ivmarkov:
> > What we have instead (primarily in `esp-idf-hal`) is the notion of `Config` (which IS actually a builder, but does not contain peripherals) + all peripherals directly passed to a `new` constructor at driver construction time.
> > But maybe let's solve the other issues first and leave the renames / consistency with the other code for the end. Up to you actually how you would like to tackle this.
> 
> Okay, that makes sense for reducing the memory footprint. Do you have an example of that? It seems feasible to me.

It is not so much for memory footprint, as it is for consistency.

As for an example - every single driver in `esp-idf-hal`. Take `CanDriver` or any other.

ivmarkov:
There are other issues that need addressing, like the fact that `SpiDeviceBuilder` needs to be lifetimed, as in `SpiDeviceBuilder<'d>` but we'll get there when we get there. :)

src/sd/spi.rs
```diff
@@ -0,0 +1,183 @@
...
+impl From<sdspi_device_config_t> for SpiDeviceBuilder {
+    fn from(config: sdspi_device_config_t) -> Self {
+        Self(config)
+    }
+}
+
+impl Default for SpiDeviceBuilder {
+    fn default() -> Self {
+        Self(sdspi_device_config_t {
+            host_id: spi_host_device_t_SPI2_HOST,
```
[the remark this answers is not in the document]
AlixANNERAUD: Ok, that makes sense.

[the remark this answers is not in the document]
ivmarkov: > Regarding the other pins (cd, wp, int) - can the driver work without these, or are these mandatory?

If you can address this too, as it is a bit important...

AlixANNERAUD:
Has the binding between rust's `std::fs` and esp-idf `vfs` been done?

ivmarkov:
@AlixANNERAUD Not yet. Sorry for being slow with my feedback. Will return some more today.

src/sd/mmc.rs
```diff
@@ -0,0 +1,191 @@
...
+        self.configuration.__bindgen_anon_1.gpio_cd = pin.into_ref().deref().pin();
+        self
+    }
+
+    pub fn set_write_protect_pin(mut self, pin: impl Peripheral<P = impl InputPin> + 'd) -> Self {
+        self.configuration.__bindgen_anon_2.gpio_wp = pin.into_ref().deref().pin();
+        self
+    }
+
+    pub fn get_inner(&self) -> &sdmmc_slot_config_t {
```
ivmarkov, line 184: Why pub?

src/sd/mmc.rs
```diff
@@ -0,0 +1,191 @@
...
+        &self.configuration
+    }
+
+    pub fn get_slot(&self) -> Slot {
```
ivmarkov, line 188: Ditto?

src/sd/mmc.rs
```diff
@@ -0,0 +1,191 @@
...
+    _p: PhantomData<&'d mut ()>,
+}
+
+/// Indicates that card detect line is not used
+const SDMMC_SLOT_NO_CD: i32 = -1;
+/// Indicates that write protect line is not used
+const SDMMC_SLOT_NO_WP: i32 = -1;
+
+#[derive(Clone, Copy, Debug)]
+pub enum Slot {
```
ivmarkov, line 25: Why pub?

src/fs/fat.rs
```diff
@@ -0,0 +1,102 @@
...
+            SdDevice::Mmc(slot_configuration) => unsafe {
+                esp_vfs_fat_sdmmc_mount(
+                    builder.base_path.as_ptr(),
+                    host.get_inner_handle() as *const sdmmc_host_t,
+                    slot_configuration.get_inner() as *const sdmmc_slot_config_t as *const c_void,
+                    &builder.mount_config as *const esp_vfs_fat_mount_config_t,
+                    &mut card as *mut *mut sdmmc_card_t,
+                )
+            },
+            SdDevice::Spi(spi_device) => unsafe {
```
ivmarkov, line 89: See my other comment for using `esp!` instead of `check_and_return`

src/fs/fat.rs
```diff
@@ -0,0 +1,102 @@
...
+use crate::sd::host::SdDevice;
+use crate::{sd::host::SdHost, sys::*};
+
+pub struct FatBuilder {
+    mount_config: esp_vfs_fat_mount_config_t,
+    base_path: CString,
+}
+
+impl Default for FatBuilder {
+    fn default() -> Self {
```
ivmarkov, line 15: Let's get away from the `builder` pattern, as it is not used anywhere else in the `esp-idf-*` crates.
My suggestion:
- create a `config` public module
- Create (inside that module) a `Config` struct
- Put `max_files`, `format_if_mount_failed`, `allocation_unit_size`, etc., **except** `base_path` (see below why) in the `Config` struct
- Implement sane defaults for all parameters of the `Config` struct and besides arg-less `new`, implement `Default` for it. Thus, the user can just pass `&Config::new()` or `&Default::default()` if she does not care
- Please look at how these `Config` structs are used throughout `esp-idf-hal`
- Create a `new` constructor on the `Fat` structure which takes:
  * `SdHost`
  * `&config::Config` (our new structure)
  * `base_path` (for mounting) - this we should keep outside of `Config` as it does not have a sane default

src/sd/host.rs
```diff
@@ -0,0 +1,169 @@
...
+    #[cfg(not(any(
+        esp_idf_version_major = "4",
+        all(esp_idf_version_major = "5", esp_idf_version_minor = "0"),
+        all(esp_idf_version_major = "5", esp_idf_version_minor = "1"),
+    )))] // For ESP-IDF v5.2 and later
+    pub fn set_input_delay_phase(&mut self, phase: DelayPhase) {
+        self.host.input_delay_phase = phase as sdmmc_delay_phase_t;
+    }
+
+    pub fn get_device(&self) -> &SdDevice {
```
ivmarkov, line 166: pub? Why?

src/sd/host.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub fn get_inner_handle(&self) -> &sdmmc_host_t {
+        &self.host
+    }
+
+    #[cfg(not(any(
+        esp_idf_version_major = "4",
+        all(esp_idf_version_major = "5", esp_idf_version_minor = "0"),
+        all(esp_idf_version_major = "5", esp_idf_version_minor = "1"),
+    )))] // For ESP-IDF v5.2 and later
+    pub fn set_input_delay_phase(&mut self, phase: DelayPhase) {
```
ivmarkov, line 162: Belongs to a `Config` struct. See my other comments for the general pattern w.r.t. `Config`.

src/sd/host.rs
```diff
@@ -0,0 +1,169 @@
...
+        self.host.command_timeout_ms = timeout as i32;
+        self
+    }
+
+    pub fn set_io_voltage(mut self, voltage: f32) -> Self {
+        self.host.io_voltage = voltage;
+        self
+    }
+
+    pub fn set_speed(mut self, high_speed: bool) -> Self {
```
ivmarkov, line 143: Belongs to a `Config` struct. See my other comments for the general pattern w.r.t. `Config`.

src/sd/host.rs
```diff
@@ -0,0 +1,169 @@
...
+            host,
+        }
+    }
+
+    pub fn set_command_timeout(mut self, timeout: u32) -> Self {
+        self.host.command_timeout_ms = timeout as i32;
+        self
+    }
+
+    pub fn set_io_voltage(mut self, voltage: f32) -> Self {
```
ivmarkov, line 138: Belongs to a `Config` struct. See my other comments for the general pattern w.r.t. `Config`.

src/sd/host.rs
```diff
@@ -0,0 +1,169 @@
...
+            command_timeout_ms: 0,
+        };
+
+        Self {
+            device: SdDevice::Mmc(configuration),
+            host,
+        }
+    }
+
+    pub fn set_command_timeout(mut self, timeout: u32) -> Self {
```
ivmarkov, line 133: Belongs to a `Config` struct. See my other comments for the general pattern w.r.t. `Config`.

ivmarkov (changes requested):
Sorry for the delay! Holidays over here. :)

src/sd/spi.rs
```diff
@@ -0,0 +1,129 @@
...
+        let result = unsafe { sdspi_host_remove_device(self.handle) };
+
+        if result != ESP_OK {
+            panic!("Failed to remove SPI device");
+        }
+    }
+}
+
+impl<'d> SpiDevice<'d> {
+    pub fn initialize_host() -> Result<(), EspError> {
```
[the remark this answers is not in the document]
AlixANNERAUD: Certainly, we can achieve that, but it requires using a static mutable boolean as an initialization flag, as the 'init' function cannot be invoked more than once.

src/sd/host.rs
```diff
@@ -0,0 +1,169 @@
...
+    device: SdDevice<'d>,
+    host: sdmmc_host_t,
+}
+
+#[cfg(not(any(
+    esp_idf_version_major = "4",
+    all(esp_idf_version_major = "5", esp_idf_version_minor = "0"),
+    all(esp_idf_version_major = "5", esp_idf_version_minor = "1"),
+)))] // For ESP-IDF v5.2 and later
+pub enum DelayPhase {
```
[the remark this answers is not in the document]
AlixANNERAUD: Ok but why deriving `Default` ?

src/sd/mmc.rs
```diff
@@ -0,0 +1,191 @@
...
+                8
+            } else {
+                4
+            }
+        } else {
+            1
+        }
+    }
+
+    pub fn set_card_detect_pin(mut self, pin: impl Peripheral<P = impl InputPin> + 'd) -> Self {
```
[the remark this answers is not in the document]
AlixANNERAUD: Since it's optional, and considering the constructor is currently quite extensive with 10 arguments, I thought it would be better to leave it here.

src/sd/spi.rs
```diff
@@ -0,0 +1,129 @@
...
+        let result = unsafe { sdspi_host_deinit() };
+
+        if result == ESP_OK {
+            Ok(())
+        } else {
+            Err(result)
+        }
+    }
+
+    pub fn get_device_configuration(&self) -> &sdspi_device_config_t {
```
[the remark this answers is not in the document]
AlixANNERAUD: Since the FAT system requires SPI device configuration to mount the card.

src/sd/mmc.rs
AlixANNERAUD: Since the FAT (esp_vfs_fat_sdspi_mount) system requires MMC device configuration to mount the card.

src/sd/mmc.rs
AlixANNERAUD: Since SdHost (`sdmmc_host_t`) requires slot of MMC device configuration.

src/sd/mmc.rs
AlixANNERAUD: Same thing as above.

src/sd/mmc.rs
ivmarkov: `pub(crate)` then please.

src/sd/mmc.rs
ivmarkov: See above.

examples/sd_spi.rs
```diff
@@ -0,0 +1,63 @@
+#[cfg(esp32)]
+fn main() -> anyhow::Result<()> {
+    use esp_idf_hal::{
```
ivmarkov, line 3: Nit: `esp_idf_svc::hal` instead. This way we hint the user that she only needs to depend on `esp_idf_svc`, as it re-exports the hal and sys crates anyway.

examples/sd_mmc.rs
```diff
@@ -0,0 +1,57 @@
+#[cfg(all(esp32, esp_idf_soc_sdmmc_host_supported))]
+fn main() -> anyhow::Result<()> {
+    use esp_idf_hal::{
```
ivmarkov, line 3: See my other comment w.r.t. `esp_idf_hal`.

src/fs/config.rs
```diff
@@ -0,0 +1,43 @@
...
+            #[cfg(not(
+                esp_idf_version_major = "4",
+            ))] // For ESP-IDF v5.0 and later
+            disk_status_check_enable: false,
+        }
+    }
+}
+
+impl Configuration {
+    pub fn new() -> Self {
```
ivmarkov, line 26: Make this `const fn` and move the implementation _here_. Then have `Default::default` delegate to the `const fn` constructor. This way the user can construct configuration in a co st context too, if she wants to.

src/sd/config.rs
```diff
@@ -0,0 +1,66 @@
...
+                all(esp_idf_version_major = "5", esp_idf_version_minor = "0"),
+                all(esp_idf_version_major = "5", esp_idf_version_minor = "1"),
+            )))] // For ESP-IDF v5.2 and later
+            input_delay_phase: DelayPhase::Phase0,
+        }
+    }
+}
+
+impl Configuration {
+    pub fn new() -> Self {
```
ivmarkov, line 63: See my other comment for `const fn new`.

ivmarkov:
@AlixANNERAUD I'll merge this shortly, but _might_ do a round of final renames/changes post commit - hope you don't mind. For one, this code is not `#[cfg()]`-ed yet on the VFS and FAT components being enabled, so it will fail to compile when ESP IDF is compiled without support for VFS.

ivmarkov:
@AlixANNERAUD I'll merge this shortly, but _might_ do a round of final renames/changes post commit - hope you don't mind. For one, this code is not `#[cfg()]`-ed yet on the VFS and FAT components being enabled, so it will fail to compile when ESP IDF is compiled without support for VFS.

AlixANNERAUD:
> @AlixANNERAUD I'll merge this shortly, but _might_ do a round of final renames/changes post commit - hope you don't mind. For one, this code is not `#[cfg()]`-ed yet on the VFS and FAT components being enabled, so it will fail to compile when ESP IDF is compiled without support for VFS.

Thank you very much for your help, it's really kind of you to have taken the time to assist me. I've added the missing `cfg for vfs/fat.

Vollbrecht:
> > @AlixANNERAUD I'll merge this shortly, but _might_ do a round of final renames/changes post commit - hope you don't mind. For one, this code is not `#[cfg()]`-ed yet on the VFS and FAT components being enabled, so it will fail to compile when ESP IDF is compiled without support for VFS.
> 
> Thank you very much for your help, it's really kind of you to have taken the time to assist me. I've added the missing `cfg` for vfs/fat.

Thanks for pushing the PR through!

niuhuan:
my tf card init faild on esp32-s3 use this example .

https://github.com/esp-rs/esp-idf-svc/issues/439

PR #427, opened 2024-05-11
Exposes ctrl_port in the configuration of an HttpServer

Angelo13C:
Hi!
I am making an IP camera using an ESP32CAM, and I need to have 2 HTTP servers. The first one handles most of the requests (like index...) but then I need a second one which only handles the `/stream` which is going to have a `loop` inside that continuously sends the images to the HTTP client (check this: https://stackoverflow.com/questions/47729941/mjpeg-over-http-specification).

If I only have 1 server as soon as I receive a `/stream` request everything is blocked (because of the `loop`).
The problem now is that since `ctrl_port` is hidden from Rust, I can't start a second server since the `ctrl_port` is already occupied by the first one.

### DIFF
--- a/src/http/server.rs
+++ b/src/http/server.rs
@@ -77,6 +77,7 @@ pub use super::*;
 #[derive(Copy, Clone, Debug)]
 pub struct Configuration {
     pub http_port: u16,
+    pub ctrl_port: u16,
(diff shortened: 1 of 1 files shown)


Vollbrecht:
LGTM!

PR #428, opened 2024-05-16
Support extracting the protocol chosen with ALPN.

jothan:
The additional handshake info is returned by the negotiate() methods. Doing it this way should incur no space overhead for users that simply discard the return value.


ivmarkov:
Thanks!

PR #429, opened 2024-05-16
Make EspTls and EspAsyncTls Send when possible

jothan:
This is required, for example, by the Hyper HTTP2 client implementation.

### DIFF
--- a/src/tls.rs
+++ b/src/tls.rs
@@ -723,7 +723,7 @@ mod esptls {
         not(esp_idf_version_major = "4"),
         any(not(esp_idf_version_major = "5"), not(esp_idf_version_minor = "0"))
     ))]
-    pub struct EspAsyncTls<S>(core::cell::RefCell<EspTls<S>>)
+    pub struct EspAsyncTls<S>(EspTls<S>)
     where
         S: PollableSocket;
 
@@ -742,7 +742,7 @@ mod esptls {
         ///
         /// * `ESP_ERR_NO_MEM` if not enough memory to create the TLS connection
         pub fn adopt(socket: S) -> Result<Self, EspError> {
-            Ok(Self(core::cell::RefCell::new(EspTls::adopt(socket)?)))
+            Ok(Self(EspTls::adopt(socket)?))
         }
 
         /// Establish a TLS/SSL connection using the adopted socket.
@@ -774,10 +774,7 @@ mod esptls {
             rcfg.non_block = false;
 
             let res = loop {
-                let res = self
-                    .0
-                    .borrow_mut()
-                    .internal_connect(hostname, 0, true, &rcfg);
+                let res = self.0.internal_connect(hostname, 0, true, &rcfg);
 
                 match res {
                     Err(e) => self.wait(e).await?,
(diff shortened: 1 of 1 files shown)


ivmarkov:
Removal of `RefCell` is not ideal though. Did you need that for `Send`?

Without `read(&self)` and `write(&self)` you can't easily implement full duplex read+write (i.e. TLS socket splitting), and that's a necessity for some protocols (e.g. websockets).

jothan:
Oh, right, I forgot that RefCell is Send.

I don't mind putting the RefCell back in for that case.

However, an external wrapper could provide the same thing, such as:

```rust
struct ReadHalf<'a, S>(&'a RefCell<EspAsyncTls<S>>);
struct ReadHalf<S>(Rc<RefCell<EspAsyncTls<S>>>);
struct ReadHalf<S>(Arc<Mutex<EspAsyncTls<S>>>);
```

Let me know what you think.

ivmarkov:
> Oh, right, I forgot that RefCell is Send.
> 
> I don't mind putting the RefCell back in for that case.
> 
> However, an external wrapper could provide the same thing, such as:
> 
> ```rust
> struct ReadHalf<'a, S>(&'a RefCell<EspAsyncTls<S>>);
> struct ReadHalf<S>(Rc<RefCell<EspAsyncTls<S>>>);
> struct ReadHalf<S>(Arc<Mutex<EspAsyncTls<S>>>);
> ```

Right, but then you also need to re-implement `async fn read(&mut self, ...)` for `ReadHalf` by re-doing the `poll_fn` dance and calling `poll_read` on the `EspAsyncTls` contained inside the `RefCell` / `Rc<RefCell>` / `Arc<Mutex>` wrapper.

Ditto for `WriteHalf` and `write`.

We can go either way - either restoring the `RefCell`, or providing some variant of the wrappers (`ReadHalf` and `WriteHalf`) and then - if the provided variant is using regular `&'a` ref + `RefCell` (which is what I would suggest) - maybe also `AsyncTls::split(&mut self) -> (ReadHalf<'a, S>, WriteHalf<'a. S>)`?

But the point is, I don't think we can just remove the splitting functionality and claim the users can "just do it" (as in, "look: you have `poll_read`/`poll_write` already, what else do you want??") simply because a lot of the folks don't have that deep understanding you are showing here.

So I'm not sure which way we should go (w.r.t. restoring some sort of splitting).
On one hand, extra wrappers look nice(r) and you get the extra benefit that (advanced) users can copy-paste and adjust them if - say - then need a `'static` variant or a `'static + Sync` variant (your `Rc` and `Arc + Mutex` variations respectively).

On the other hand, I got some slapping already from users that we are changing the API way too often for only "theoretical" reasons without practical benefit, so I'm not sure which way to go either. Even if this is relatively new code.

What is your gut feeling?

ivmarkov:
Updated the above comment, as I had forgotten that we actually have something called `wait` instead of the usual `poll_read` / `poll_write`, but that doesn't change the problem or the arguments.

P.S. Wondering whether we should expose the traditional `poll_read` / `poll_write` instead, and whether I'm not having a bug actually in-there getting-by with a single `wait`...?

jothan:
I did a bit of poll_read/write work here, not PR quality though: https://github.com/esp-rs/esp-idf-svc/compare/master...jothan:esp-idf-svc:poll_wait

I'd be in favor of adding wrappers for splitting (maybe with a single and multi-threaded variant), but as you said, I don't know how much we want to annoy the current API users.

ivmarkov:
Looking again at it:
* `wait` is currently not even `pub`, so it needs to become `pub`
* Better to have `poll_read` / `poll_write` as `pub` though, as I'm in the meantime pretty sure there is a bug in there with the single `wait`. Why should a `write` operation wait for the socket to become readable?

> I did a bit of poll_read/write work here, not PR quality though: [master...jothan:esp-idf-svc:poll_wait](https://github.com/esp-rs/esp-idf-svc/compare/master...jothan:esp-idf-svc:poll_wait)
> 
> I'd be in favor of adding wrappers for splitting (maybe with a single and multi-threaded variant), but as you said, I don't know how much we want to annoy the current API users.

Hmmm, the more I look at the existing code, the more I'm not having a warm fuzzy feeling. It might have bugs, but then if we go your `poll_read`/`poll_write` direction, we still have to check it works, as it might make an existing problem even worse.

Do you have a use case for socket splitting right now? 
* If yes - then I say you can do it your way and then naturally you can test it too.
* If not - then let's leave the `RefCell` in-place? I'll surely have a use case "soon" (as in a few weeks) but it has been "in a few weeks" for the last several months. So when I get there, I'll test the existing impl, and might fix it / change it adjust it.

jothan:
I personally don't have a use case for splitting right now, so I'm going to adjust this PR to put the RefCell back in.

ivmarkov:
Thanks!

PR #431, opened 2024-05-21
🗑️ cleanup deprecated idf 4.3 related code

Vollbrecht:


src/sntp.rs
```diff
@@ -45,31 +44,6 @@ mod esp_sntp {
     pub use esp_sntp_stop as sntp_stop;
 }
 
-#[cfg(any(esp_idf_version_major = "4", esp_idf_version_minor = "0"))]
```
Vollbrecht, line 48: Is this header for version 4.x and x.0 ?? If so i need to revert that back and change it to 5.1 and onwards

Vollbrecht: done

ivmarkov (approved):

PR #437, opened 2024-06-22
Mark all non-static handlers as unsafe

ivmarkov:
This PR "operates" the HTTP server wrapper in a similar way how all other APIs accepting callbacks were altered during the previous release cycle: 
- All APIs which used to take non-static callbacks, now only accept `'static` ones, which is safe
- All of them now also have a mirror `_nonstatic` APIs marked with `unsafe` which continue to accept non-`'static` callbacks, but have big fat `#Safety` strings attached

With this PR, what would happen to the code that was posted in Matrix by a user, is that the code would stop compiling, because the handlers are non-static in that they borrow a ref to a `Mutex` which lives on the stack (and the program crashes because additionally it calls `core::mem::forget` on a value that continues to live after the stack had gone).


Vollbrecht (approved):
Can you also please add a CHANGELOG entry so we don't forget. Otherwise LGTM!

ivmarkov:
> Can you also please add a CHANGELOG entry so we don't forget. Otherwise LGTM!

Sure, done. Merging.

PR #440, opened 2024-06-25
Use ws url based on current page's url

ChocolateLoverRaj:
Instead of using a fixed URL, base the ws url based off of the current http URL. This will make it easier to modify the examples, such as connecting to an existing Wi-Fi network instead of creating one.


ivmarkov:
Thanks!

PR #444, opened 2024-07-06
fix rust-analyzer by not using rust libtest harness

Vollbrecht:


Cargo.toml
```diff
@@ -15,6 +15,10 @@ build = "build.rs"
 documentation = "https://docs.esp-rs.org/esp-idf-svc/"
 rust-version = "1.77"
 
+[lib]
```
ivmarkov, line 18: * Currently does not merge due to updated `Cargo.toml`
* Don't we need this in `esp-idf-svc` and `esp-idf-sys` too?
* Is the `name` a mandatory attribute? If not, why putting it? Also why `esp_idf_svc` and not `esp-idf-svc`?

ivmarkov (changes requested):

Cargo.toml
Vollbrecht: we need it in every crate, though i was doing the updates one by one and just started here.

regarding the name,i thought it was mandatory, as it was the case with [[bin]]. Though if i check [here](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#configuring-a-target) it seams we don't need it. If you pick a name in a [lib] cargo forbids `-` and you can only use `_`

ivmarkov: > regarding the name,i thought it was mandatory, as it was the case with [[bin]]. Though if i check [here](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#configuring-a-target) it seams we don't need it. If you pick a name in a [lib] cargo forbids `-` and you can only use `_`

Agreed. I removed the `name` (seems to work without it), and pushed the change to all crates, so I'll close this.
Once we get confirmation from the poster in `esp-idf-matter` that the example works for him (or if we don't hear from him anything in the next couple of days), I'll release patch updates to both -hal and -svc (with previous -hal yanked, as the ResetReason is - formally speaking - a breaking change in the absence of `#[non_exhaustive]`.

PR #447, opened 2024-07-07
Support for esp32c2, esp32h2, esp32c6 and in future esp32h4 and esp32c5

ivmarkov:
Subject says it all.
BT controller initialization for these chips is different from the currently supported esp32/esp32c3/esp32s3.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,11 +21,12 @@ jobs:
       matrix:
         target:
           - riscv32imc-esp-espidf
+          - riscv32imac-esp-espidf
           - xtensa-esp32-espidf
           - xtensa-esp32s2-espidf
           - xtensa-esp32s3-espidf
         idf-version:
-          - v4.4.6
+          - v4.4.7
           - v5.1.2
           - v5.2
     steps:
@@ -84,6 +85,7 @@ jobs:
           chmod a+x $HOME/.cargo/bin/ldproxy
 
       - name: Build | Examples
+        if: matrix.target != 'riscv32imac-esp-espidf'
         env:
           ESP_IDF_VERSION: ${{ matrix.idf-version }}
           ESP_IDF_SDKCONFIG_DEFAULTS: "${{ github.workspace }}/.github/configs/sdkconfig.defaults"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,9 @@ build = "build.rs"
 documentation = "https://docs.esp-rs.org/esp-idf-svc/"
 rust-version = "1.77"
 
+[patch.crates-io]
+esp-idf-hal = { git = "https://github.com/esp-rs/esp-idf-hal" }
+
 [features]
 default = ["std", "binstart"]
 
--- a/src/bt.rs
+++ b/src/bt.rs
@@ -466,33 +466,33 @@ where
     fn init(_nvs_enabled: bool) -> Result<(), EspError> {
         #[cfg(esp32)]
         let mut bt_cfg = esp_bt_controller_config_t {
-            magic: ESP_BT_CONTROLLER_CONFIG_MAGIC_VAL,
-            controller_task_stack_size: ESP_TASK_BT_CONTROLLER_STACK as _,
-            controller_task_prio: ESP_TASK_BT_CONTROLLER_PRIO as _,
-            hci_uart_no: BT_HCI_UART_NO_DEFAULT as _,
-            hci_uart_baudrate: BT_HCI_UART_BAUDRATE_DEFAULT,
-            scan_duplicate_mode: SCAN_DUPLICATE_MODE as _,
-            scan_duplicate_type: SCAN_DUPLICATE_TYPE_VALUE as _,
-            normal_adv_size: NORMAL_SCAN_DUPLICATE_CACHE_SIZE as _,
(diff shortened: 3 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -21,11 +21,12 @@ jobs:
       matrix:
         target:
           - riscv32imc-esp-espidf
+          - riscv32imac-esp-espidf
           - xtensa-esp32-espidf
           - xtensa-esp32s2-espidf
           - xtensa-esp32s3-espidf
         idf-version:
-          - v4.4.6
+          - v4.4.7
```
Vollbrecht, line 29: v4.4.8 was release just a week ago as the final v4 version

src/bt.rs
```diff
@@ -570,6 +570,74 @@ where
...
+                as _,
+            dis_scan_backoff: crate::sys::NIMBLE_DISABLE_SCAN_BACKOFF as _,
+            ble_scan_classify_filter_enable: 1,
+            main_xtal_freq: crate::sys::CONFIG_XTAL_FREQ as _,
+            #[cfg(esp32c2)]
+            version_num: unsafe { crate::sys::esp_ble_get_chip_rev_version() },
+            #[cfg(esp32c6)]
+            version_num: unsafe { crate::sys::efuse_hal_chip_revision() },
+            #[cfg(not(esp32c2))]
+            cpu_freq_mhz: crate::sys::CONFIG_ESP_DEFAULT_CPU_FREQ_MHZ as _,
```
Vollbrecht, line 627: can c2 work correctly with a default cpu_freq_mhz here? Or do we need to find something that exist on c2 that represent the equivalent of DEFAULT_CPU_FREQ_MHZ?

ivmarkov: The BT controller for esp32c2 does not even have this field.

.github/workflows/ci.yml
ivmarkov: But we are still offering as default 4.4.7 in esp-idf-sys?

Cargo.toml
```diff
@@ -15,6 +15,9 @@ build = "build.rs"
 documentation = "https://docs.esp-rs.org/esp-idf-svc/"
 rust-version = "1.77"
 
+[patch.crates-io]
+esp-idf-hal = { git = "https://github.com/esp-rs/esp-idf-hal", branch = "fix-timer-gpio-v4" }
```
Vollbrecht, line 19: i merged the hal PR so that can be removed

Vollbrecht (changes requested):
Removing the patch section and we are good.

Cargo.toml
ivmarkov: It has to stay. Only the "branch" part can be removed now. :( With a workspace setup, that would've been unnecessary.

Vollbrecht: That's true.

Besides the point but just wanna say, since we had another issue with the added reset reasons that is now [fixed](https://github.com/esp-rs/esp-idf-hal/pull/444/files) in esp-idf-hal we might want to make a quick patch release for esp-idf-hal

ivmarkov: True. I've updated the changelog with both of these. Might release a new patch version shortly even though extending `ResetReason` with additional values is a breaking change. However, I guess we can consider the previous state as broken anyway? In any case we need `#[non_exhaustive]` on most of our `enum`s

ivmarkov: ... which is a breaking change which would allow us to have less breaking changes in future. BTW non-exhaustive we need not just on enums, but also on `structs`. Likely on ALL of these.

Vollbrecht: We would need "non_exhaustive" so we can "marked" it as a non-breaking change, though at the end we don't break things here. It just that esp-idf at the .patch level is adding stuff to C enums that is semi breaking stuff for us here, and a user faces it when he is actively is updating to the latest versions. Also non_exhausitve is not needed since we transpile the enum's from a list of const vars. The problem is that we are calling _ => panic in unknown variants ( which is i think overall the right call)  that is "breaking" stuff.

ivmarkov: > We would need "non_exhaustive" so we can "market" it as a non-breaking change, though at the end we don't break things here. It just that esp-idf at the .patch level is adding stuff to C enums that is semi breaking stuff for us here, and a user faces it when he is actively is updating to the latest ESP-IDF versions. 

ESP-IDF explicitly documents adding new members to a C struct / enum as a "non-breaking" change even if formally speaking it IS a breaking change. They even require a special style of initializing such a struct and switching on such an enum so that the user code does not break. Fortunately, on the Rust side, with `non_exhaustive` - for the **Rust** counterparts of these raw esp-idf-sys types we can enforce it in a clearer way.

> Also non_exhausitve is not needed since we transpile the enum's from a list of const vars. The problem is that we are calling _ => panic in unknown variants ( which is i think overall the right call) that is "breaking" stuff.

How **we** match on the raw `esp_idf_sys` constants / C structs when converting them to actual Rust enums / structs has nothing to do with `non_exhaustive`. `non_exhaustive` addresses how **the user** is matching on our own Rust level structs / enum. Specifically, when destructuring structs he should always use the `..` syntax, and when matching on enums he should always have a "catch all" `_ =>` variant, which then allows **us** to keep adding new values / members to the Rust struct / enum without this being a breaking change for the user anymore.

Issue #439, opened 2024-06-24
sd mmc init faild


niuhuan:
I am truly certain that my hardware is not faulty, and I can use the tf card normally with other firmware.

I have checked the correctness of the PINs, but still cannot read the tfcard

![image](https://github.com/esp-rs/esp-idf-svc/assets/20847533/02de3c2c-ccbd-46df-8d40-fa84d6aeb54e)

My code is copy lot of https://github.com/esp-rs/esp-idf-svc/blob/master/examples/sd_mmc.rs

```rust

    use esp_idf_svc::{
        fs::{Fat, FatConfiguration},
        hal::{
            gpio::{self, AnyIOPin},
            prelude::*,
        },
        sd::{host::SdHost, mmc::SlotConfiguration, SdConfiguration},
    };
    use std::{fs::File, io::Write};

    let slot = SlotConfiguration::new(
        pins.gpio16,
        pins.gpio12,
        pins.gpio14,
        Some(pins.gpio17),
        Some(pins.gpio21),
        Some(pins.gpio18),
        Option::<gpio::Gpio16>::None,
        Option::<gpio::Gpio17>::None,
        Option::<gpio::Gpio15>::None,
        Option::<gpio::Gpio18>::None,
        Option::<AnyIOPin>::None,
        Option::<AnyIOPin>::None,
    );

    log::info!("INIT MMC");

    let host_config = SdConfiguration::new();

    let host = SdHost::new_with_mmc(&host_config, slot);

    let fat_config = FatConfiguration::new();

    let _fat = Fat::mount(fat_config, host, "/")?;

    let mut  read = fs::read_dir("/")?;
    while let Some(entry) = read.next() {
        let entry = entry?;
        let path = entry.path();
        log::info!("File: {:?}", path);
    }

    let mut file = File::create("/test.txt")?;

    file.write_all(b"Hello, world!")?;

```

```
I (434) cpu_start: cpu freq: 160000000 Hz
I (435) cpu_start: Application information:
I (437) cpu_start: Project name:     libespidf
I (443) cpu_start: App version:      1
I (447) cpu_start: Compile time:     Jun 23 2024 21:51:15
I (453) cpu_start: ELF file SHA256:  0000000000000000...
I (459) cpu_start: ESP-IDF:          v5.1.3
I (464) cpu_start: Min chip rev:     v0.0
I (469) cpu_start: Max chip rev:     v0.99
I (473) cpu_start: Chip rev:         v0.2
[...]

PR #448, opened 2024-07-09
Sd card fix

AlixANNERAUD:
According to  #439

### DIFF
--- a/src/sd/host.rs
+++ b/src/sd/host.rs
@@ -1,3 +1,7 @@
+use core::borrow::BorrowMut;
+
+use esp_idf_hal::spi::SpiDriver;
+
 use crate::sys::*;
 
 #[cfg(esp_idf_soc_sdmmc_host_supported)]
@@ -5,14 +9,20 @@ use super::mmc::SlotConfiguration;
 
 use super::{config::Configuration, spi::SpiDevice};
 
-pub enum SdDevice<'d> {
-    Spi(SpiDevice<'d>),
+pub enum SdDevice<'d, T>
+where
+    T: BorrowMut<SpiDriver<'d>>,
+{
+    Spi(SpiDevice<'d, T>),
     #[cfg(esp_idf_soc_sdmmc_host_supported)]
     Mmc(SlotConfiguration<'d>),
 }
 
-pub struct SdHost<'d> {
-    device: SdDevice<'d>,
+pub struct SdHost<'d, T>
+where
+    T: BorrowMut<SpiDriver<'d>>,
+{
+    device: SdDevice<'d, T>,
     host: sdmmc_host_t,
 }
 
@@ -27,7 +37,10 @@ const _SDMMC_HOST_FLAG_4BIT: u32 = 1 << 1;
 const _SDMMC_HOST_FLAG_8BIT: u32 = 1 << 2;
 const _SDMMC_HOST_FLAG_DDR: u32 = 1 << 3;
 
-impl<'d> SdHost<'d> {
+impl<'d, T> SdHost<'d, T>
+where
+    T: BorrowMut<SpiDriver<'d>>,
+{
     fn set_from_configuration(mut self, configuration: &Configuration) -> Self {
         self.host.command_timeout_ms = configuration.command_timeout_ms as i32;
         self.host.io_voltage = configuration.io_voltage;
@@ -49,7 +62,7 @@ impl<'d> SdHost<'d> {
         self
     }
 
-    pub fn new_with_spi(configuration: &Configuration, device: SpiDevice<'d>) -> Self {
+    pub fn new_with_spi(configuration: &Configuration, device: SpiDevice<'d, T>) -> Self {
         let host = sdmmc_host_t {
             flags: _HOST_FLAG_SPI | _HOST_FLAG_DEINIT_ARG,
             slot: device.get_host() as i32,
@@ -190,7 +203,7 @@ impl<'d> SdHost<'d> {
         &self.host
     }
 
-    pub(crate) fn get_device(&self) -> &SdDevice {
+    pub(crate) fn get_device(&self) -> &'d SdDevice<T> {
         &self.device
     }
 }
--- a/src/sd/host.rs
+++ b/src/sd/host.rs
@@ -39,7 +39,7 @@ impl<T> SdHost<T> {
         };
 
         #[cfg(not(any(
-            esp_idf_verdevicesion_major = "4",
+            esp_idf_version_major = "4",
             all(esp_idf_version_major = "5", esp_idf_version_minor = "0"),
             all(esp_idf_version_major = "5", esp_idf_version_minor = "1"),
         )))] // For ESP-IDF v5.2 and later
--- a/src/sd/host.rs
+++ b/src/sd/host.rs
@@ -1,3 +1,7 @@
+use core::borrow::Borrow;
+
+use esp_idf_hal::spi::SpiDriver;
+
 use crate::sys::*;
 
 #[cfg(esp_idf_soc_sdmmc_host_supported)]
@@ -5,14 +9,11 @@ use super::mmc::SlotConfiguration;
 
 use super::{config::Configuration, spi::SpiDevice};
 
-pub enum SdDevice<'d> {
-    Spi(SpiDevice<'d>),
-    #[cfg(esp_idf_soc_sdmmc_host_supported)]
-    Mmc(SlotConfiguration<'d>),
-}
-
-pub struct SdHost<'d> {
-    device: SdDevice<'d>,
+/// SD host.
+///
+/// This struct is an abstraction used by [`crate::fs::Fat`] to mount SD cards containing FAT partition on the VFS.
+pub struct SdHost<T> {
+    device: T,
     host: sdmmc_host_t,
 }
 
@@ -27,7 +28,7 @@ const _SDMMC_HOST_FLAG_4BIT: u32 = 1 << 1;
 const _SDMMC_HOST_FLAG_8BIT: u32 = 1 << 2;
 const _SDMMC_HOST_FLAG_DDR: u32 = 1 << 3;
 
-impl<'d> SdHost<'d> {
+impl<T> SdHost<T> {
     fn set_from_configuration(mut self, configuration: &Configuration) -> Self {
         self.host.command_timeout_ms = configuration.command_timeout_ms as i32;
         self.host.io_voltage = configuration.io_voltage;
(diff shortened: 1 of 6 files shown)


AlixANNERAUD:
The SD SPI part is now functional, it is possible to read/write files without any issues. However, at the end of the program (destruction of _fat), I get the following error: 
```
I (1406) gpio: GPIO[5]| InputEn: 1| OutputEn: 0| OpenDrain: 0| Pullup: 0| Pulldown: 0| Intr:0 
E (1426) spi_master: spi_master_deinit_driver(309): not all CSses freed
I (1436) gpio: GPIO[23]| InputEn: 0| OutputEn: 0| OpenDrain: 0| Pullup: 1| Pulldown: 0| Intr:0 
I (1446) gpio: GPIO[19]| InputEn: 0| OutputEn: 0| OpenDrain: 0| Pullup: 1| Pulldown: 0| Intr:0 
I (1456) gpio: GPIO[18]| InputEn: 0| OutputEn: 0| OpenDrain: 0| Pullup: 1| Pulldown: 0| Intr:0 
```

ivmarkov:
Before committing fixes - which might not improve the situation - I think we have to answer ourselves the following simple question:
- Can you share an SPI bus between an SD card and another SPI slave device?

If the answer is "yes":
- Then you _have to_ pass a `Borrow<SpiDriver>` to your SD-SPI driver, _and you have to keep it in your driver until your driver is dropped_, period. You should also _not_ try to initialize/free the bus yourself, as this is what the `SpiDriver` does. It represents an initialized, shareable SPI **bus**.

2. If the answer is "no" (i.e. an SD-MMC card cannot share the SPI bus with other SPI devices):
- Then you have two options:
  - Option 1: Still use an `SpiDriver` _but then pass it as `BorrowMut<SpiDriver>` as the bus will not be shared
  - Option 2: Don't use an `SpiDriver` and implement your SDMMC driver in such a way that it takes an `SPI` peripheral directly, **as the `SpiDriver` itself does**.

In any case and regarding usage of `SpiDeviceDriver`: `SpiDeviceDriver` models one device potentially out of many that share the same SPI Bus. However, you should be using `SpiDeviceDriver` in your code ONLY if you plan to call its Rust methods (as in read/write etc.). But you don't do that _as you are wrapping a C driver that expects CS and creates its "own" version of `SpiDeviceDriver` in C_.


So if you could start by answering the first question, and then we proceed from there?

ivmarkov:
More clarifications if still not completely clear:
- An SPI bus (`SpiDriver`) represents the "in" and "out" SPI pins, the clock pin and the spi peripheral itself
- An SPI "Device" (`SpiDeviceDriver` or its "I-cannot-share-the-bus" `SpiBusDriver` cousin) represent one SPI device thart wants to use the bus. The CS pin (plus locking the bus) is a way to tell the "salve" SPI listening on the other end (as your SDMMC card) that we are sending data to it, not to some other device on the same bus (CS = "chip-select", i.e. we indicate to which device we want to talk via the shared clock, in and out pins)
  - However, your SDMMC C code already _acts_ as an `SpiDeviceDriver` in that it expects directly a CS pin. So you should not create an `SpiDeviceDriver` at all. 
  - **IF** the SDMMC C code also acts as an SPI Bus which I doubt though (i.e. it takes as input the in, out and clock pins rather than `spi_host_device_t` then you don't need to keep around `SpiDevice` borrow; but I doubt that)

ivmarkov:
Also our naming is a bit weird I admit, maybe we need to change it in future:
- `SpiDriver` - represents the SPI **bus**
- `SpiBusDriver` - represents a single SPI device on the bus that cannot share the bus with other devices (perhaps we should've named this SpiSingleDeviceDriver` or something
- `SpiDeviceDriver` - represents an one out of many SPI devices that can share the bus
- `SpiSoftCsDeviceDriver` - just like `SpiDeviceDriver` but allows you to create more than 6 devices on a single bus by operating the CS pin in software (and paying the price in terms of more latency)

ivmarkov:
Looking at the existing code, I'm getting convinced that it is more correct that what you are trying to introduce in the PR, in that the existing code correctly takes `SpiDriver` instead of `SpiDeviceDriver`. However it still has at least these issues:
- You are [passing `SpiDriver` by moving it into your driver](https://github.com/esp-rs/esp-idf-svc/blob/master/src/sd/spi.rs#L29). You should only do that if SDMMC cannot share the bus (i.e. the question from my first post)
- The current code however frees the bus as in [here](https://github.com/esp-rs/esp-idf-svc/blob/master/src/sd/spi.rs#L19)? This you should absolutely not do as you are double-freeing the bus this way. Remember: your driver does not "own" the bus. `SpiDriver` owns it and the bus will be freed when `SpiDriver` is dropped. Automatically

There might be other issues; I can help identifying those but you have to explain a bit (in English) what you are trying to fix with your new PR. Adding some small description of this PR would definitely help a lot.

src/sd/spi.rs
```diff
@@ -1,32 +1,23 @@
...
-            panic!("Failed to remove SPI device");
-        }
-    }
-}
-
 impl<'d> SpiDevice<'d> {
-    pub fn new(
-        driver: SpiDriver<'d>,
+    pub fn new<T>(
+        host: spi_host_device_t,
```
ivmarkov, line 19: This is incorrect and regressing the existing code: by replacing `SpiDriver` with a raw `spi_host_device_t` you just lost any warranty, that `spi_host_device_t` will be valid until your `SpiDevice` is valid! By using `SpiDriver` or `Borrow<SpiDriver>` or `BorrowMut<SpiDriver>` you have the warranty, that the user cannot drop the `SpiDriver` while your own `SpiDevice` is still alive.

src/sd/spi.rs
```diff
@@ -1,32 +1,23 @@
...
+        _: &mut SpiDeviceDriver<'d, T>,
```
ivmarkov, line 20: This is also incorrect at least for two reasons:
- You don't need the `SpiDeviceDriver` and you are not using it actually?!
- You are not holding onto the `SpiDeviceDriver` throughout the lifetime of `SpiDevice`?

ivmarkov (commented):
As mentioned in my earlier comments: you do need to explain - in English - what is the original problem you are trying to fix. Otherwise I'm afraid we might be "fixing" it by introducing additional problems. As in one step forward, two steps back. :)

AlixANNERAUD:
Thank you for this very comprehensive and useful feedback. As mentioned, I was inspired by the `eth` driver. Ultimately, SpiDriver is indeed sufficient (only in `Borrow` because the SPI bus on an [SD card can be shared](https://docs.espressif.com/projects/esp-idf/en/v5.2.2/esp32/api-reference/peripherals/sdspi_share.html)).

AlixANNERAUD:
The CI failed due to a dependency download issue.

ivmarkov:
> There might be other issues; I can help identifying those but you have to explain a bit (in English) what you are trying to fix with your new PR. Adding some small description of this PR would definitely help a lot.

@AlixANNERAUD Please explain **in English** what this PR is trying to fix.

src/sd/host.rs
```diff
@@ -38,7 +39,7 @@ impl<'d> SdHost<'d> {
         };
 
         #[cfg(not(any(
-            esp_idf_version_major = "4",
+            esp_idf_verdevicesion_major = "4",
```
Vollbrecht, line 42: typo lets CI fail here

ivmarkov:
> I think I rushed the bug fix and completely missed the real issue, which only was incorrect pins in the example and poor management of the SpiDriver's lifetime.

@AlixANNERAUD OK so are you saying that there were _no other issues_ besides:
* Double-free on the SPI BUS, as you were free-ing the SPI bus from `SpiDevice`?
* Wrong pins in the example?

But then, why in #439 the original poster complained? Did he also used the wrong pins, as he copy-pasted from the example?

(The double-free only affects SD card cleanup, it cannot cause the other issue which is that the card could not be initialized.)

ivmarkov:
@AlixANNERAUD You have also removed `SdDevice` which was a public API. So this would be a breaking change and we have to issue a new release. 

I think the API is simpler that way so I would rather keep your change, but maybe we should've put it under the "experimental" flag given that the API is not settled down yet. Anyway...

src/sd/spi.rs
```diff
@@ -52,15 +59,16 @@ impl<'d> SpiDevice<'d> {
...
-        }
+        })
     }
 
     pub fn get_host(&self) -> spi_host_device_t {
-        self.driver.host()
+        self.host
     }
 
     pub fn set_clock(&mut self, clock: u32) -> Result<(), EspError> {
```
ivmarkov, line 74: This is wrong, you are doing it on the wrong handle...

ivmarkov: Also, `enable_intrerrupt` and `wait_interrupt` are wrong for the same reasons.

Oh well... I think we need to more strictly follow the three layers in ESP IDF:
* SPI (or SDMMC) **host** driver; where the SPI driver also uses `SpiDriver)
* SDMMC **higher layer** driver (that uses one of the above host drivers)
* FAT driver (that uses the SDMMC higher layer driver from above)

Otherwise it is a bit messy, both for us, and for the users...

AlixANNERAUD:
> > There might be other issues; I can help identifying those but you have to explain a bit (in English) what you are trying to fix with your new PR. Adding some small description of this PR would definitely help a lot.
> 
> @AlixANNERAUD Please explain **in English** what this PR is trying to fix.

This PR aims to partially address issues in #439 related to SD SPI. In my initial PR (#422), I misunderstood some concepts of the ESP-IDF SD/FS abstraction layer and the SPI interface, leading to a non-working state. This PR fixes:

- The example code (wrong pins and mounting point).
- Lifetime issues where structure instances were dropped prematurely (e.g., `Fat::mount`).
- Proper usage of the SPI driver as a shared SPI driver (`Borrow`).
- Double SPI bus freeing.

However, the MMC part still doesn't work because I currently don't have the necessary hardware on hand, nor access to Wokwi hardware, so I cannot fix this part.

AlixANNERAUD:
> @AlixANNERAUD You have also removed `SdDevice` which was a public API. So this would be a breaking change and we have to issue a new breaking release (new minor version). (UPDATED: But the `Borrow<SpiDevice>` is also a breaking change anyway.)
> 
> I think the API is simpler that way so I would rather keep your change, but maybe we should've put it under the "experimental" flag given that the API is not settled down yet. Anyway...

Yes, I know, it's far from ideal. It depends on when the next minor release is scheduled. In any case, the code concerning the SD card part is not functional. If the next minor release is several months away, I might try to implement a fix (which seems possible) with the current minor version that does not break the current API.

AlixANNERAUD:
> @AlixANNERAUD Also SDMMC's Host driver "slots" should be modeled as peripherals... at the moment you use `NotUsed` for the slots on s3 and p4 as if you can create as many SDMMC drivers as pins you have, which is incorrect...

Yes, indeed, that seems more accurate since there appear to be only two slots on all ESP32s with an SDMMC slot (the only difference being the support for the GPIO matrix).

ivmarkov:
@AlixANNERAUD - closing this in favor of https://github.com/esp-rs/esp-idf-svc/pull/454, where the latter PR is changes on top of the changes in your (this) PR (even if the subsequent changes are quite extensive).

PR #452, opened 2024-07-12
feat(eth): Implement alternative polling mode

stabl-gjn:
Add the option to use a polling strategy instead of a interrupt pin for the SPI EthDriver.
The old api was left intact, while a new `EthDriver::new_spi_with_event_source` was added.

### DIFF
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -114,6 +114,111 @@ pub enum SpiEthChipset {
     KSZ8851SNL,
 }
 
+/// Source/mechanism to use for getting notifications/events from the emac.
+///
+/// # Availability
+///
+/// Pre version `v5.1.4` of esp-idf, only an interrupt pin could be used as source:
+///
+/// - v4.4: <https://github.com/espressif/esp-idf/blob/e499576efdb086551abe309a72899302f82077b7/components/esp_eth/include/esp_eth_mac.h#L461-L464>
+/// - v5.0: <https://github.com/espressif/esp-idf/blob/bcca689866db3dfda47f77670bf8df2a7ec94721/components/esp_eth/include/esp_eth_mac.h#L513-L517>
+/// - v5.1.3: <https://github.com/espressif/esp-idf/blob/e7771c75bd1dbbfb7b3c5381be7e063b197c9734/components/esp_eth/include/esp_eth_mac.h#L612-L617>
+///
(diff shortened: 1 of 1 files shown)


stabl-gjn:
Not 100% sold on the big cfg include/exclude statement and happy over any suggestions.

Vollbrecht:
> Not 100% sold on the big cfg include/exclude statement and happy over any suggestions.

Yeah i also don't like it, though that's what we get when upstream introduces new stuff on patch level releases :dagger: 

Long term we may need to come up with some macro magic or some other alternative. Only having NAND style logic with "any" and "not" is a bit limiting ;D 

I think the rest itself looks fine. With what board / phy chip did you test this stuff here?

Vollbrecht:
@ivmarkov anything you don't like here otherwise i will merge it?

ivmarkov:
Give me a day or so to take a look...

Vollbrecht:
Yeah maybe we can get away with not introducing the SpiEventSource struct, as its more or less just used as a glorified config struct with an exlusive option either poll_period or with pin. We could maybe model the internal init_spi to take an Option<i32> and than that way, we can use either the new spi_with_event_source constructor just taking a poll_period_ms and it its case setting the event_source pin as a None value. 

So still having two different constructors, so we don't break current usage and only changing the internal init_spi. Or is there a special foreseeable usage for an SpiEventSource struct the way it is currently modelt @stabl-gjn ?

ivmarkov:
The only nit I have is that I don't think `EventSource` should be a `struct`. It looks to me more of a candidate for an `enum`.
But that's not the end of the world, so I'll merge as-is.

Thanks!

stabl-gjn:
Yeah i added the extra struct/enum because i though it would make sense to unify the constructors in the future (probably next breaking change), instead of having two options (just thought that would be nicer instead of having 3-4 constructors).
I also first tought about making it an enum, but to not need to make it generic over the input pin (and having to specify that type when creating a polling config because it can not be inferred) and not wanting to publicly expose the raw gpio pin number as variable ( EventSource::Interrupt(gpio_int)).
So i went with the struct instead to hide the implementation details but i am always open for suggestions :)

Thanks for managing this so quickly

jothan:
This branch broke the master branch for me (I do not have ethernet enabled).

I included a tentative bugfix here: https://github.com/esp-rs/esp-idf-svc/pull/457

PR #455, opened 2024-07-13
Wifi event details

jothan:
Exposes some more details. In my use case, I want to use the event stream to display the currently associated BSSID to the user.

### DIFF
--- a/src/eventloop.rs
+++ b/src/eventloop.rs
@@ -246,7 +246,7 @@ impl<'a> EspEvent<'a> {
     ///
     /// Care should be taken to only call this function on fetch data that one is certain to be
     /// of type `P`
-    pub unsafe fn as_payload<P: Copy + Send + 'static>(&self) -> &P {
+    pub unsafe fn as_payload<P: Copy + Send + 'static>(&self) -> &'a P {
         let payload: &P = if mem::size_of::<P>() > 0 {
             self.payload.unwrap() as *const _ as *const P
         } else {
--- a/src/netif.rs
+++ b/src/netif.rs
@@ -602,6 +602,10 @@ impl<'a> DhcpIp6Assignment<'a> {
         self.0.esp_netif
     }
 
+    pub fn addr(&self) -> core::net::Ipv6Addr {
+        Newtype(self.0.ip6_info.ip).into()
+    }
+
     pub fn ip(&self) -> [u32; 4] {
         self.0.ip6_info.ip.addr
     }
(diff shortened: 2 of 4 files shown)


src/eventloop.rs
```diff
@@ -246,7 +246,7 @@ impl<'a> EspEvent<'a> {
     ///
     /// Care should be taken to only call this function on fetch data that one is certain to be
     /// of type `P`
-    pub unsafe fn as_payload<P: Copy + Send + 'static>(&self) -> &P {
+    pub unsafe fn as_payload<P: Copy + Send + 'static>(&self) -> &'a P {
```
Vollbrecht, line 249: Did the compiler wanted that here? P is bounded by 'static. But it would also live to be informed about the 'a lifetime here?

Vollbrecht (commented):
Overall looks fine as i read through once.  Will have a closer look later

src/eventloop.rs
ivmarkov: @Vollbrecht There is quite a difference between the lifetime of `P` and the lifetime of `&P`.

Regardless of that, I have the same question though - @jothan - given that both `'a` and `'_` (as in `&'_ self`) are covariant, why do you need `&'a P` instead of `&'_ P` as it used to be?

ivmarkov:
@jothan Aside from the `'a` lifetime nit (see above) I think the rest looks quite nice!
So if you can justify the `'a` lifetime introduction - or - if not needed after all - roll it back - I think this is ready for merging.

Let me know about that and thanks for the great work - and being able to navigate this undocumented code, and the `*Ref` tricks I introduced recently so as to avoid heap allocs and/or stack blowups.

src/eventloop.rs
jothan: If I leave the 'a part out, the payload reference gets tied to the &self lifetime that does not live long enough when returned from *Event::deserialize. By adding the 'a to &'a P, the lifetime of the Option<&'a ffi::c_void> in EspEvent is simply passed through. Let me know if I am making sense.

```rust
   Compiling esp-idf-svc v0.49.1 (prog/esp-idf-svc)
error: lifetime may not live long enough
    --> src/wifi.rs:2239:50
     |
2234 |     fn deserialize<'d>(data: &crate::eventloop::EspEvent<'d>) -> WifiEvent<'d> {
     |                    --        - let's call the lifetime of this reference `'1`
     |                    |
     |                    lifetime `'d` defined here
...
2239 |             wifi_event_t_WIFI_EVENT_SCAN_DONE => WifiEvent::ScanDone(unsafe { data.as_payload() }),
     |                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ associated function was supposed to return data with lifetime `'d` but it is returning data with lifetime `'1`

error: could not compile `esp-idf-svc` (lib) due to 1 previous error
```

ivmarkov: OK. 
Let's merge for now with your `as_payload` lifetime fix in place, though I think the signature of `fn deserialize` might actually be wrong. I don't see why it should not be:
`fn deserialize<'d>(data: &'d crate::eventloop::EspEvent<'d>) -> WifiEvent<'d>`

... but I'll play with that a bit post-merge of your changes.

PR #457, opened 2024-07-15
BUGFIX: The SpiEventSource impl block should be compiled under the same conditions as the st…

jothan:
…ruct block.


ivmarkov:
Thanks!

jothan:
Thanks for the quick merge !

PR #458, opened 2024-07-19
Fix set_target_level

ivmarkov:
I'm relatively sure `set_target_level` is not currently defined correctly.

This PR fixes this. That is, until / if we decide to de-couple the Rust logging from the ESP IDF logging, which is possible since > 2 years. I.e. since [these changes landed](https://github.com/esp-rs/esp-idf-svc/commit/6e24987f217fbc951161d7caf19ae9841e70b036).

### DIFF
--- a/src/log.rs
+++ b/src/log.rs
@@ -1,12 +1,17 @@
 //! Logging
 use core::fmt::Write;
 
-use ::log::{Level, LevelFilter, Metadata, Record};
+use alloc::collections::BTreeMap;
+use alloc::string::String;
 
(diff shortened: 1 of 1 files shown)


ivmarkov:
This change we need to commit ASAP, but afterwards, we might consider an alternative: now that `Cstr` is part of `core::ffi` and given that `esp_idf_svc::log` is ESP-IDF specific, we can just switch the API (a breaking change of course) to just use `&'static CStr` instead of non-static `&str` and thus get rid of the `alloc` module completely.

PR #459, opened 2024-07-22
get_scan_result version stomps the scan state from Done to Idle.

jothan:
I am using the is_scan_done function to check if I can start a new scan. I noticed this problem when I switched from using get_scan_result_n() to get_scan_result(). The state behavior should at least be matched.


ivmarkov:
Yes, this is a copy paste error. Thanks for spotting & fixing!

PR #461, opened 2024-07-27
Make async mqtt client implement Send

HalfVoxel:
There were already some Send implementations.
However, they weren't completely correct, which made it impossible to use the async mqtt client in e.g. a tokio worker thread.

Note that it could be used in the async main task, because that never sends off any work to other worker threads, and thus Send is not required. The errors would surface when trying to use tokio::spawn with an async mqtt client.

See also https://users.rust-lang.org/t/implementation-of-send-is-not-general-enough-but-cannot-make-it-more-general/115087

Also. I note that the Unblocker deletes the FreeRTOS task when it is dropped. This seems sketchy. Deleting the task will, if I understand correctly, cause it to stop running immediately, without running any associated `Drop` implementations. Couldn't this cause memory leaks? Wouldn't it be better to just let it close the channel and let the worker thread clean itself up?

### DIFF
--- a/src/private/zerocopy.rs
+++ b/src/private/zerocopy.rs
@@ -83,7 +83,7 @@ where
     }
 }
 
-unsafe impl<T> Send for Receiver<T> where T: Send + 'static {}
+unsafe impl<'a, T> Send for Receiver<T> where T: Send + 'a {}
 
 pub struct QuitOnDrop<T>(Arc<Channel<T>>)
 where
@@ -182,8 +182,8 @@ where
     }
 }
 
-unsafe impl<T> Send for Channel<T> where T: Send + 'static {}
-unsafe impl<T> Sync for Channel<T> where T: Send + 'static {}
+unsafe impl<'a, T> Send for Channel<T> where T: Send + 'a {}
+unsafe impl<'a, T> Sync for Channel<T> where T: Send + 'a {}
 
 #[derive(Copy, Clone, Debug)]
 enum State<T> {
--- a/src/private/unblocker.rs
+++ b/src/private/unblocker.rs
@@ -20,6 +20,10 @@ where
     task: TaskHandle_t,
 }
 
+/// SAFETY: Unblocker uses a raw pointer in Unblocker::task. This causes rust to treat it as not Send.
+/// However, this task handle is only ever used to delete the FreeRTOS task, which is safe to do from any thread.
+unsafe impl<T: Send + 'static> Send for Unblocker<T> {}
+
 impl<T> Unblocker<T>
 where
     T: Send + 'static,


src/private/zerocopy.rs
```diff
@@ -83,7 +83,7 @@ where
     }
 }
 
-unsafe impl<T> Send for Receiver<T> where T: Send + 'static {}
+unsafe impl<'a, T> Send for Receiver<T> where T: Send + 'a {}
```
ivmarkov, line 86: Why is this necessary and what benefit does it bring? `T: 'static` is still throughout all of `Receiver` and `Channel` and I think removing it would require some very careful thinking, as that might be unsound.

src/private/zerocopy.rs
```diff
@@ -182,8 +182,8 @@ where
     }
 }
 
-unsafe impl<T> Send for Channel<T> where T: Send + 'static {}
-unsafe impl<T> Sync for Channel<T> where T: Send + 'static {}
+unsafe impl<'a, T> Send for Channel<T> where T: Send + 'a {}
```
ivmarkov, line 185: Ditto.

src/private/zerocopy.rs
```diff
@@ -182,8 +182,8 @@ where
...
+unsafe impl<'a, T> Sync for Channel<T> where T: Send + 'a {}
```
ivmarkov, line 186: Ditto.

ivmarkov (changes requested):

ivmarkov:
> Note that it could be used in the async main task, because that never sends off any work to other worker threads, and thus Send is not required. The errors would surface when trying to use tokio::spawn with an async mqtt client.

To use `async` code with a **work stealing** executor, like tokio's default one (because you can still use tokio `LocalSet`), you need the `Future`s _themselves_ to be `Send`. 

Most of the futures in `esp-idf-svc` are not `Send` because this is often difficult to do and work-stealing is not such a big deal on embedded devices. 

With that said, I agree that if it is _relatively easy_ to make the futures of some driver/service `Send`, we should do it.
In your particular case we have hit the jackpot by just making `Unblocker` `Send`, and this works simply because the ESP IDF `MqttClient` is actually anything but natively async. So we simulate async-ness by spinning a hidden thread. Terrible, I know, but that's as much as we can do and I think even this is too much. We just need all-Rust App-layer network implementations like the ones provided by `edge-net` and other crates. Unfortunately, no all-rust, embedded-friendly MQTT client that supports the 3.1 protocol is available anywhere yet.

> See also https://users.rust-lang.org/t/implementation-of-send-is-not-general-enough-but-cannot-make-it-more-general/115087

Ah! I take my workds back for the `'a` lifetime. another Rust paperclip. :(


> Also. I note that the Unblocker deletes the FreeRTOS task when it is dropped. This seems sketchy. Deleting the task will, if I understand correctly, cause it to stop running immediately, without running any associated `Drop` implementations. Couldn't this cause memory leaks? Wouldn't it be better to just let it close the channel and let the worker thread clean itself up?

Yes! This is spot-on. Would you enhance the PR with this fix? Basically

src/private/zerocopy.rs
ivmarkov: Got it - it is this one: https://github.com/rust-lang/rust/issues/64552

src/private/zerocopy.rs
ivmarkov: Ditto - https://github.com/rust-lang/rust/issues/64552

HalfVoxel:
> Ah! I take my words back for the 'a lifetime. Another Rust paperclip. :(

Yes, I believe this is a limitation in the current rust compiler.

Essentially, rust does not understand that the original bound is enough.

The new bound 
```rust
unsafe impl<'a, T> Send for Receiver<T> where T: Send + 'a {}
```
essentially says, for all lifetimes `'a` where `T: Send + 'a` implement `Send` for `Receiver<T>`. But since `Receiver<T>` requires T to be `'static`, this will not allow anything that is unsound.

HalfVoxel:
> So we simulate async-ness by spinning a hidden thread. Terrible, I know, but that's as much as we can do and I think even this is too much.

Oof. That sounds horrible. :sob:

ivmarkov:
Since the `'a` lifetimes seem to _have_ to be there because of the rustc bug, I'll merge this change as-is.
Though I would appreciate if you take the effort to fix the `FreeRTOS` task deletion to happen from within the unblocker trait itself.


Need to review if we need the same unsafe pattern elsewhere. We might be...

ivmarkov:
Thanks!

ivmarkov:
> > So we simulate async-ness by spinning a hidden thread. Terrible, I know, but that's as much as we can do and I think even this is too much.
> 
> Oof. That sounds horrible. 😭

Didn't you see what `Unblocker` _actually_ does? :p

I mean... if you would like to contribute a truly-async client to `edge-net` or - like - improve [`rust-mqtt`](https://github.com/obabec/rust-mqtt) (seems to be a bit of an abandonware ATM and does not really support 3.11) but in the meantime, that's all we have...

HalfVoxel:
> Didn't you see what Unblocker actually does? :p

Hmm. Everything I've seen seems to indicate it uses a separate thread that waits using conditional variables and that kind of stuff. But I haven't seen anything that *spins* (i.e. looping using a short delay, which is how I interpret "spin").
Where is that code?

ivmarkov:
> > Didn't you see what Unblocker actually does? :p
> 
> Hmm. Everything I've seen seems to indicate it uses a separate thread that waits using conditional variables and that kind of stuff. But I haven't seen anything that _spins_ (i.e. looping using a short delay, which is how I interpret "spin"). Where is that code?

My English. :) I didn't mean to say that we busyloop by polling stuff with short sleeps in-between. That would truly be indeed very horrible.

By "spinning" I meant simply that the blocking MQTT ESP IDF client is turned into an async one by working it in a separate thread and then delegating to that separate thread the execution. And using condvars to sync indeed.

Still pretty gross, as that's not a "native" async. A bit like the hack how `smol` and `async-io` turn the blocking Rust STD file IO into "async". That's fake async as you spin a separate hidden thread in there, with its own stack etc.

HalfVoxel:
> My English. :) I didn't mean to say that we busyloop by polling stuff with short sleeps in-between. That would truly be indeed very horrible.

Ah. I see!
Yeah, not the prettiest, but workable.

PR #464, opened 2024-07-28
Avoid potential memory leak when dropping mqtt clients

HalfVoxel:
- Improve docs for mqtt/client
- Improve docs for private/zerocopy
- Avoid potential memory leak when dropping mqtt clients

### DIFF
--- a/src/private/unblocker.rs
+++ b/src/private/unblocker.rs
@@ -87,7 +87,7 @@ where
             worker();
         }
         unsafe {
-            // FreeRTOS tasks must delete themselves, returning from the task function
+            // FreeRTOS tasks must delete themselves. Returning from the task function
             // without deleting the task will cause a crash.
             // This function will immediately stop running. This means we have to ensure
             // that all memory is freed before calling vTaskDelete.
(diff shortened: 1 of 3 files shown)


ivmarkov:
Ping me when this is ready for review.

HalfVoxel:
Should be ready now.

src/private/unblocker.rs
```diff
@@ -73,10 +79,22 @@ where
...
+        }
+        unsafe {
+            // FreeRTOS tasks must delete themselves. Returning from the task function
+            // without deleting the task will cause a crash.
+            // This function will immediately stop running. This means we have to ensure
+            // that all memory is freed before calling vTaskDelete.
+            // Fortunately, this function does not manage any memory at this point.
+            // The scope above guarantees that all relevant variables are dropped at this point.
+            // See https://www.freertos.org/implementing-a-FreeRTOS-task.html
+            task::destroy(std::ptr::null_mut());
```
Vollbrecht, line 96: replace `std` with `core` please. That will also make CI happy.

HalfVoxel: Fixed

PR #465, opened 2024-08-01
add esp32p4 support

Vollbrecht:
Still TODO


ivmarkov:
I think we should close this because p4 support is in since quite some time.

Issue #383, opened 2024-03-11
Ethernet event deserialization panic


torkleyy:
Not quite sure how that can happen, using `EspSystemEventLoop::subscribe_async::<EthEvent>(..)`, I was able to trigger this panic:

```
thread '<unnamed>' panicked at ~/.cargo/git/checkouts/esp-idf-svc-1132d30333045bb2/a3aefb7/src/eth.rs:1059:13:
Unknown event ID: 4
```

Using ESP-IDF 5.1.3

It's not a valid discriminant according to the docs, but maybe it's a good idea to add an `Unknown` variant?

Vollbrecht:
hmm sounds like an of-by-one error. Maybe there is some bug lurking somewhere. Just downgrading it to a Unknown variant is i think not a thing here as it is really unexpected here. 

If it is some sort of a of-by-one, can this be triggerd when you force a ETHERNET_EVENT_DISCONNECTED externally?

torkleyy:
Not quite sure how to trigger it, but I've encountered a similar bug:

```rust
let _ip_events = sysloop
  .subscribe_async::<IpEvent>()
  .unwrap();
```

```
I (100457) wifi:state: init -> auth (b0)
thread '<unnamed>' panicked at /Users/thomas/.cargo/git/checkouts/esp-idf-svc-1132d30333045bb2/b84572d/src/netif.rs:738:13:
Unknown event ID: 40
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
I (100467) wifi:state: auth -> assoc (0)
abort() was called at PC 0x4018d802 on core 0

TRACE 40083584 (40083596) panic_abort
TRACE 4008b240 (4008b245) esp_system_abort
TRACE 400925c0 (4009263e) abort
TRACE 4018d7fc (4018d802) panic_abort::__rust_start_panic::abort
TRACE 4018d7f0 (4018d7f6) __rust_start_panic
TRACE 401a4830 (401a4836) rust_panic
TRACE 401a5034 (401a52d9) std::panicking::rust_panic_with_hook
TRACE 401a4fe8 (401a5014) std::panicking::begin_panic_handler::{{closure}}
TRACE 401a4fdc (401a4fe4) std::sys_common::backtrace::__rust_end_short_backtrace
TRACE 401a56f8 (401a5717) rust_begin_unwind
TRACE 4017c754 (4017c770) core::panicking::panic_fmt
[...]

PR #466, opened 2024-08-04
Filter asynchronous events

Rantanen:
Applies an event source filter to asynchronously subscribed events. Previously async subscriptions would attempt to deserialize all events in the event loop with the specified deserializer.

Fixes (suspected) #383 for me.

There's probably a better way to implement the event filter so that it gets applied in the event loop thread before channeling the events all the way to the async thread, but this is what I came up with as a quick fix.

Feel free to close the PR and implement a better approach. :)

### DIFF
--- a/src/eventloop.rs
+++ b/src/eventloop.rs
@@ -441,12 +441,16 @@ where
             self.given = false;
         }
 
-        if let Some(data) = self.receiver.get_shared_async().await {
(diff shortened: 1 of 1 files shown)


ivmarkov:
Yep, ideally we should just not subscribe to all events I stead, but such a fix would be more involved.

For now, I'll just merge yours as a quick fix, thanks!

PR #469, opened 2024-08-13
ESP Modem

DaneSlattery:


src/modem.rs
```diff
@@ -0,0 +1,259 @@
...
+            serial,
+            netif,
+            base,
+            _d: PhantomData,
+        })
+    }
+
+    unsafe extern "C" fn post_attach(netif: *mut esp_netif_obj, args: *mut c_void) -> i32 {
+        let driver= unsafe{std::ptr::slice_from_raw_parts_mut(args, size_of::<ppp_netif_driver())}
+        let ifconfig = esp_netif_driver_ifconfig_t{handle}
```
DaneSlattery, line 44: Really struggling to wrap my head around this part at the moment

ivmarkov: OK. Let me step back and try to explain `esp_netif_driver_base_t`, what is so base about it and how the ESP IDF folks envisioned it is going to be used.

It is called "base", as it is a form of "poor man's" inheritance in the C world, that - fortunately - can be used from within Rust as well.

* The ESP IDF assumes that you are going to have some sort of a PPP "driver" struct.  
* The closest in your current code to a driver is the `PppNetif` struct (but we won't use that directly, read on).
* When you call `esp_netif_attach`, you pass a `&mut` so to say, to your driver (say, to `PppNetif` or something similar)
* When the time had come for ESP-IDF to call `post_attach` (which would be much, much later, long after the call to `esp_netif_attach` is done and you have exited the `PppNetif::new` constructor), ESP IDF will do the following unsafe cast of the `void* base_ptr` that you pass to it: _it will try to cast it to `*esp_netif_base_t`_, so that it can extract from the `esp_netif_base_t` struct the `post_attach` callback so that it can call **you**
* "So far so good", you would say - "my `base_ptr` is anyway _A_ `esp_netif_base_t` instance?
  * And here we come to the "base" part. Yes, you are passing a pointer to a `esp_netif_base_t` instance so that would work, but the idea actually is to pass something _BIGGER_ that `esp_idf_base_t` to `esp_netif_attach`, which contains - besides `esp_netif_base_t` - _also your own "user" data_ - user data, that _you_ can "get back" when your own `PppNetif::post_attach` is called and that's data _that you might need_ so that _your_ `PppNetif::post_attach` can do its job.

The following magical struct would work and you can pass to ESP IDF as a `base_ptr`  (I mean, you can pass a _pointer_ to it, of course, not the "whole" struct):
```rust
#[repr(C)]
struct MyDriver {
    base: esp_netif_base_t,
    netif: EspNetif,
    foo: Foo,
    bar: Bar,
    ... more stuff
}
```

Since it is `repr(C)` (i.e. fields-reordering that Rust sometimes does is forbidden) and since `base` is the FIRST member of this struct, when ESP IDF tries to cast `&MyDriver` to `&esp_netif_base_t` it would actually work! And the benefit is that you are passing to ESP IDF something much larger. Still - ESP IDF is able to cast it to something smaller (to esp_netif_base_t, because, well, it is the _first_ member in the struct layout), extract from the "smaller" part what it needs (the netif handle and your `post_attach` callback) and then - _when the time had come_ - call your `post_attach` callback with a pointer to `MyDriver`, so that you have your own data in your `post_attach` function.

OK, seemingly all is well and good, and you might be tempted to slap a `#[repr(C)]` on top of your `PppNetif` struct and then pass a `&PppNetif` reference to `esp_netif_attach`  (strictly speaking a `&PppNetif as *const _ as *const c_void as *mut _`).

Will that work?
No.

...which is the second issue you need to workaround: unlike C (and C++), Rust has _move_semantics. What it means, is that unless you take special care, passing a `&some_rust_struct` might crash, because at the time when the C code needs to use `&some_rust_struct`, `some_rust_struct` might have _moved_ to another memory location, so `&some_rust_struct` - at that time - is **invalid**.

In your particular case, passing `&PppNetif` to `esp_netif_attach` - _from within `PppNetif::new`_ - is not OK, because - even if you construct `PppNetif` eariler in `new` - after `new` is done, `PppNetif` will be _moved_ to a new location (as `new` returns `Self` = `PppNetif`)...

The easiest way to workaround this is to pass to `esp_netif_attach` something else which:
* Still contains all the data you'll need in `PppNetif::post_attach`
* Is **boxed**. I.e. if we call this type `PppNetifDriver`, you need to - in `PppNetif::new` - create it with `let driver = Box::new(PppNetifDriver { ... })` and then pass to `esp_netif_attach` a `driver.as_ptr()` pointer to the box
* ... and of course you are free to preserve `driver` (`Box<PppNetifDriver>`) inside `PppNetif` so that at the very least, the heap memory is dropped when the user drops `PppNetif` and also the destructor of `PppNetifDriver` (if any) runs as part of the `PppNetif` destructor
* **The most important is**: since the box is allocated on-heap, when `PppNetif` is moved around, `PppNetifDriver` **won't move**. It is on the heap after all. So you can pass a reference to the heap memory to the ESP IDF C code, and (as long as `PppNetif` is still alive), that heap memory would be valid and OK to use.
* Obviously, in `PppNetif::post_attach` you need to do `unsafe { (args as *mut PppNetifDriver).as_mut() }.unwrap()` and you'll get a `&mut PppNetifDriver` and then you can use it however you want to finish the "post-attached" logic.

(
Digression:
The above "anti-move" approach utilizes a heap allocation. While this is sub-optimal:
- ESP IDF itself allocates like mad, so I think we doing an allocation in these circumstances (one-time creation of a long-living object - the driver) is justified
- The alternative (Rust pinned types) is considered a complex animal in Rust, and it will _affect the API of `PppNetif`_ in that (a) initialization will become two-phase; (b) users would be required to pin `PppNetif` before executing the second-phase initialization. Given that there is a large base of users who reach out to the ESP IDF wrappers not because they are more complete, but because they are perceived as "simpler" than bare-metal, we definitely do not want to introduce pinning in the API.


)

Hope that helps. Please follow up with more questions if something is unclear.

ivmarkov (commented):
I've left a long comment to what I believe is your main problem with `post_attach`.

src/modem.rs
ivmarkov: Thinking about your code, I feel the easiest way to apply the patterns I'm talking about might be to do something like:
```rust
pub struct PppNetifDriver {
    netif: EspNertif,
    inner: Box<PppNetifDriverInner>,
}

impl PppNetifDriver {
    /// Create a new PPP driver with the provided ipv4 configuration and a TX
    /// callback that the driver would call when it wants to ingest a PPP packet into the 
    /// underlying transport.
    /// The transport can be anything, but would typically be UART, and the callback implementation
    /// is simply expected to write the PPP packet into UART
    pub fn new<F>(conf: &crate::ipv4::Configuration, tx: F) -> Result<Self, EspError> 
    where 
        F: FnMut(&[u8]) -> Result<(), EspError>,
    {
        let netif = EspNetif::new(...);
        let inner = Box::new(PppNetifDriverInner {
            base: esp_netif_driver_base_t {
                netif: netif.handle(),
                post_attach: PppNetifDriverInner::raw_post_attach,
            },
            tx: Box::new(tx),
        });

        let inner_ptr = inner.as_ptr() as *const c_void as *mut _;

        esp!(unsafe { esp_event_handler_register(NETIF_PPP_STATUS, ESP_EVENT_ANY_ID, netif.handle()) })?;
        esp!(unsafe { esp_event_handler_register(IP_EVENT, IP_EVENT_PPP_GOT_IP, esp_netif_action_connected, netif.handle()) })?;
        esp!(unsafe { esp_event_handler_register(IP_EVENT, IP_EVENT_PPP_LOST_IP, esp_netif_action_disconnected, netif.handle()) })?;

        esp!(unsafe { esp_netif_attach(netif.handle(), inner_ptr) })?;

        Ok(Self {
            netif,
            inner,
        })
    }

    /// Ingest a PPP packet into the driver. The PPP packet can arrive from anywhere, but 
    /// typically from UART, by reading from it.
    pub fn rx(&self, data: &[u8]) -> Result<(), EspError> {
        self.inner.rx()
    }
}

#[repr(C)]
struct PppNetifDriverInner {
    base: esp_netif_driver_base_t,
    tx: Box<dyn FnMut(&[u8]) -> EspError>, // Boxing only necessary so as to erase `F` from the driver type signature; if we keep it, then we can geit rid of this particular boxing. But given that we box anyway the inner driver, what gives?
}

impl PppNetifDriverInner {
    fn post_attach(&mut self, netif_handle: *mut esp_netif_obj) -> Result<(), EspError> {
    }

    unsafe extern "C" fn post_attach(netif: *mut esp_netif_obj, args: *mut c_void) -> i32 {
        let this = (args as *mut PppNetifDriverInner).as_mut() }.unwrap();
        match this.post_attach(netif) {
            Ok(_) => ESP_OK,
            err(e) => e,code(),
        }
    }
}
```

ivmarkov: Heh.
Now that I think about that code from above some more, _there is nothing PPP-specific in that code_!
It is just a way to attach an RX/TX functionality to a Netif so that you ingest/ingress (RX) packets into the netif, and you can egress (TX) packets from the netif. It might be used with a PPP Netif, or it might be used with - whatever.

So if we fix the naming and generalize the code just a tad more, we end up with:
```rust
pub struct EspNetifDriver<T> 
where 
	T: Borrow<EspNetif>,
{
    inner: Box<EspNetifDriverInner<T>>,
}

impl<T> EspNetifDriver<T> 
where 
	T: Borrow<EspNetif>,
{
    /// Create a new netif driver around the provided `EspNetif` instance and a TX
    /// callback that the driver would call when it wants to ingest a packet into the 
    /// underlying transport.
    /// 
    /// The transport can be anything, but with - say - PPP netif - it would typically be UART, and the callback implementation
    /// is simply expected to write the PPP packet into UART
    ///
    /// `post_attach_cfg` is a netif-specifif configuration that will be executed after the netif is attached
    /// See `ppp_config` below for a PPP-specific driver post-attach configuration
    pub fn new<F, P>(netif: T, post_attach_cfg: P, tx: F) -> Result<Self, EspError> 
    where 
        F: FnMut(&[u8]) -> Result<(), EspError>,
        P: FnMut(&EspNetif) -> Result<(), Error>,
    {
        let inner = Box::new(PppNetifDriverInner {
            base: esp_netif_driver_base_t {
                netif: netif.borrow().handle(),
                post_attach: PppNetifDriverInner::raw_post_attach,
            },
            netif,
            post_attach_cfg: Box::new(post_attach_cfg),
            tx: Box::new(tx),
        });

        let inner_ptr = inner.as_ptr() as *const c_void as *mut _;

        // TODO: We should rather use the safe event loop API
        // TODO: May not even be necessary to do this from _within_ the driver - it can be done completely in user code
        // TODO Also, PPP-specific, and best left to the user to do, from outside
        // esp!(unsafe { esp_event_handler_register(NETIF_PPP_STATUS, ESP_EVENT_ANY_ID, PppNetifDriverInner::raw_on_ppp_changed) })?;

        // TODO: Check if `EspNetif` is not doing this already...
        esp!(unsafe { esp_event_handler_register(IP_EVENT, IP_EVENT_PPP_GOT_IP, esp_netif_action_connected, inner.netif.borrow().handle()) })?;
        // TODO: Ditto
        esp!(unsafe { esp_event_handler_register(IP_EVENT, IP_EVENT_PPP_LOST_IP, esp_netif_action_disconnected, inner.netif.borrow().handle()) })?;

        esp!(unsafe { esp_netif_attach(inner.netif.borrow().handle(), inner_ptr) })?;
        esp!(unsafe { esp_netif_action_start(inner.netif.borrow().handle(), core::ptr::null_mut(), 0, core::ptr::null_mut()) })?;

        Ok(Self {
            inner,
        })
    }

    /// Ingest a packet into the driver. 
    /// The packet can arrive from anywhere, but with say - a PPP netif - it would be a PPP packet arriving typically from UART, by reading from it.
    pub fn rx(&self, data: &[u8]) -> Result<(), EspError> {
        esp!(unsafe { esp_netif_receive(self.inner.netif.borrow().handle(), data.as_raw_ptr() as *mut c_void, len as _, core::ptr::null_mut()) })
    }
}

impl<T> Drop for EspNetifDriver<T> 
where 
	T: Borrow<EspNetif>,
{
    fn drop(&mut self) {
        esp!(unsafe { esp_netif_action_stop(self.inner.netif.borrow().handle(), core::ptr::null_mut(), 0, core::ptr::null_mut()) }).unwrap();

        // TODO: See constructor
        // esp!(unsafe { esp_event_handler_unregister(NETIF_PPP_STATUS, ESP_EVENT_ANY_ID, &EspNetifDriverInner::raw_on_ppp_changed) }).unwrap();

        // TODO: See constructor
        esp!(unsafe { esp_event_handler_unregister(IP_EVENT, IP_EVENT_PPP_GOT_IP, esp_netif_action_connected) }).unwrap();
        // TODO: See constructor
        esp!(unsafe { esp_event_handler_unregister(IP_EVENT, IP_EVENT_PPP_LOST_IP, esp_netif_action_disconnected) }).unwrap();
    }
}

#[repr(C)]
struct EspNetifDriverInner<T> 
where 
	T: Borrow<EspNetif>,
{
    base: esp_netif_driver_base_t,
    tx: Box<dyn FnMut(&[u8]) -> EspError>, // Boxing only necessary so as to erase `F` from the driver type signature; if we keep it, then we can geit rid of this particular boxing. But given that we box anyway the inner driver, what gives?
    netif: T,
}

impl<T> EspNetifDriverInner<T> 
where 
	T: Borrow<EspNetif>,
{
    fn post_attach(&mut self, netif_handle: *mut esp_netif_obj) -> Result<(), EspError> {
        let driver_ifconfig = esp_netif_driver_ifconfig_t {
            transmit = Self::raw_tx;
            handle = self as *mut c_void,
        };

        // d->base.netif = esp_netif; TODO: This is weird; the netif in base is already set on constructor?
        
        esp!(unsafe { esp_netif_set_driver_config(netif_handle, &driver_ifconfig) })?;

		(self.post_attach_cfg)()?;

        Ok(())
    }

    fn tx(&mut self, data: &[u8]) -> Result<(), EspError> {
        (self.tx)(data)
    }

    fn on_ppp_changed(&mut self, event_base: esp_event_base_t,
                           event_id: i32, event_data: *const c_void) {
        if event_id > NETIF_PPP_ERRORNONE && event_id < NETIF_PP_PHASE_OFFSET {
            todo!()
            // ESP_LOGI("esp_modem_netif", "PPP state changed event %" PRId32, event_id);
            // only notify the modem on state/error events, ignoring phase transitions
            // ppp->signal.set(PPP_EXIT);
        }
    }

    unsafe extern "C" fn raw_tx(h: *mut c_void, buffer: *mut c_void, len: usize) -> i32 {
        let this = (args as *mut Self).as_mut() }.unwrap();
        let data = core::slice::from_raw_parts(buffer as *mut u8, len);
        let result = match this.tx(data) {
            Ok(_) => ESP_OK,
            err(e) => e,code(),
        };

        // TODO: Might not be necessary, but if I remember correctly, the Netif API
        // wanted that _we_ free the buffer; in any case needs to be compared with the C ESP Modem code
        free(buffer);

        result
    }

    unsafe extern "C" fn raw_post_attach(netif: *mut esp_netif_obj, args: *mut c_void) -> i32 {
        let this = (args as *mut Self).as_mut() }.unwrap();
        match this.post_attach(netif) {
            Ok(_) => ESP_OK,
            err(e) => e,code(),
        }
    }

    unsafe extern "C" fn raw_on_ppp_changed(arg: *mut c_void, event_base: esp_event_base_t,
                           event_id: i32, event_data: *const c_void) {
        let this = (args as *mut Self).as_mut() }.unwrap();
        this.on_ppp_changed(event_base, event_id, event_data);        
    }
}

/// PPP-specific configuration of a Netif
pub fn ppp_config(netif: &EspNetif) -> Result<(), EspError> {
    // check if PPP error events are enabled, if not, do enable the error occurred/state changed
    // to notify the modem layer when switching modes
    let mut ppp_config = esp_netif_ppp_config_t {
        ppp_phase_event_enabled = true,    // assuming phase enabled, as earlier IDFs
        ppp_error_event_enabled = false,   // don't provide cfg getters so we enable both events
        ...Default::default()
    };
    
    #[cfg(not(esp_idf_version_major = "4"))]
    esp!(unsafe { esp_netif_ppp_get_params(esp_netif, &ppp_config) })?;

    if !ppp_config.ppp_error_event_enabled {
        ppp_config.ppp_error_event_enabled = true;
        esp!(unsafe { esp_netif_ppp_set_params(esp_netif, &ppp_config) })?;
    }
    
    Ok(())
}
```

ivmarkov (commented):
I've left some sample code now.

src/modem.rs
DaneSlattery: > Since it is repr(C) (i.e. fields-reordering that Rust sometimes does is forbidden) and since base is the FIRST member of this struct, when ESP IDF tries to cast &MyDriver to &esp_netif_driver_base_t it would actually work! And the benefit is that you are passing to ESP IDF something much larger. Still - ESP IDF is able to cast it to something smaller (to esp_netif_driver_base_t, because, well, it is the first member in the struct layout), extract from the "smaller" part what it needs (the netif handle and your post_attach callback) and then - when the time had come - call your post_attach callback with a pointer to MyDriver, so that you have your own data in your post_attach function.

The fact that casting `EspNetifDriverInner` to   `esp_netif_driver_base_t` takes the first member feels like arcane magic. Brilliant.

src/netif.rs
```diff
@@ -873,20 +873,22 @@ where
...
     use core::borrow::Borrow;
+    use core::marker::PhantomData;
 
     use crate::handle::RawHandle;
     use crate::sys::*;
 
     use super::EspNetif;
 
-    pub struct EspNetifDriver<T>
+    pub struct EspNetifDriver<'d, T>
```
ivmarkov, line 883: I don't mind introducing the `'d` lifetime (in fact, I was planning to do it anyway, so good catch!), as it would allow you to borrow non-static stuff into the `tx` closure, BUT you have to realize this is **unsafe in Rust**. And then reason why Rust allows you to do it, is because the code inside `EspNetifDriver` is littered with `unsafe` so Rust can't notice the unsafeness of what you are doing.

It is unsafe because, the moment you call `core::mem::forget(your-esp-netif-driver)`, from the point of view of Rust - everything you have borrowed in the `tx` closure is no longer borrowed and can be re-used.

Except the underlying C code **does not think that's the case and will happily continue to use your borrowed stuff (including the closure itself) thus leading to memory corruption**. That's exactly what Rust can't notice - it can't "see through" the C code and realize that our non-static closure is in fact being sent to a hidden FreeRTOS task/thread which will _continue to operate even if you `core::mem::forget` the driver_.

What we do in [other places of the code](https://github.com/esp-rs/esp-idf-svc/blob/master/src/wifi.rs#L1004) is to rename the `new_*` constructors that take non-static closures to `new_nonstatic_*`, slap a big comment that users should be aware that these are unsafe (as in, users should not use `core::mem::forget` on a driver created with a non-static constructor), and well - mark them as `unsafe`.

So I suggest you rename the eixsting constructors to `new_nonstatic_*` slap the same comment on top, and then re-implement regular `new_*` constructors that take `'static` closures and which call into the `new_nonstatic_*` ones (as they become safe when `'d: 'static`).

src/netif.rs
```diff
@@ -873,20 +873,22 @@ where
...
     where
         T: Borrow<EspNetif>,
     {
-        inner: alloc::boxed::Box<EspNetifDriverInner<T>>,
+        inner: alloc::boxed::Box<EspNetifDriverInner<'d, T>>,
+        _d: PhantomData<&'d ()>,
```
ivmarkov, line 888: This `PhantomData` is completely unnecessary, because another member - `EspNetifDriverInner` - is already lifetimed by `'d` so there is nothing "phantom" here. The phantom data is also incorrect - since the closure is `FnMut`, it should be should be `PhantomData<&'d mut ()>` instead.

ivmarkov (changes requested):
A few things to fix. BTW, is this sort of working now, or what happens when you try to run it?

src/netif.rs
DaneSlattery: The reason I introduced the `'d` lifetime is because the serial device (in this case a `UartTxDriver`) that is given to the `tx` closure is not a `'static` lifetime.

DaneSlattery:
Right now I'm not seeing any events on the `IP_EVENT_PPP_GOT_IP`  or `NETIF_PPP_ERRORUSER` topics.

This is what my console shows so far, I'm not sure yet if the modem is fully able to connect to the LTE network: 
```
I (7373) esp_idf_svc::modem: connect Some("115200")
D (7383) esp_netif_lwip: LwIP stack has been initialized
D (7383) esp_netif_lwip: esp-netif has been successfully initialized
D (7393) esp_netif_lwip: check: remote, if=0x3fc9fa70 fn=0x420e510c
0x420e510c - esp_netif_new_api
    at /home/dane/dev/esp-idf-svc/.embuild/espressif/esp-idf/v5.2.2/components/esp_netif/lwip/esp_netif_lwip.c:690
D (7393) esp_netif_objects: esp_netif_add_to_list_unsafe netif added successfully (total netifs: 1)
D (7403) esp-netif_lwip-ppp: esp_netif_new_ppp: PPP connection created: 0x3fca7c48
D (7413) esp_netif_lwip: call api in lwip: ret=0x0, give sem
I (7423) esp_idf_svc::netif::driver: Post attach ifconfig: esp_netif_driver_ifconfig { handle: 0x3fca8f14, transmit: Some(0x4200b920), transmit_wrap: None, driver_free_rx_buffer: None }
0x4200b920 - esp_idf_svc::netif::driver::EspNetifDriverInner<T>::raw_tx
    at /home/dane/dev/esp-idf-svc/src/netif.rs:1103
I (7433) esp_idf_svc::netif::driver: Netif config: esp_netif_ppp_config { ppp_phase_event_enabled: true, ppp_error_event_enabled: false }
D (7453) esp_netif_handlers: esp_netif action has started with netif0x3fca7a18 from event_id=0
D (7453) esp_netif_lwip: check: remote, if=0x3fca7a18 fn=0x420e53c8
0x420e53c8 - esp_netif_start_api
    at /home/dane/dev/esp-idf-svc/.embuild/espressif/esp-idf/v5.2.2/components/esp_netif/lwip/esp_netif_lwip.c:1068
D (7463) esp_netif_lwip: esp_netif_start_api 0x3fca7a18
D (7473) esp-netif_lwip-ppp: esp_netif_start_ppp: Starting PPP connection: 0x3fca7c48
D (7473) esp_netif_lwip: check: local, if=0x3fca7a18 fn=0x420e5c88
0x420e5c88 - esp_netif_update_default_netif_lwip
    at /home/dane/dev/esp-idf-svc/.embuild/espressif/esp-idf/v5.2.2/components/esp_netif/lwip/esp_netif_lwip.c:316
D (7483) esp_netif_lwip: esp_netif_update_default_netif_lwip 0x3fca7a18
D (7493) esp_netif_lwip: call api in lwip: ret=0x0, give sem
```

src/modem.rs
```diff
@@ -0,0 +1,241 @@
...
+            if len > 0 {
+                driver.rx(&buff)?;
+            }
+        }
+
+        Ok(())
+    }
+
+    fn tx(writer: &mut UartTxDriver, data: &[u8]) -> Result<(), EspError> {
+        writer.write(data)?;
```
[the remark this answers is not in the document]
DaneSlattery: I used `write_all` from embedded_io

src/modem.rs
```diff
@@ -0,0 +1,335 @@
...
+        esp!(unsafe {
+            esp_event_handler_register(
+                NETIF_PPP_STATUS,
+                ESP_EVENT_ANY_ID as _,
+                Some(Self::raw_on_ppp_changed),
+                netif.handle() as *mut core::ffi::c_void,
+            )
+        })?;
+        let (mut tx, rx) = self.serial.borrow_mut().split();
+        let driver = unsafe {
```
ivmarkov, line 101: This line - and the `loop {}` below is probably the one thing in `EspModem/SimModem` which is specific to the underlying platform (i.e. ESP-IDF, `embassy-net-ppp` and Linux). We can probably consider isolating this code in a separate function which is `#[cfg(target_os = "espidf")]`. Or even better - we can introduce a trait - `PppRunner` or something that the `SimModem` expects on constructor and which trait is responsible for connecting the `Read/Write` IO to the underlying PPP stack once the negotiation with the modem is complete.

Something like (in blocking-IO terms, for now):
```rust
trait PppRunner {
    /// Hand the read/write pipes of the modem to the underlying platform PPP stack once 
    /// the negotiation with the modem with AT commands is over and it is time to switch to PPP data mode
    fn run<R, W>(&mut self, read: R, write: W) -> Result<(), R::Error>
    where 
        Self: Sized,
        R: embedded_io::Read,
        W: embedded_io::Write<Error = R::Error>;
}
```

... and the implementation for e.g. ESP IDF would be something like:
```rust
struct EspPppRunner(());

impl EspPppRunner {
    pub const fn new() -> Self {
        Self(())
    }

    pub fn run<R, W>(read: R, write: W) -> Result<Self, EspError> 
    where 
        R: embedded_io::Read,
        W: embedded_io::Write<Error = R::Error>,
    {
        let netif = EspNetif::new(NetifStack::Ppp)?;
        let driver = EspNetifDriver::new_ppp(move |data| esp_idf_hal::io::Write::write_all(write, data).map_err(|e| e.0))?;

        let mut buff = [0u8; 64];

        loop {
            let len = rx.read(&mut buff, BLOCK)?;
            if len > 0 {
                driver.rx(&buff[..len])?;
            }
        }

        // TODO: We should watch out for PPP frames that signify end of PPP communication and exit the loop    
    }
}

impl PppRunner for EspPppRunner {
    fn run<R, W>(&mut self, read: R, write: W) -> Result<(), R::Error> 
    where 
        R: embedded_io::Read,
        W: embedded_io::Write<Error = R::Error>,
    {
        EspPppRunner::run(self, read, write)
    }
}
```

Note also, that what we can also do is **not even deal** with the PPP stack in your `SimModem` implementation, and this way it would become pure Rust. 

Basically, rather than renaming `setup_data_mode` to    `run` - as I suggested above - we can leave it named `setup_data_mode` or shorten to `negotiate` or something, and then just document that once `negotiate` is over, the user is supposed to establish the native PPP connection of the `read/write` pipes to the underlying platform, using platform-specific means (whatever they are) without even bothering to introduce a `PppRunner` trait to your `SimModem`, because - after all - the main value of your `SimModem` would be in the pre-PPP "AT commands" phase where - ideally and over time - it would gain support for other various modems besides `SIM7600`.

src/modem.rs
```diff
@@ -0,0 +1,335 @@
...
+    T: BorrowMut<UartDriver<'d>> + Send,
+{
+    pub fn new(serial: T) -> Self {
+        Self {
+            serial,
+            _d: PhantomData,
+        }
+    }
+
+    pub fn send_cmd<CMD: AtatCmd>(&mut self, cmd: &CMD) -> Result<CMD::Response, atat::Error> {
```
ivmarkov, line 37: I don't think this needs to be public. Also maybe move below `setup_data_mode` which I think is/should be the main (and only) public API of your modem struct?

src/modem.rs
```diff
@@ -0,0 +1,335 @@
...
+        let mut buff = [0u8; 64];
```
ivmarkov, line 38: Allocating these buffs on-stack seems wasteful. How about changing `EspModem::new` constructor to take buffer and then re-use the buffer for all commands, and also potentially for the PPP connection (the RX loop)?

src/modem.rs
```diff
@@ -0,0 +1,335 @@
...
+        let len = self
+            .serial
+            .borrow_mut()
+            .read(&mut buff, TickType::new_millis(1000).ticks())
+            .map_err(|_err| atat::Error::Read)?;
+        log::info!("got response {:?}", &buff[..len]);
+        cmd.parse(Ok(&buff[..len]))
+    }
+
+    pub fn setup_data_mode(&mut self, sysloop: EspSystemEventLoop) -> Result<(), EspError> {
```
ivmarkov, line 64: Given that this function does _more_ than just setting up data mode - i.e. it "runs" the PPP stack RX/TX loop, I suggest to rename it to simply "run". It should also run "forever" or until the PPP TX/RX loop exits because it had lost connection to the modem, or because the loop receives a PPP frame that designates the modem had exited.

ivmarkov (changes requested):
Most important in my feedback is that we need to think what is `EspModem`'s purpose in life, so to say? 

I.e. shall we hard-wire it to Espressif MCUs _and_ to UART (as it is right now), or can we do something (a) more flexible, that supports more protocols than just UART (b) make it useful in contexts beyond ESP IDF (i.e. embassy-net and Linux) (c) Make it pure-Rust and ideally having life on its own, outside of `esp-idf-svc`?

This way, a future `sim-modem` crate or something might become _the_ way on embedded, to do the "pre-PPP" AT-commands-based negotiation with various supported modems and can be used in many more contexts than just what you are currently aiming at - esp32 with the modem connected over UART.

src/modem.rs
ivmarkov: ... or - if we decide that `EspModem/SimModem` should only be platform-neutral, "pure Rust" and only dealing with the "pre-PPP" phase, we can leave it named `setup_data_mode` OR maybe `negotiate`.

src/modem.rs
```diff
@@ -0,0 +1,335 @@
...
+use crate::{
+    eventloop::EspSystemEventLoop,
+    handle::RawHandle,
+    netif::{EspNetif, EspNetifDriver, NetifStack},
+    sys::*,
+};
+
+pub struct EspModem<'d, T>
+where
+    T: BorrowMut<UartDriver<'d>>,
```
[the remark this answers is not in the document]
ivmarkov: Update 3: [esp32p4 seems to even support USB 2.0 high-speed](https://www.espressif.com/en/news/ESP32-P4), i.e. 480MBIT

src/modem.rs
```diff
@@ -0,0 +1,335 @@
+use at_commands::{builder::CommandBuilder, parser::CommandParser};
```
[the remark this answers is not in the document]
DaneSlattery: image

This is how I understand the `SimModem` will work. I want to be able to keep track of which mode the device is in using command mode and data mode.

I also want to provide the user code access to the `Read` and `Write` interface so that they can hook up the ppp `rx` and `tx` through `SimModem`

I think the `SimModem` will also be generic over type `T`, where `T: DCE`. DCE will be a new introduced trait, with a single impl for `SimComA7600`, which takes care of getting the device into data mode. I also think we should allow for `send_cmd` to be public so that you can still do things like:
get the iccid,
get the imei
get the signal strength

src/modem.rs
```diff
@@ -0,0 +1,335 @@
...
+        // check pdp network reg
+        self.read_gprs_registration_status()?;
+
+        //configure apn
+        self.set_pdp_context()?;
+
+        // start ppp
+        self.set_data_mode()?;
+
+        // now in ppp mode.
```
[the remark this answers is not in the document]
DaneSlattery: Yes this is valid. I do think the subscription is currently duplicated in `netif.rs`, but I think we should keep it there and remove it here.

src/modem.rs
DaneSlattery: I like the idea to change this to `negotiate`. let 'run' become a user concern.

src/modem.rs
DaneSlattery: I think this should be public within the `DCE` trait.

src/modem.rs
DaneSlattery: Yes, this is a good idea

src/modem.rs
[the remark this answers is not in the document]
ivmarkov: > image

Not really sure what is "DTE" on that picture? For your modem, the "terminal" is just the pair of `embedded_io::Read + embedded_io::Write`. Why do you need anything else here?

> I also want to provide the user code access to the `Read` and `Write` interface so that they can hook up the ppp `rx` and `tx` through `SimModem`

I do not understand this. It is the **user code** that will give **your** sim modem the `Read` + `Write` traits, so the user code **apriori** has access to those. You only need to do something special if/when you plan to support MUX mode, where somehow on top of this user-provided Read+Write you'll implement two "virtual" tunnels:
- One, for commands, but you keep this inside the modem
- Another, for PPP (data) -> this you need to expose to the users

But maybe we need to think about this once you have support for MUX? Because at the moment, in the context of "either-commands-or-data" you don't need to provide to the user anything...

> I think the `SimModem` will also be generic over type `T`, where `T: DCE`. DCE will be a new introduced trait, with a single impl for `SimComA7600`, which takes care of getting the device into data mode.

Isn't this the same as my `PppRunner` thing?
By the way, I see you are copying the terminology from the C ESP modem. No offense, but this terminology is highly confusing... DTE, DCE... yeah.

> I also think we should allow for `send_cmd` to be public so that you can still do things like: get the iccid, get the imei get the signal strength

Sure no prob.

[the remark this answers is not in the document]
DaneSlattery: > Isn't this the same as my PppRunner thing?
> By the way, I see you are copying the terminology from the C ESP modem. No offense, but this terminology is highly confusing... DTE, DCE... yeah.

No not quite. `PPPRunner` describes how to feed the `tx` and `rx`. `ModemDevice`(the C api calls this `DCE`) describes how to get a specific device into data mode. For this case `SIM7600` sends those specific commands, but not every modem will have the same formatting, delimiters, responses or newlines etc.

`DTE` is any "terminal", but essentially a device that implements `embedded_io::Read+embedded_io::Write`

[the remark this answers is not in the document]
ivmarkov: Ah, ok. Instead of having multiple `SimModem`s, like `struct Sim7600Modem`, `struct Br96Modem`, `struct Sim800Modem` where these all implement a common `trait SimModem` trait, you want to model it slightly differently.

OK.

src/modem.rs
```diff
@@ -0,0 +1,373 @@
...
+            let (ccid,) = CommandParser::parse(&buff[..len])
+                .expect_identifier(b"\r\n+ICCID: ")
+                .expect_raw_string()
+                .expect_identifier(b"\r\n\r\nOK\r\n")
+                .finish()?;
+
+            Ok(heapless::String::try_from(ccid).unwrap())
+        }
+
+        fn reset(comm: &mut UartDriver) -> Result<(), ModemError> {
```
DaneSlattery, line 243: @ivmarkov 
Please note the difference between these two functions. I tried to make `SimModem` generic across `R,W`, but the read is giving me an issue.

One takes the `UartDriver` and one takes `embedded_io::Read+embedded_io::Write`

The blocking read in `embedded_io::Read` is never returning in my example code, and I have no idea why.

The timeout-based read in `UartDriver` works every time.

DaneSlattery: I think there may be an issue with the impl of `embedded_io::Read` or a deeper fundamental problem in esp-idf with the uart, because this function sometimes does not return.
```
impl<'d> embedded_io::Read for UartDriver<'d> {
    fn read(&mut self, buf: &mut [u8]) -> Result<usize, Self::Error> {
        UartDriver::read(self, buf, delay::BLOCK).map_err(EspIOError)
    }
}
```

src/modem.rs
DaneSlattery: I have created a `run` loop, but the only issue is that it mutably borrows. I need to work around that so that one can call "is_connected" etc

src/modem.rs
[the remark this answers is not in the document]
DaneSlattery: I opted to keep it simpler. `SimModem` is a trait with 
`negotiate()` and `get_mode()`. Can probably add a few more general  things such as signal strength, ICCID, IMEI.

`EspModem` stays as a kind of glue between `EspNetifDriver` and `UartDriver`

src/modem.rs
ivmarkov: It would only ever return when the other side sends at least one byte to you. Or else it will wait forever. You should only call it without a timeout (as you do) only when you expect an answer to something, or like a constant inflow of data.

ivmarkov: Whe you say the timeout based read works every time... are you actually checking the returned result? Because the only difference between a timeout read and a blocking read is that the blocking read will only return when it had read at least one Bute, while the timeout read might return when the timeout had expired. But under the hood, they are a single ESP IDF function?!

DaneSlattery: https://github.com/esp-rs/esp-idf-hal/issues/475 See this ticket. I am sure there is an issue

DaneSlattery: > When you say the timeout based read works every time... are you actually checking the returned result? Because the only difference between a timeout read and a blocking read is that the blocking read will only return when it had read at least one Bute, while the timeout read might return when the timeout had expired. But under the hood, they are a single ESP IDF function?!

Yes I do check the returned result. In this case, the buffer contains the value I expect from the modem, and I can parse it using `at_commands`. If the parse were failing (it hasn't happened yet), then I would be sure that the full message had not been read before the timeout.

src/modem.rs
```diff
@@ -427,31 +426,167 @@ pub mod sim {
...
                 Self::new()
             }
         }
 
         impl SimModem for SIM7600 {
-            fn negotiate(&mut self, comm: &mut UartDriver) -> Result<(), ModemError> {
-                reset(comm)?;
+            fn negotiate(
+                &mut self,
+                comm: &mut UartDriver,
```
ivmarkov, line 566: You are returning the `UartDriver` where you already had a nice abstraction just because you need to do a read-with-a-timeout.

This is unnecessary. Reading with a timeout is as simple as combining `Read` with - say - `embedded_hal_async::Delay` like so:

```rust
async fn read_with_timeout<R: embedded_io_async::Read, D: embedded_hal_async::DelayNs>(read: R, delay: D, buf: &mut [u8], timeout_ms: u16) -> Result<usize, R::Error> {
    let read = read.read(buf);
    let timeout = delay.delay_ms(timeout_ms as _);

    match embassy_futures::select::select(read, timeout).await {
        Either::First(result) => result,
        Either::Second(_) => Ok(0), // 0 bytes read indicates a timeout
    }
}
```

src/modem.rs
```diff
@@ -462,39 +597,45 @@ pub mod sim {
...
+            buff: &mut [u8],
+        ) -> Result<(RSSI, BitErrorRate), ModemError> {
+            let cmd = CommandBuilder::create_execute(buff, true)
                 .named("+CSQ")
                 .finish()?;
 
             comm.write(cmd).map_err(|_| ModemError::IO)?;
 
-            let len = comm.read(&mut buff).map_err(|_| ModemError::IO)?;
+            let len = comm
```
ivmarkov, line 610: I don't think this type of reading (and all others in all `get_*` methods below and above) is reliable.

You are not taking into account, _that you might not get the complete response as a single shot to `read`_. Here for example, the fact that you get `\r\n+CSQ: 19,99\r\n\r\nOK\r\n` out of a single `read` is a pure coincidence, and might break any time. Just think what would've happened, if I've kept the other code in `UartDriver::read`, which did return - on an empty ringbuffer - **a single byte** after waiting for it. Havoc.

I think you need a utility method that probably takes a callback closure of sorts, where the callback closure is containing the parsing code of `CommandParse::parse` and then - in the absence of a reliable end-of-command-response termination character (like a CR or LF) that you might wait for - is trying to apply parsing on the buffer repeatedly, and if the data in the buffer is incomplete - tries to read a bit more at the end of the buffer.

For that purpose, you might want to take a look at `embedded_io_async::BufRead` as its API is handy for such cases (and implementing `BufRead` over `Read` is trivial, if not done already in the lib itself.

Alternative: you might want to layer a async fn which reads line-by-line from a `BufRead` and refactor your parsing to line-by-line (i.e. CommandParser would expect lines somehow).

Reading line-by-line simply means you call `BufRead::read` repeatedly until you notice a sequence of `\r\n` in the buffer, then consume everything until (and trimming) `\r\n` and pass this as the "line" to your stateful parser of sorts.

ivmarkov: A third alternative (but I find this really ugly) and it will make the negotiation slower, is to `read` in your buffer - in a loop - until a given timeout. Where the expectation is that until this timeout had expired, the modem would've replied - in your buffer - with the complete response to your command.

But of course this introduces an arbitrary variable in all of this (the "read till you get the whole response" timeout, which would be - what? - a trial and error exercise).

ivmarkov (changes requested):

src/modem.rs
ivmarkov: @DaneSlattery  Ah sorry!
I did not realize your code still lives in "blocking"-land and not in `async`-land yet. :D Perhaps we can delay the re-introduction of `embedded_io_async::Read + Write` until after we start to switch this code from blocking to async! For now, maybe just make it work with `UartDriver` (where the you-assume-that-your-read-reply-will-contain-the-whole-response issue is _the_ more important issue), and once we have something mostly working reliably, we can switch from blocking to async. And then still layer blocking _on top_ of async (Or use `AsyncUartDriver` maybe).

ivmarkov: (
The issue at hand is that stuff like `read_with_timeout` is much, much more difficult to implement in blocking-land (would require scoped threads) than it is in async-land. This is one place where `async` is strictly more powerful than blocking - the "intra-task concurrency" from the blog of Boats, if you recall that.
)

src/modem.rs
```diff
@@ -0,0 +1,752 @@
...
+            serial,
+            status,
+            _subscription: subscription,
+            netif: EspNetif::new(NetifStack::Ppp)?,
+            _d: PhantomData,
+        })
+    }
+
+    /// Run the modem network interface. Blocks until the PPP encounters an error.
+    pub fn run(&mut self, mut buffer: [u8; 64]) -> Result<(), EspError> {
```
ivmarkov, line 47: Nit: `buffer: &mut [u8]`. (a) you probably don't want to move the buffer into `run` (b) why constrain the size of the buffer to 64 bytes? User could pass whatever she wants.

ivmarkov:
@DaneSlattery because this is a large effort with many moving parts, I hope you don't mind if I merge a portion of the changes already now? What I think is ready for merge is your "nonstatic" enhancements to `EspNetifDriver` (basically "as-is") as well as the mapping of the PPP event loop events (I'll only rename `ModemEvent` to `PppEvent` and move it to inside `netif.rs` instead of `modem.rs` as it is kinda sorta Netif stuff too - and - is kinda generic too (who knows what else besides an LTE modem might want to use PPP).

To have you listed as a contributor to this code in GIT history and yet - to save you some time - I'll fork your PR branch, revert the other changes which are still not ready (`EspModem`) and the ESP-IDF-agnostic modem stuff, and then merge that way, so it would be your changes and then my changes on top and we listed both as contributors.

Once the changes land in `master`, you should re-base your branch on top.

**But** please tell me if you prefer with me to wait until we have everything in-place instead. I'm still not sure how I feel specifically about the `EspModem` struct, but the other changes I enlisted absolutely need to land inside `esp-idf-svc` as they are very useful stuff which is ESP-IDf specific anyway

src/modem.rs
ivmarkov: Looking at the `atat` crate, [here is a similar "loop" to what I mean](https://github.com/FactbirdHQ/atat/blob/master/atat/src/asynch/simple_client.rs#L52). Basically, the loop reads at the end of the buffer and then hands the buffer to a "digester". The digester would return `DigestResult::None` if the data in the buffer is "incomplete" (i.e. we have to read more to get the whole response).

DaneSlattery:
> @DaneSlattery because this is a large effort with many moving parts, I hope you don't mind if I merge a portion of the changes already now? What I think is ready for merge is your "nonstatic" enhancements to EspNetifDriver (basically "as-is") as well as the mapping of the PPP event loop events (I'll only rename ModemEvent to PppEvent and move it to inside netif.rs instead of modem.rs as it is kinda sorta Netif stuff too - and - is kinda generic too (who knows what else besides an LTE modem might want to use PPP).

I am happy to get this in so long, go ahead, I'll rebase once the dust is settled. I have started porting stuff to a `SimModem` crate so that `SimModem` is generic with `embedded_io::Read+embedded_io::Write` (now that `UartDriver` does what we expect, I can actually move away from the TimeOut based code `read`) 

I saw your comments on the PR, and I was only using the timeout to force `UartDriver:::read` to return the bytes. I was confident that after 1s the modem will have replied, but that is completely arbritrary. I like the solution to feed parts into a `Digester` that is a state machine searching for some `terminator` (like `'\r\nOK\r\n'.as_bytes()`. This can happen after a few `read` cycles. After the terminator is found, then we can try parse it using the `at_commands` parser. If the terminator is not found, the digester will keep waiting. It would actually be really useful if `at_commands::parser` could provide `expects_terminator(...)` and throw a different error if the terminator is not found.


> But please tell me if you prefer with me to wait until we have everything in-place instead. I'm still not sure how I feel specifically about the EspModem struct, but the other changes I enlisted absolutely need to land inside esp-idf-svc as they are very useful stuff which is ESP-IDF specific anyway

Go ahead. I do think `EspModem` has a place in `esp-idf-svc` as a glue, but perhaps in a different form. It's serving quite a few functions that might be better suited to `EspNetifDriver`, where `EspModem` can just wrap `EspNetifDriver`. That said, it's still quite specific to the `esp-idf`, and it will be a challenge for anyone to use a modem with `esp-idf-svc` without the glue.

src/modem.rs
DaneSlattery: I saw your comments on the PR, and I was only using the timeout to force UartDriver:::read to return the bytes. I was confident that after 1s the modem will have replied, but that is completely arbritrary. I like the solution to feed parts into a Digester that is a state machine searching for some terminator (like '\r\nOK\r\n'.as_bytes(). This can happen after a few read cycles. After the terminator is found, then we can try parse it using the at_commands parser. If the terminator is not found, the digester will keep waiting. It would actually be really useful if at_commands::parser could provide expects_terminator(...) and throw a different error if the terminator is not found.

src/modem.rs
DaneSlattery: @ivmarkov any ideas for the mutable borrow here. I want to run the `run` loop in a separate thread (spawned in userland), but I also want to do things like `EspModem.wait_netif_up()` after the `run` loop, but this requires another borrow. You can see my failed attempts in the `examples/lte_modem.rs`

src/modem.rs
DaneSlattery: > I think you need a utility method that probably takes a callback closure of sorts, where the callback closure is containing the parsing code of CommandParse::parse and then - in the absence of a reliable end-of-command-response termination character (like a CR or LF) that you might wait for - is trying to apply parsing on the buffer repeatedly, and if the data in the buffer is incomplete - tries to read a bit more at the end of the buffer.

I do think at least that `\r\nOK\r\n` is a reliable end of command termination.

src/modem.rs
ivmarkov: Using an `Arc<Mutex<...>>` **externally** to `EspModem::run` will be impossible, because the mutex will lock the modem for as long as the `run` method "runs" so to say, which is not what we want.

So if we have any interior mutability (i.e. `Mutex` or similar) it has to be internal to the modem struct itself.

I think we have two options:
- Option A: (I have a slight prefernce for this) Turn `run(&mut self)` into `run(&self)` and then internally use `Mutex`es (i.e. interior mutability)
- Option B: Introduce `EspModem::split(&mut self) -> (EspModemRunner<'_>, EspModemStatus<'_>), where `EspModemRunner` just has a `EspModemRunner::run(&mut self)` and `EspModemStatus` has whatever you need in terms of status (`is_connected` etc.)

Given that both options will likely require some form of interior mutability after all (i.e. a `Mutex` or suchlike) I would say perhaps Option A?

src/modem.rs
DaneSlattery: Yeah done

src/modem.rs
DaneSlattery: I like option A, although it seems like we would need to internally Arc+mutex the `netif` and `serial` members

src/modem.rs
DaneSlattery: This is what I came up with:
```

pub enum DigestError {
    Error,
    Continue,
    BufferOverflow,
}

pub trait Digester {
    /// Termination characters if the response indicates the OK state.
    const OK_TERMINATOR: &'static str;
    /// Termination characters if the response indicates the Error state.
    const ERR_TERMINATOR: &'static str;
    /// parse the buffer, searching for finaliser.
    /// If the
    fn digest(&mut self, data: &[u8]) -> Result<&[u8], DigestError>;
}

pub struct SimDigester<const N: usize> {
    buf: [u8; N],
    pos: usize,
}

impl<const N: usize> Digester for SimDigester<N> {
    const OK_TERMINATOR: &'static str = "\r\nOK\r\n";
    const ERR_TERMINATOR: &'static str = "\r\nERROR\r\n";

    fn digest(&mut self, data: &[u8]) -> Result<&[u8], DigestError> {
        let incoming_len = data.len();
        if self.pos + incoming_len > N {
            return Err(DigestError::BufferOverflow);
        }
        self.buf[self.pos..self.pos + incoming_len].copy_from_slice(data);
        self.pos += incoming_len;
        // look for the terminator, get to the choppa
        let ok_len = Self::OK_TERMINATOR.len();
        let end = &self.buf[(self.pos - ok_len)..self.pos];
        if end == Self::OK_TERMINATOR.as_bytes() {
            // ready to parse
            return Ok(&self.buf[..self.pos]);
        }
        let err_len = Self::ERR_TERMINATOR.len();
        let end: &[u8] = &self.buf[(self.pos - err_len)..self.pos];
        if end == Self::ERR_TERMINATOR.as_bytes() {
            // result indicated failure
            return Err(DigestError::Error);
        }
        Err(DigestError::Continue)
    }
}
```

I can also modify `Digester` to be more like `BufRead`, if that is better approach?

ivmarkov: I don't think you need the digester thing given that you can just use `BufRead` indeed.

Be mindful of the last response to the command with which you switch to PPP. There you might have to read byte-by-byte, or else you might read a bit into the first PPP packet right after the end of the last AT response. Speculating of course, not sure if the modem will start sending PPP packets immediately, but you might need to check that.

DaneSlattery: I think it might be good while I'm at it to introduce `BufReader` struct to `embedded-io`, similar to [std](https://doc.rust-lang.org/std/index.html)::[io](https://doc.rust-lang.org/std/io/index.html)::[BufReader](https://doc.rust-lang.org/std/io/struct.BufReader.html#).

```

pub struct BufReader<const N: usize, R: ?Sized> {
    buf: [u8; N],
    pos: usize,
    inner: R,
}

impl<const N: usize, R: Read> ErrorType for BufReader<N, R> {
    type Error = R::Error;
}

impl<const N: usize, R: Read> BufRead for BufReader<N, R> {
    fn consume(&mut self, amt: usize) {
        // self
        self.pos += amt;
    }

    fn fill_buf(&mut self) -> Result<&[u8], Self::Error> {
        // fill the inner buffer
        let read_count = self.inner.read(&mut self.buf[self.pos..])?;
        // return the read bytes
        Ok(&self.buf[self.pos..(self.pos + read_count)])
    }
}
```

DaneSlattery: I struggled to get this to work with `embedded_io::BufRead`, and even more struggle with `std::io::BufReader`.

However, this is my current approach, let me know if this is suitable (currently targeting `std::io::Read`, so that I can debug it on my local PC)

```rust

pub struct SIM7600(CommunicationMode);

impl<T: std::io::Read + std::io::Write> SimModem<T> for SIM7600 {
    fn negotiate(&mut self, comm: &mut T, buffer: &mut [u8]) -> Result<(), ModemError> {
        reset(comm, buffer)?;
// others ignored here
        self.0 = CommunicationMode::Data;
        Ok(())
    }

    fn get_mode(&self) -> &CommunicationMode {
        &self.0
    }
}



fn reset<T: std::io::Read + std::io::Write>(
    comm: &mut T,
    buff: &mut [u8],
) -> Result<(), ModemError> {
    let cmd = CommandBuilder::create_execute(buff, false)
        .named("ATZ0")
        .finish()?;
    log::info!("Send Reset");

    comm.write(cmd).map_err(|_| ModemError::IO)?;

    // read until terminator
    let len = read_until_term(comm, buff)?;

    log::info!("got response{:?}", core::str::from_utf8(&buff[..len]));
    CommandParser::parse(&buff[..len])
        .expect_identifier(b"ATZ0\r")
        .expect_identifier(b"\r\nOK\r\n")
        .finish()?;
    Ok(())
}

const OK_TERMINATOR: &'static str = "\r\nOK\r\n";
const ERR_TERMINATOR: &'static str = "\r\nERR\r\n";

fn read_until_term<T: std::io::Read>(comm: &mut T, buff: &mut [u8]) -> Result<usize, ModemError> {
    let mut pos = 0;

    loop {
        // start filling the buffer
        let len = comm.read(&mut buff[pos..]).map_err(|w| ModemError::IO)?;
        if len == 0 {
            continue;
        }
        pos += len;
        log::info!(
            "Len: {}, data = {:?}",
            len,
            std::str::from_utf8(&buff[..pos])
        );

        let ok_len = OK_TERMINATOR.len();
        if pos >= ok_len {
            let end = &buff[pos - ok_len..pos];
            log::info!(" end = {:?}", std::str::from_utf8(&buff[pos - ok_len..pos]));
            if end == OK_TERMINATOR.as_bytes() {
                // buff[..pos].copy_from_slice(data);
                return Ok(pos);
            }
        }

        let err_len = ERR_TERMINATOR.len();
        if pos >= err_len {
            let end = &buff[pos - err_len..pos];

            if end == ERR_TERMINATOR.as_bytes() {
                return Err(ModemError::DigestError);
            }
        }
        // return Err(ModemError::DigestError);
    }
}
```

ivmarkov: Also this:
```rust
fn fill_buf(&mut self) -> Result<&[u8], Self::Error> {
        // fill the inner buffer
        let read_count = self.inner.read(&mut self.buf[self.pos..])?;
        // return the read bytes
        Ok(&self.buf[self.pos..(self.pos + read_count)])
}
```

needs to be:
```rust
fn fill_buf(&mut self) -> Result<&[u8], Self::Error> {
        // fill the inner buffer
        let read_count = self.inner.read(&mut self.buf[self.pos..])?;
        self.pos += read_count; // You need to update the end of your data
        // return the read bytes
        Ok(&self.buf[..self.pos]) // You always return ALL of your data UNTIL the user - via `consume` - tells you to remove N bytes from the beginning of the buffer
}
```

ivmarkov: As for the approach that uses a raw `Read` (`read_until_term`)... I'm not sure it is correct either. The problem is, the call to `T::read` might read data **beyond** your line terminator (i.e. data from the _next_ line). You need to preserve this data somewhere. And then re-use it in the _next_ call to `read_until_term`.

That's why STD line readers are implemented with `BufRead` rather than with `Read`. Since `BufRead` "owns" the buffer, it can **preserve** whatever remains in it (i.e. whatever is still not declared as "consumed" by the downstream code) for your _next_ call to `read_until_terminator`, that would use `BufRead`.

ivmarkov: (Update: remove the comment above `Line::Continue` as you actually can't get rid of it as the `BufRead::fetch_buf` does not guarantee you to fetch extra, if there is already content in the buffer. The only way to recover whole lines from the above example code is by appending continue+continue+...+end in a large enough buffer. Or by switching your parsing logic to an iterator-like approach where you process byte-by-byte (= char-by-char for ascii).

DaneSlattery: > Be mindful of the last response to the command with which you switch to PPP. There you might have to read byte-by-byte, or else you might read a bit into the first PPP packet right after the end of the last AT response. Speculating of course, not sure if the modem will start sending PPP packets immediately, but you might need to check that.

With this requirement, I think we will need a byte-by-byte approach anyway, at least for the last packet. 

> The above is wrong.
> consume needs to physically remove amt bytes from the beginning of the buffer (i.e. the bytes of &buf[..amt]), by doing an overlapping-copy of &buf[amt..pos] (src) into &buf[..pos - amt] (dest).
> And then you need to do self.pos -= amt;.

I think I fundamentally didn't understand `BufRead`, but I think I get it now.

> Why? UartDriver already does implement Read, so BufReader<N, UartDriver<'d>> would just work.

Yes you are right. I have implemented a `BufReader` targeting `embedded-io`. https://github.com/rust-embedded/embedded-hal/pull/627, that should work with `UartDriver`. 

> As for the approach that uses a raw Read (read_until_term)... I'm not sure it is correct either. The problem is, the call to T::read might read data beyond your line terminator (i.e. data from the next line). You need to preserve this data somewhere. And then re-use it in the next call to read_until_term.

In this case, i'm not sure that is so much of an issue because the AT commands are request ->response, although I see some mention of the idea of "URC" or Unsolicited Result Code, but I haven't encountered them yet, at least for SIM7600.

> Not sure what is the difficulty with BufRead but here's something which returns stuff line by line (with a simpler terminator indeed, but you can extend to a more complex one):

BufRead seemed to be returning the entire `buffer[self.pos..]`, including the un-initialized bits. I think I still prefer an approach that reads the entire buffer until `\r\nOK\r\n` is found, but I am still working on it.

ivmarkov: > BufRead seemed to be returning the entire `buffer[self.pos..]`, including the un-initialized bits. I think I still prefer an approach that reads the entire buffer until `\r\nOK\r\n` is found, but I am still working on it.

No, `BufRead::fill_buf` should - by definition -  **only** return a slice from its internal buffer, which contains valid, initialized data (and all of it; that is a- ll of it that had not yet been `consume`d).

With that said, I DO agree that at least because of the last request-response which happens _right before_ the channel starts talking PPP, you might have to switch to a byte-by-byte consumption of the input data, because otherwise you risk "eating up" the beginning of the first incoming PPP packet into the internal buffer of `BufRead` (though I would check if PPP really operates in a way where it might stanrt sending you data immediately after the response to your last AT command).

**OR** - and maybe that's better? - you can just **continue** utilizing the _same_ `BufRead` object that was initially used for AT commands _also_ for the PPP stream? That's completely possible, and BTW layering `Read` over `BufRead` should be trivial I think (though maybe bearing an unnecessary overhead as with PPP you don't need the "naive" overlapping copy, that the hidden `BufRead` would inevitably do).

But... your call. :)

DaneSlattery: I think we will have to create a new `BufReader` in `negotiate()` that is passed into each of the read/write parsing functions.

DaneSlattery:
I'm picking this up

PR #470, opened 2024-08-15
Remove unnecessary buffer draining in HTTP client example

weiying-chen:
#404 renders this code unnecessary:

```
// Drain the remaining response bytes
while response.read(&mut buf)? > 0 {}
```

Note: I tested on my machine. The [http_client.rs](https://github.com/esp-rs/esp-idf-svc/blob/master/examples/http_client.rs) example still runs successfully.


weiying-chen:
The CI is broken or something?

Vollbrecht:
yeah gh actions are degraded if you look at https://www.githubstatus.com/ .. will rerun the action as soon as they are up running again

Vollbrecht (approved):
Thanks!

PR #471, opened 2024-08-16
Added http_local_network_server example

MakeIt4Less:
The current http_server example creates an access point that a user needs to connect to in order to access the server. This example connects to the users local network and allows any device connected to the same network to access the server.


Vollbrecht (changes requested):
I fail to see how this is significantly different from our already existing example. If you want to provide examples feel free to make them about aspects that are not already covert or try to improve existing ones.

MakeIt4Less:
It is not much different but shows how you can create a server on your local network as opposed to requiring the server to be on its own standalone access point. For me connecting to a network as a client and then acting as a server was not intuitive and I figured this example might save someone else some headache later on.

ivmarkov:
I think putting a short comment [here](https://github.com/esp-rs/esp-idf-svc/blob/master/examples/http_server.rs#L103) that explains that just switching from `AccessPointConfiguration` to `ClientConfiguration` is as easy as pie might be good enough.

Also note that the configuration of the HTTP server _itself_ is completely orthogonal to how the Wifi network is configured. It is another topic that the configuration of the Wifi should probably not be in the `configure_server` function but maybe a separate `configure_wifi` one (all other examples do it like that, I think). If you would like to PR such a change (and then a comment on top of `configure_wifi` explaining the above) I think that might be a better PR, than just duplicating the example just because of how the Wifi happens to be configured in the original one. Or else where do we stop? We probably don't need a 3rd example - HTTP-over-ethernet? :)

MakeIt4Less:
I have pushed a change removing the new test and adding a comment. As far as separating out the WiFi configuration it seems to be dealers choice in the examples. I factored it out to be in line with the http_client example.

ivmarkov:
Thanks!

PR #473, opened 2024-08-24
Netif PPP

ivmarkov:
* Allow non-static callbacks for `EspNetifDriver`
* Change `T: Borrow<EspNetif>` to `T: BorrowMut<EspNetif>` which is the correct one, as the driver mutates the netif
* Remove the pesky `got_ip_event_id / lost_ip_event_id` from the constructor functions; these are extracted directly from the wrapped `EspNetif`
* Add explicit `start`/`stop`/`is_started` methods to make the API more flexible, and to match more closely the API of `EspEth(Driver)` and `EspWifi(Driver)`
* Expose `EspNetifDriver::netif` and `EspNetifDriver::netif_mut`
* Map PPP event loop events to `PppEvent`
* Implement `PppConfiguration` as well as `EspNetif::set_ppp_conf / EspNetif::get_ppp_conf`
* Remove `new_slip` and `new_ppp` constructor function variants now that there is a safe `set_ppp_conf` / `get_ppp_conf` API
* Convert got/lost event IDs to `Option<NonZeroU32>` which is equivalent to `u32` where 0 means no event
* Implement `PppAuthentication` as well as `EspNetif::set_ppp_auth`


src/netif.rs
```diff
@@ -897,18 +901,11 @@ mod driver {
         ///   `EspNetif` instance is expected to be a PPP netif.
         /// * The `tx` callback implementation is simply expected to write the
         ///   PPP packet into e.g. UART
-        #[cfg(esp_idf_lwip_ppp_support)]
```
Vollbrecht, line 900: did you accidentally removed the ppp_support cfg here?

src/netif.rs
```diff
@@ -952,6 +949,154 @@ mod driver {
...
+        /// The destructor of the service takes care - prior to the service being dropped and e.g.
+        /// the stack being unwind - to remove the closure from the hidden thread and destroy it.
+        /// Unfortunately, when the service is forgotten, the un-subscription does not happen
+        /// and invalid references are left dangling.
+        ///
+        /// This "local borrowing" will only be possible to express in a safe way once/if `!Leak` types
+        /// are introduced to Rust (i.e. the impossibility to "forget" a type and thus not call its destructor).
+        pub fn new_nonstatic<F, P>(
+            netif: T,
+            got_ip_event_id: Option<i32>,
```
Vollbrecht, line 1092: if you are talking about ip_event_id's you are talking about [this](https://github.com/espressif/esp-idf/blob/master/components/esp_netif/include/esp_netif_types.h#L98) right? Why do we accept a i32 here and not one of this concrete variants?

src/netif.rs
ivmarkov: Yes but this method will be removed shortly, as it no longer makes sense. New changes upcoming. Will temporarily turn into "draft".

Vollbrecht (commented):
Overall looks fine, Though added two small questions.

src/netif.rs
ivmarkov: because `ip_event_t` is anyway defined as `pub type ip_event_t = ::core::ffi::c_uint (= u32);` by `bindgen`. A C-style enum is not converted to a Rust enum with our current `bindgen` settings, and long ago we decided not to touch this (it is like that by default in `bindgen`).

Vollbrecht: To be clear, i mainly talking about the ergonomics of the API. The user see's that this api wants a u32 or something, but now needs to find something in the bindgen that is an ip_event_t_* to get the actual thing. So there is a slightly disconnect here.

ivmarkov: What would you suggest?
`NonZeroIp_event_t` does not exist and if I replace `Option<NonZeroU32>` with `Option<ip_event_t>` it is not so perfect either, as user then might pass Some(0), which does not make sense, as it is the same as `None`.

It should be either just `ip_event_t` or `Option<NonZeroU32>` and each has pros and cons.

src/netif.rs
```diff
@@ -1102,14 +1207,16 @@ mod driver {
...
+            #[allow(clippy::let_and_return)]
             let result = match this.tx(data) {
                 Ok(_) => ESP_OK,
                 Err(e) => e.code(),
             };
 
             // TODO: Might not be necessary, but if I remember correctly, the Netif API
             // wanted that _we_ free the buffer; in any case needs to be compared with the C ESP Modem code
-            free(buffer);
+            // free(buffer);
```
Vollbrecht, line 1219: did it turned out to not be necessary ?

ivmarkov: No idea for now :) that's why I've kept the comment for now. Hopefully, once the esp modem is relatively complete, we'll know...

src/netif.rs
Vollbrecht: Ah ok the code example made it more clear. So we essentially use just some of the NetifConfiguration parameters here so the user dont need to provide the NonZerou32 from thin air.

Wouldn't it in that case be a bit more nicer to use for somebody who just look at this, if we accept for example an &NetifConfiguration and internally extract the id? 
 
if we in the example end up promoting to passing it from a netifconfiguration, why not clearly communicate it in the methods argument type?

ivmarkov: Even passing `NetifConfiguration` is dirty, because you might pass one netif configuration to the `EspNetif` you are wrapping, while another - to the `EspNetifDriver` that wraps the netif.

Look now. The got_ip/lost_ip are extracted **from** the `EspNetif` now. I.e. they are the same that are used for configuring the netif, as it also needs to know about those :)
`EspNetif` is no longer just a handle, but how many netifs is one going to have? Definitely not a thousand for this to be a concern.

Vollbrecht (approved):
Thanks for working on it! Looks good now. 

@DaneSlattery since you started the complete PPP inclusion feel free to also have a look.

ivmarkov:
> Thanks for working on it! Looks good now.
> 
> @DaneSlattery since you started the complete PPP inclusion feel free to also have a look.

Dane is in the root of this PR. :) My changes are on top of his ones (PPP events mapping and non-static netif driver).

I'll merge. I think the new `EspNetifDriver` is strictly more powerful than the previous version, so I don't anticipate API drama. If there are any bugs, we'll fix later.

PR #478, opened 2024-09-03
Added esp_wifi_sta_get_rssi function in EspWifi

MartinBroers:
I have added the function `esp_wifi_sta_get_rssi` in the Wifi trait. Nothing fancy and works well when connected. I can rename the function if that works better.

### DIFF
--- a/src/wifi.rs
+++ b/src/wifi.rs
@@ -1322,6 +1322,14 @@ impl<'d> WifiDriver<'d> {
             tx_status,
         );
     }
+
+    pub fn get_rssi(&self) -> Result<i32, EspError> {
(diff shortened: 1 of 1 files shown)


ivmarkov (changes requested):
Can you add the same to `WifiDriver`? Then you can just delegate from `EspWifi::get_rssi` to the driver impl.

ivmarkov:
...and this is not added to the Wifi trait. But let's not do this for now anyway.

MartinBroers:
I've added it to `WifiDriver`. Hope this is the way you meant it?

ivmarkov:
Thanks!

PR #479, opened 2024-09-04
Fix NVS example logging incorrect variable

cadeo111:
Fix incorrect key variable being logged for the struct storage for raw NVS access example


Vollbrecht (commented):
Thanks!

PR #482, opened 2024-09-06
Expose esp_mqtt_client_set_uri. Fix issue #481.

### LINKED ISSUE #419: MQTT on esp32c6 (and probably esp32c3) different behaviour than esp32
This is more of a note for other that stumble into this and hopefully saves them a few hours of debugging. (If there is a better place for this, I will gladly move it)

Testing around with the ESP32C6 I noticed that the mqtt client behaves way different.

First of all a (currently) non recoverable panic in the main loop happens, if you try to subscribe to any topic, before the connected callback is received (I do not know if this was ever valid, but it worked):

```
let mut client =  EspMqttClient::new_cb(&config.mqtt_url, &mqtt_client_config, move |event| {yadda yadda}
client.subscribe(stay_alive_topic.as_str(), ExactlyOnce)?;

```
If you do this you get a:
```
Guru Meditation Error: Core  0 panic'ed (Load access fault). Exception was unhandled.

Core  0 register dump:
MEPC    : 0x42072eca  RA      : 0x42072ec6  SP      : 0x40839a50  GP      : 0x408173e4  
0x42072eca - run_event_loop
    at /home/empire/workspace/PlantCtrl/rust/.embuild/espressif/esp-idf/v5.2.1/components/mqtt/esp-mqtt/mqtt_client.c:1539
0x42072ec6 - run_event_loop
    at /home/empire/workspace/PlantCtrl/rust/.embuild/espressif/esp-idf/v5.2.1/components/mqtt/esp-mqtt/mqtt_client.c:1539
0x408173e4 - __global_pointer$
    at ??:??
```

the same code works on a esp32 with the xtensa toolchain, so apparently there is some difference in the blocking/async behaviour of the idf core here. 
While I would prefer the subscribe return a error result, I'm not sure if this is even possible (or a good idea for that) without storing the connection state from the callbacks somewhere in the idf wrapper.

As a **workaround** waiting for the connected event in the callback, and only subscribing after it works fine on the C6.

Additionally, since the naming is confusing. The boolean connected in the embedded_svc::mqtt::client::EventPayload::Connected(connected)
actually represents session_present. This flag serves as an indicator for the client, informing it whether a previously established session is still available on the broker.  -> Does it make sense to rename this to session_present? If so I would offer to create a pull request for it.

### LINKED ISSUE #481: Dropping EspAsyncMqttConnection results in an assertion failure.
Dropping EspAsyncMqttConnection results in an assertion failure in `.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ffi/c_str.rs:417:13`.

The chip used here is esp32c6.
The esp-idf-svc version used is: "0.49" with features: embassy-sync, std, embassy-time-driver

The same example using the regular blocking mqtt client and connection works just fine.
Dropping client first, connection second doesn't work either.
Here is a minimal example to reproduce the error.

```
fn main() {
    // It is necessary to call this function once. Otherwise some patches to the runtime
    // implemented by esp-idf-sys might not link properly. See https://github.com/esp-rs/esp-idf-template/issues/71
    esp_idf_svc::sys::link_patches();

    // Bind the log crate to the ESP Logging facilities
    esp_idf_svc::log::EspLogger::initialize_default();

    let peripherals = Peripherals::take().unwrap();
    let sysloop = EspSystemEventLoop::take().unwrap();
    let nvs = EspDefaultNvsPartition::take().unwrap();

    let timer = EspTaskTimerService::new().unwrap();
    let mut ticker = timer.timer_async().unwrap();

    let wifi = EspWifi::new(peripherals.modem, sysloop.clone(), Some(nvs.clone())).unwrap();
    let wifi = BlockingWifi::wrap(wifi, sysloop).unwrap();


    let (client, connection) = EspAsyncMqttClient::new("mqtt://192.168.178.119:1883", &Default::default()).unwrap();
    drop(connection);
}
```

JakobLachermeier:
Exposes `esp_mqtt_client_set_uri` in both blocking and async clients.
Fixes #481 by moving unsafe { core::ffi::CStr::from_bytes_with_nul_unchecked(&work.topic) } into  their respective commands. I'm not sure why this works.
Doesn't fix #419. I'm also not sure why the original code works on xtensa but not riscv.

### DIFF
--- a/src/mqtt/client.rs
+++ b/src/mqtt/client.rs
@@ -622,13 +622,7 @@ impl<'a> EspMqttClient<'a> {
     }
 
     pub fn set_uri_cstr(&mut self, uri: &core::ffi::CStr) -> Result<MessageId, EspError> {
-        Self::check(
-            unsafe {
-                esp_mqtt_client_set_uri(
-                    self.raw_client,
-                    uri.as_ptr())
-            }
-        )
+        Self::check(unsafe { esp_mqtt_client_set_uri(self.raw_client, uri.as_ptr()) })
     }
 
     extern "C" fn handle(
@@ -748,7 +742,7 @@ struct AsyncWork {
     topic: alloc::vec::Vec<u8>,
     payload: alloc::vec::Vec<u8>,
     result: Result<MessageId, EspError>,
-    broker_uri: alloc::vec::Vec<u8>
+    broker_uri: alloc::vec::Vec<u8>,
 }
 
 pub struct EspAsyncMqttClient(Unblocker<AsyncWork>);
@@ -805,18 +799,14 @@ impl EspAsyncMqttClient {
             AsyncCommand::Publish { qos, retain },
             Some(topic),
             Some(payload),
-            None
+            None,
         )
         .await
     }
 
     pub async fn set_uri(&mut self, broker_uri: &str) -> Result<MessageId, EspError> {
-        self.execute(
-            AsyncCommand::SetUri,
-            None,
-            None,
-            Some(broker_uri)
-        ).await
+        self.execute(AsyncCommand::SetUri, None, None, Some(broker_uri))
+            .await
     }
 
     async fn execute(
@@ -865,7 +855,7 @@ impl EspAsyncMqttClient {
             topic: alloc::vec::Vec::new(),
             payload: alloc::vec::Vec::new(),
             result: Ok(0),
-            broker_uri: alloc::vec::Vec::new()
+            broker_uri: alloc::vec::Vec::new(),
         };
         // Repeatedly share a reference to the work until the channel is closed.
         // The receiver will replace the data with the next work item, then wait for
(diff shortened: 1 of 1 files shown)


ivmarkov:
> Fixes #481 by moving unsafe { core::ffi::CStr::from_bytes_with_nul_unchecked(&work.topic) } into their respective commands. I'm not sure why this works. Doesn't fix #419. I'm also not sure why the original code works on xtensa but not riscv.

I'm also not sure why this works if it works. It might just be masking out a deeper problem we are not seeing, as moving the topic unwrapping inside the legs of the `match` is equivalent to the previous code (I mean, if we put aside your new `SetUri` branch which does not consume the topic and where the value of the topic might be invalid of course).

With that said, I'm OK to merge this. When I have to change the URI, I tend to re-create the whole MQTT client and connection, but I guess having `set_uri` might be more ergonomic for certain use cases.

PR #487, opened 2024-09-19
ESP IDF 5.3 compatibility

ivmarkov:


ivmarkov:
@Vollbrecht Sorry for skipping your code review - the PR looks bigger than it is, but in reality it just cfgs-out some BT initialization settings (UART?) which are no longer there on the c6 with ESP IDF >= 5.3

PR #488, opened 2024-09-19
Add poll_read/write and implement futures_io::AsyncRead/Write for EspAsyncTls

jothan:
This integrates EspAsyncTls into the wider async ecosystem and should reduce the need for glue code.

In the case that glue code is needed, the poll_read and poll_write functions can be easily used to integrate with another trait.


ivmarkov:
Yep, why not, especially given that `futures-io` is an optional dependency.

Though if we do this here, I wonder whether we should also do it in the `hal` crate, where we do implement `embedded_io_async::Read/Write` for certain drivers?...

PR #495, opened 2024-09-29
Expose netif DNS setting functions

jothan:
I'm using this to make a captive portal using the DNS method (edge_captive made it really simple).

### DIFF
--- a/src/netif.rs
+++ b/src/netif.rs
@@ -479,7 +479,7 @@ impl EspNetif {
         }
     }
 
-    fn set_dns(&mut self, dns: ipv4::Ipv4Addr) {
+    pub fn set_dns(&mut self, dns: ipv4::Ipv4Addr) {
         let mut dns_info: esp_netif_dns_info_t = Default::default();
 
         unsafe {
@@ -509,7 +509,7 @@ impl EspNetif {
         }
     }
 
-    fn set_secondary_dns(&mut self, secondary_dns: ipv4::Ipv4Addr) {
+    pub fn set_secondary_dns(&mut self, secondary_dns: ipv4::Ipv4Addr) {
         let mut dns_info: esp_netif_dns_info_t = Default::default();
 
         unsafe {
--- a/src/netif.rs
+++ b/src/netif.rs
@@ -531,7 +531,7 @@ impl EspNetif {
         Ok(unsafe { from_cstr_ptr(ptr) }.try_into().unwrap())
     }
 
-    fn set_hostname(&mut self, hostname: &str) -> Result<(), EspError> {
+    pub fn set_hostname(&mut self, hostname: &str) -> Result<(), EspError> {
         let hostname = to_cstring_arg(hostname)?;
 
         esp!(unsafe { esp_netif_set_hostname(self.handle, hostname.as_ptr() as *const _) })?;


ivmarkov:
@jothan Just to make sure I understand the need for this:

In general, the `EspNetif` API was created with immutability in mind first (though we are slowly drifting away from that).
What I mean by this, is that you choose - upfront - prior to the creation of the netif, what configuration it is going to have, by using a `NetifConfiguration` struct.

This struct has a variant, `Router(RouterConfiguration)` that has the DNS fields you are trying to set. While it does not have the `hostname` field, I'm actually not even sure why you need it, given that `hostname` is only useful in the presence of DHCP, and only for **clients**, which want to register themselves in the DHCP server under that hostname, if the DHCP server happens to do local DNS resolution too.

So, would you elaborate why do you need to have these APIs as public in the first place?

jothan:
I want the DHCP client hostname to contain the device name and a part of the serial number.

For the DNS, it is my understanding that the primary DNS of the AP interface is sent out by the DHCP server to clients to use as their DNS. I want the clients to use the ESP32 as DNS server to redirect them via captive portal techniques to the configuration page.

Hope that explains my use cases.

jothan:
Also, the netif is created by the Wi-Fi calls. Should I create the interface configuration I want and hot-swap the netif in ?

ivmarkov:
> Hope that explains my use cases.

It does. However it does not explain, why you need mutating operations to achieve this. Namely:

> I want the DHCP client hostname to contain the device name and a part of the serial number.

Sure. Just use a `NetifConfiguration` instance where `ip_configuration` is set to `Some(ipv4::Configuration::Client(ClientConfiguration::Dhcp(DhcpClientSettings { hostname: <your-host-name>})))`?

Note that the hostname you are setting on the netif _anyway_ only has any use when you use a DHCP confdiguration, and in no other cases. I.e. using a statically configured IP and hoping that somehow your esp client will be pingable by its hostname is just not possible. For such cases, the only solution is `mDNS` (but then you don't need the router to serve as a DNS proxy anyway).


> For the DNS, it is my understanding that the primary DNS of the AP interface is sent out by the DHCP server to clients to use as their DNS. I want the clients to use the ESP32 as DNS server to redirect them via captive portal techniques to the configuration page.

Assuming your clients use DHCP, they will use whatever DNS the DHCP server sends them. So you need to configure the DNS IPs _on the router_. Wouldn't a router `ip_configuration` set to `Some(ipv4::Configuration::Router(RouterConfiguration { dns: Some(<your-router-ip>), ...}))` work?

ivmarkov:
> Also, the netif is created by the Wi-Fi calls. Should I create the interface configuration I want and hot-swap the netif in ?

Hot-swapping is one option, but we also have `EspWifi::wrap_all` for that purpose.

ivmarkov:
> Is there a way to grab the STA mac address before using the wrap_all call ?

Can't recall off the top of my head, but for setting the MAC I think we have a mutating method on `EspNetif`? (As I said, we are slowly drifting towards mutation.)

Vollbrecht:
> Is there a way to grab the STA mac address before using the wrap_all call ?

If you know the base mac address, its always the same [schema](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/misc_system_api.html#mac-address). STA is always +0 offset.
You can get it via the esp_read_mac() function

jothan:
Ok, so I'm getting the default AP netif config like this:

```rust
let mut ap_cfg = NetifConfiguration::wifi_default_router();
```

Nowhere in that struct do I have a way to get the IP that will be assigned to the ESP32 SoftAp. I would like to fill https://docs.esp-rs.org/esp-idf-svc/esp_idf_svc/ipv4/struct.RouterConfiguration.html with this IP.

jothan:
Ok, so I managed to do what I need to do using the mainline esp-idf-svc. I did the manual chain with configuration of going from Driver to AsyncWifi, which is non-obvious but got me what I needed.

I also managed to grab the STA MAC from WifiDriver in that chain.

Should we close this PR then ?

ivmarkov:
> Ok, so I managed to do what I need to do using the mainline esp-idf-svc. I did the manual chain with configuration of going from Driver to AsyncWifi, which is non-obvious but got me what I needed.
> 
> I also managed to grab the STA MAC from WifiDriver in that chain.
> 
> Should we close this PR then ?

Let me think about this overnight. Not sure - would it work in all cases?

PR #498, opened 2024-10-10
Support for LittleFS

ivmarkov:
This is the last filesystem of the three supported on the ESP-IDF via the VFS API.

Similarly to our OneWire support and mDNS support, it uses an external component from the ESP-IDF Manager.

While the component is not in the Espressif namespace, apparently it is considered stable enough and supported enough to be used, given that [the official ESP IDF docu directly documents it](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-guides/file-system-considerations.html#littlefs).

* W.r.t. NOR flash, this seems to be the recommended choice given that Spiffs is not supported anymore, and FATFS is unreliable on power loss (modulo NVS of course)
* W.r.t. SD cards, it does support those, with the caveat that a max of 4GB from the SD card can be used, as everything in LittleFs seems to be limited to 32bits. :-)


src/io.rs
```diff
@@ -143,4 +143,119 @@ pub mod vfs {
...
+
+            sys::esp!(unsafe { sys::esp_vfs_littlefs_register(&conf) })?;
+
+            Ok(Self {
+                _littlefs: littlefs,
+                partition_raw_data,
+            })
+        }
+
+        pub fn info(&mut self) -> Result<crate::fs::littlefs::LittleFsInfo, sys::EspError> {
```
Vollbrecht, line 202: nit: this could just be a &self or?

Vollbrecht (approved):
Nice work, looks good overall.

I have just two small questions though 

I assume it should also work if we use the SdMmcHostDriver on targets that support it? Should we than maybe add a comment in the example so people notice that there is a option here as we don't promote it in any of the examples? 

Should we maybe group all FS related examples in there own directory so its simpler to find them since slowly the number of example grows?

src/io.rs
ivmarkov: Why not. Fixed.

ivmarkov:
> Nice work, looks good overall.
> 
> I have just two small questions though
> 
> I assume it should also work if we use the SdMmcHostDriver on targets that support it? Should we than maybe add a comment in the example so people notice that there is a option here as we don't promote it in any of the examples?

Yes it should be well possible. (Too lazy to test that, but the API is anyway identical as the difference is only in how the `SdCard` driver is instantiated. And - FATFS was tested with both SPI and SDMMC).

> Should we maybe group all FS related examples in there own directory so its simpler to find them since slowly the number of example grows?

Let's postpone for another time? 
If we do this grouping, we should rather do it everywhere, i.e. for the "wifi", "http server", http client", "thread", all examples in the "hal" and so on. Perhaps we can do it once/if the examples have their own `Cargo.toml` and `sdkconfig.defaults` conf.

PR #500, opened 2024-10-14
Document OTA API

armandas:
I've been recently implementing OTA for my project and noticed the documentation was lacking. I decided to add documentation to improve my own understanding and hopefully make it more accessible to others.

I used [ESP-IDF OTA documentation](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/ota.html) for reference.

As I'm not an expert here, I would appreciate if someone could check the correctness, especially the module doc.

Additionally, I'm not sure about the contents of `EspOtaUpdate::flush()`. On one hand, this function does nothing and the user may want to know about it, on the other hand, I just exposed implementation details. Should the documentation be written with the assumption that this function does something? This would encourage users to use it and ensure compatibility with future updates.

### DIFF
--- a/src/ota.rs
+++ b/src/ota.rs
@@ -3,6 +3,39 @@
 //! The OTA update mechanism allows a device to update itself based on data
 //! received while the normal firmware is running (for example, over Wi-Fi or
 //! Bluetooth.)
+//!
+//! # Examples
+//!
+//! The following example shows approximate steps for performing an OTA update.
+//!
+//! ```
+//! // 1. Obtain an instance of OTA:
+//! let mut ota = EspOta::new().expect("obtain OTA instance");
+//!
+//! // 2. Initiate update and obtain an instance of `EspOtaUpdate`:
+//! let mut update = ota.initiate_update().expect("initiate OTA");
+//!
+//! // 3. Write the program data:
+//! while let Some(data) = my_wireless.get_ota_data() {
+//!     update.write(&data).expect("write OTA data");
+//! }
+//!
+//! // 4. Finalize update:
+//! update.complete().expect("complete OTA");
+//!
+//! // 5. Reboot:
+//! esp_idf_svc::hal::reset::restart();
+//! ```
+//! After rebooting and confirming that the new firmware works, mark it as valid.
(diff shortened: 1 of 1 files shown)


src/ota.rs
```diff
@@ -183,8 +238,8 @@ impl<'a> EspOtaUpdate<'a> {
         Ok(())
     }
 
+    /// No-op: the OTA update is aborted when `EspOtaUpdate` is dropped.
```
DaneSlattery, line 241: This consumes `self` without borrowing, forcing `EspOtaUpdate` to be dropped when the function is called.

The comment might be better as `Cancel the update`

src/ota.rs
```diff
@@ -218,6 +273,7 @@ pub struct EspOtaUpdateFinished<'a> {
 }
 
 impl EspOtaUpdateFinished<'_> {
+    /// Sets the boot partition to the newly updated app partition.
```
DaneSlattery, line 276: Note that the user must still reboot for the updated app partition to go into effect

src/ota.rs
```diff
@@ -239,6 +300,11 @@ impl EspOta {
         Ok(Self(()))
     }
 
+    /// Returns the currently configured boot slot.
```
DaneSlattery, line 303: boot slot that will go into effect on the next reboot.

DaneSlattery (commented):
This is all I could think of, looks good.

DaneSlattery:
This is for your perusal as well in case you wanted a more in depth example https://github.com/DaneSlattery/esp-ota

src/ota.rs
armandas: Thanks. I feel like this is already implied, i.e. "boot slot" will obviously be used for the next boot. Or am I missing something?

src/ota.rs
armandas: I added "The app will be run on the next boot.". Does that sound OK?

src/ota.rs
DaneSlattery: No I think you're right

ivmarkov:
Thanks!

PR #502, opened 2024-10-17
Change default eth key

DaneSlattery:
According to this issue, https://github.com/espressif/esp-idf/issues/10359

The default name for an ethernet netif is `ETH_DEF`, but it is hardcoded as `ETH_CL_DEF`. The mDNS service attempts to look up the netif based on the name during initialisation, but it won't find it unless the user wraps their own netif with the new name.

### DIFF
--- a/src/netif.rs
+++ b/src/netif.rs
@@ -157,7 +157,7 @@ impl NetifConfiguration {
                 | esp_netif_flags_ESP_NETIF_FLAG_EVENT_IP_MODIFIED,
             got_ip_event_id: NonZeroU32::new(ip_event_t_IP_EVENT_ETH_GOT_IP as _),
             lost_ip_event_id: NonZeroU32::new(ip_event_t_IP_EVENT_ETH_LOST_IP as _),
-            key: "ETH_CL_DEF".try_into().unwrap(),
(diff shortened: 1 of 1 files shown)


ivmarkov:
Hmmmm, what about WIFI? Is there any "default" key which is expected there? We currently have "WIFI_STA_DEF" and "WIFI_AP_DEF". Are these correct?

What about the so called "router" Eth, which is nothing else but a fixed-conf Eth?

DaneSlattery:
Looking at the code here,
https://github.com/espressif/esp-protocols/blob/85202456578019689f8d3527042df5462da010f9/components/mdns/mdns.c#L151 it seems like the WIFI names are good defaults.

The Ethernet router, PPP or thread interfaces don't seem to be considered in this case.

ivmarkov:
> What about the so called "router" Eth, which is nothing else but a fixed-conf Eth?

What about this? I think you should change it there as well.

DaneSlattery:
It doesn't seem to interact with the DNS system at all, so I'm not sure what it should be. The `ETH_RT_DEF` name seems fine to me

ivmarkov:
OK fine, I'll merge as is, and will change the RT key if necessary later.

PR #521, opened 2024-11-22
Fix missing newline if CONFIG_LOG_COLORS=n is set

### LINKED ISSUE #520: EspLogger::log without colors (CONFIG_LOG_COLORS=n) misses newline
Since this commit #494, when you write a log item without colors, a newline at the end is missing. Specifically because of this commit: https://github.com/esp-rs/esp-idf-svc/commit/74c7a27af7afa5477405a90259172681c05ddec9

Now if the color variable is None, writeln! is never being called. Is it as simple as just ```else { writeln!("")}```?

mike1703:
Fixes #520 where no newlines were printed if `CONFIG_LOG_COLORS=n` as the newline was only printed if `color.is_some()`.
Now the `writeln!()` call is at the end and is always called.

### DIFF
--- a/src/log.rs
+++ b/src/log.rs
@@ -272,8 +272,9 @@ impl ::log::Log for EspLogger {
             }
             write!(stdout, ") {}: {}", target, args).unwrap();
             if color.is_some() {
-                writeln!(stdout, "\x1b[0m").unwrap();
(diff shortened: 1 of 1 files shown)


ivmarkov:
Thanks!

PR #525, opened 2024-12-09
Expose src_addr and des_addr in espnow recv cb

julienvincent:
Currently the espnow bindings for `register_recv_cb` only give a handle on the src_addr from the sending peer. This makes it impossible to determine if the frame was sent on the BROADCAST address.

This change updates the first argument to the receive cb to accept a new `ReceiveInfo` struct containing `src_addr` and `dst_addr`.

This allows writing code along the lines of:

```rust
controller.register_recv_cb(|info, data| {
  if info.dst_addr == &espnow::BROADCAST {
    log::info!("Received broadcast!");
    return;
  }

  // do something else with 'normal' packet.
})?;
```

### DIFF
--- a/src/espnow.rs
+++ b/src/espnow.rs
@@ -20,8 +20,14 @@ type Singleton<T> = Mutex<Option<Box<T>>>;
 
 pub const BROADCAST: [u8; 6] = [0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF];
 
+#[derive(Debug, Clone)]
+pub struct ReceiveInfo<'a> {
+    pub src_addr: &'a [u8; 6],
+    pub dst_addr: &'a [u8; 6],
+}
+
 #[allow(clippy::type_complexity)]
-static RECV_CALLBACK: Singleton<dyn FnMut(&[u8], &[u8]) + Send + 'static> = Mutex::new(None);
+static RECV_CALLBACK: Singleton<dyn FnMut(&ReceiveInfo, &[u8]) + Send + 'static> = Mutex::new(None);
 #[allow(clippy::type_complexity)]
 static SEND_CALLBACK: Singleton<dyn FnMut(&[u8], SendStatus) + Send + 'static> = Mutex::new(None);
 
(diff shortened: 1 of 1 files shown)


ivmarkov:
Can you fix the build for 4.4.x which we (still) support?
If your functionality is impossible to expose for 4.4.x, perhaps you can only do it for 5.x?

julienvincent:
Ah, I didn't spot that this was setup differently for v4 and v5. Would it be acceptable to accept the old `&[u8]` for v4 and the new `ReceiveInfo` for v5?

julienvincent:
@ivmarkov Ok this is compiling now and works locally for me on v5.

I am passing a `ReceiveInfo` to the callback for both v4 and v5 but the `des_addr` will be a initialised slice when compiled for v4.

ivmarkov:
> @ivmarkov Ok this is compiling now and works locally for me on v5.
> 
> I am passing a `ReceiveInfo` to the callback for both v4 and v5 but the `des_addr` will be a `0` initialised slice when compiled for v4.

In `ReceiveInfo`, why not:
```rust
pub des_addr: Option<&'a [u8; 6]>,
```
?

ivmarkov:
Also, `des_addr` sounds a bit weird. Shouldn't it be `dst_addr`?

julienvincent:
> > @ivmarkov Ok this is compiling now and works locally for me on v5.
> > I am passing a `ReceiveInfo` to the callback for both v4 and v5 but the `des_addr` will be a `0` initialised slice when compiled for v4.
> 
> In `ReceiveInfo`, why not:
> 
> ```rust
> pub des_addr: Option<&'a [u8; 6]>,
> ```
> 
> ?

It seems a bit weird to me to do this just to support an old version of IDF. I assume that v4 will be deprecated at some point and then this artifact would remain behind. It would also make the usage in v5 worse - where this would always be a `Some`.

I'd be happy to make the change though. Up to you!

---

> Also, `des_addr` sounds a bit weird. Shouldn't it be `dst_addr`?

I was just copying the terminology from esp-idf verbatim (https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/network/esp_now.html#_CPPv417esp_now_recv_info) but I don't have an opinion. Happy to change this to `dst_addr` if you prefer.

ivmarkov:
> It seems a bit weird to me to do this just to support an old version of IDF. I assume that v4 will be deprecated at some point and then this artifact would remain behind. It would also make the usage in v5 worse - where this would always be a `Some`.

Fair enough, leave as is.

> > Also, `des_addr` sounds a bit weird. Shouldn't it be `dst_addr`?
> 
> I was just copying the terminology from esp-idf verbatim (https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/network/esp_now.html#_CPPv417esp_now_recv_info) but I don't have an opinion. Happy to change this to `dst_addr` if you prefer.

Sure but maybe we should not copy their mistakes. Either `sor` and `des` or (better) `src` and `dst`.

julienvincent:
@ivmarkov Ok, done 👍🏻

ivmarkov:
Thanks!

PR #526, opened 2024-12-10
Allow setting skip_unhandled_events on esp timer

julienvincent:
Currently the esp timer service hardcodes the `skip_unhandled_events` flag to `false` which means there is no way to construct a timer that doesn't wake the device from light sleep.

This change introduces a `skip_unhandled_events` fn on the `EspTimerService` which allows changing the value of this timer flag.

---

I added this as a mutating fn on the timer service - but I wonder if it's a better design to make the `new` constructor accept this flag. I didn't want to change it's signature though to avoid introducing a breaking change.

### DIFF
--- a/src/timer.rs
+++ b/src/timer.rs
@@ -207,26 +207,24 @@ where
     where
         F: FnMut() + Send + 'static,
     {
-        self.internal_timer(callback)
+        self.internal_timer(callback, false)
+    }
+
+    /// Same as `timer` but does not wake the device from light sleep.
+    pub fn timer_nowake<F>(&self, callback: F) -> Result<EspTimer<'static>, EspError>
+    where
+        F: FnMut() + Send + 'static,
+    {
+        self.internal_timer(callback, true)
     }
 
     pub fn timer_async(&self) -> Result<EspAsyncTimer, EspError> {
-        let notification = Arc::new(Notification::new());
+        self.internal_timer_async(false)
+    }
 
-        let timer = {
-            let notification = Arc::downgrade(&notification);
-
-            self.timer(move || {
-                if let Some(notification) = notification.upgrade() {
-                    notification.notify(NonZeroU32::new(1).unwrap());
-                }
-            })?
-        };
-
-        Ok(EspAsyncTimer {
-            timer,
-            notification,
-        })
+    /// Same as `timer_async` but does not wake the device from light sleep.
+    pub fn timer_async_nowake(&self) -> Result<EspAsyncTimer, EspError> {
+        self.internal_timer_async(true)
     }
 
     /// # Safety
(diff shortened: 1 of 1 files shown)


ivmarkov:
It is desirable to keep backwards comkpatibility in `EspTimerService`. While we broke backwards compatibility just a couple of days ago in `EspNow`, `EspNow` is a bit of a niche, leaf-only code, while `EspTimerService` is used everywhere.

With that said, I don't think it is ideal to have the newly introduced flag on the global `EspTimerService` level. Isn't it better to introduce it (in the form of new functions) when you are creating a timer instead? I.e. with extra, new `timer_*` methods?

julienvincent:
> With that said, I don't think it is ideal to have the newly introduced flag on the global `EspTimerService` level. Isn't it better to introduce it (in the form of new functions) when you are creating a timer instead? I.e. with extra, new `timer_*` methods?

@ivmarkov Ok I tried this out but it just seems a bit messy to me. There then needs to be multiple new methods all doing the same thing sans flag.

See https://github.com/esp-rs/esp-idf-svc/commit/8af264d63506bb301f27b2916df71e3ca9802fb3 for WIP implementation of the above.

Is this what you had in mind?

ivmarkov:
> > With that said, I don't think it is ideal to have the newly introduced flag on the global `EspTimerService` level. Isn't it better to introduce it (in the form of new functions) when you are creating a timer instead? I.e. with extra, new `timer_*` methods?
> 
> @ivmarkov Ok I tried this out but it just seems a bit messy to me. There then needs to be multiple new methods all doing the same thing sans flag.
> 
> See [8af264d](https://github.com/esp-rs/esp-idf-svc/commit/8af264d63506bb301f27b2916df71e3ca9802fb3) for WIP implementation of the above.
> 
> Is this what you had in mind?

Yes, that's what I had in mind.
It might be a bigger change, but it is still might be a bit less risky than your original suggestion, in that while the `EspTimerService` is not technically a singleton, it is used as such, and it might become a singleton in future.

If that happens, then you have to decide  - for your whole app - if you want ALL timers to skip events or not skip events. Hence why I think it is better to have it on a per-timer level, which mostly follows the ESP IDF API/approach anyway.

If you could lower a bit the copy-pasted methods in your https://github.com/esp-rs/esp-idf-svc/commit/8af264d63506bb301f27b2916df71e3ca9802fb3 (perhaps, by introducing an extra `internal_*` method or two), then I think that would be the better suggestion, yes.

julienvincent:
Thanks for the explanation that all makes perfect sense.

I've made the requested changes, including deduplicating the `timer_async` implementation.

ivmarkov:
Thanks!

PR #531, opened 2024-12-22
OTA

ivmarkov:
Implements a new type `EspFirmwareInfoLoad` that has a reduced memory consumption, as per the docs.

### DIFF
--- a/src/ota.rs
+++ b/src/ota.rs
@@ -160,7 +160,7 @@ impl EspFirmwareInfoLoad {
     ///
     /// Returns `true` if the information was successfully fetched.
     /// Returns `false` if the firmware data has not been loaded completely yet.
-    pub fn fetch(&self, data: &[u8], info: &mut FirmwareInfo) -> Result<bool, EspError> {
+    pub fn fetch(&self, data: &[u8], info: &mut FirmwareInfo) -> Result<bool, EspIOError> {
         let loaded = data.len()
             >= mem::size_of::<esp_image_header_t>()
                 + mem::size_of::<esp_image_segment_header_t>()
@@ -225,6 +225,10 @@ impl EspFirmwareInfoLoad {
 
         Ok(())
     }
+}
+
+impl io::ErrorType for EspFirmwareInfoLoad {
+    type Error = EspIOError;
 }
 
 #[derive(Debug)]
--- a/src/ota.rs
+++ b/src/ota.rs
@@ -54,42 +54,14 @@ pub use embedded_svc::ota::{FirmwareInfo, LoadResult, Slot, SlotState, UpdatePro
 use crate::sys::*;
 
 use crate::io::EspIOError;
-use crate::private::{common::*, cstr::*, mutex};
+use crate::private::{cstr::*, mutex};
 
 static TAKEN: mutex::Mutex<bool> = mutex::Mutex::new(false);
 
-impl From<Newtype<&esp_app_desc_t>> for FirmwareInfo {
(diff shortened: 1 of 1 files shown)


src/ota.rs
```diff
@@ -163,6 +149,88 @@ impl FirmwareInfoLoader for EspFirmwareInfoLoader {
...
+///
+/// This is a more efficient version of the now-deprecated `EspFirmwareInfoLoader`.
+pub struct EspFirmwareInfoLoad;
+
+impl EspFirmwareInfoLoad {
+    /// Fetches firmware information from the firmware binary data chunk loaded so far.
+    ///
+    /// Returns `true` if the information was successfully fetched.
+    /// Returns `false` if the firmware data has not been loaded completely yet.
+    pub fn fetch(&self, data: &[u8], info: &mut FirmwareInfo) -> Result<bool, EspIOError> {
```
pbert519, line 163: @ivmarkov I'm currently trying to port the EspFirmwareInfoLoader to EspFirmwareInfoLoad. 
While i understand the performance changes, i have a few question about this changes.

Why is FirmwareInfo passed as mutable reference, in my point of view it would be much easier to just return FirmwareInfo instead of the bool.

Also is there any reason why fetch takes &self?

ivmarkov: > @ivmarkov I'm currently trying to port the EspFirmwareInfoLoader to EspFirmwareInfoLoad. While i understand the performance changes, i have a few question about this changes.
> 
> Why is FirmwareInfo passed as mutable reference, in my point of view it would be much easier to just return Return<Option, EspIOError> instead of the bool.

That's how it used to be in the previous loader. 

However, `FirmwareInfo` has a relatively large footprint of ~ 500 bytes. Why I designed it like this is another topic, but fixing `embedded-svc` (from where it comes) is not a topic right now. Therefore, it is not ideal if it is passed with moves as it occupies a large stack size then, especially in an async context, where futures' memory layout is not optimized w.r.t. input arguments (they are duplicated).

> Also is there any reason why fetch takes &self?

No particular reason, but doesn't hurt either, as `FirmwareInfoLoad` is currently an empty struct, so the compiler would optimize away the `&self` anyway.

pbert519: > That's how it used to be in the previous loader.

The performance optimatisations i noticed where that EspFirmwareInfoLoader did have a copy of the buffer, while EspFirmwareInfoLoad just works an a slice.
Good point with the copy on return, however for non async cases, the compiler should complety optimize that and the api is much nicer. It dosn't help that `Firmwareinfo` does not have a default implementation, but that's a `embedded_svc` topic.

> No particular reason, but doesn't hurt either, as FirmwareInfoLoad is currently an empty struct, so the compiler would optimize away the &self anyway.

Sure it doesn't hurt, but it's confusing for user of the API, at least thats my experience.

ivmarkov: > however for non async cases, the compiler should complety optimize that and the api is much nicer

I wish I was also as confident as you are in that. Look [here](https://github.com/rust-lang/rfcs/pull/2884).

> 
> > No particular reason, but doesn't hurt either, as FirmwareInfoLoad is currently an empty struct, so the compiler would optimize away the &self anyway.
> 
> Sure it doesn't hurt, but it's confusing for user of the API, at least thats my experience.

If nothing else, these changes were already released, and I don't think the API breakage of removing `&self` is worth it.

pbert519: Thanks for the insights.
I agree that a new release is not worth changing the &self.

Issue #246, opened 2023-03-29
Missing wraper for promiscous mode.


Frostie314159:
The enablement of promiscous mode isn't wrapped in a safe function.

Vollbrecht:
do you want to create an PR for it ?

Frostie314159:
Of course. Btw, ever noticed that the promisc mode described in the docs is actually monitor mode?

Vollbrecht:
oh that sounds interesting. i never played with both before on esp, still on my todo list. does the esp driver can sniff on other packets while being an ap itself and send data to clients or does it come with some limitations?

owenthewizard:
Promiscuous mode for Eth would be nice too.

PR #532, opened 2024-12-23
Safe wrapper for eth promiscuous mode

indexds:
Hello, this is my attempt at #246. This works locally, I could add wrappers for wifi as well for the sake of symmetry, though I have no use for them.

lmk

### DIFF
--- a/src/eth.rs
+++ b/src/eth.rs
@@ -1094,7 +1094,7 @@ impl<'d, T> EthDriver<'d, T> {
     fn eth_mac_default_config(_mdc: i32, _mdio: i32) -> eth_mac_config_t {
         eth_mac_config_t {
             sw_reset_timeout_ms: 100,
-            rx_task_stack_size: 2048,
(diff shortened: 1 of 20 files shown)


src/eth.rs
```diff
@@ -1032,6 +1032,24 @@ impl<'d, T> EthDriver<'d, T> {
         Ok(())
     }
 
+    pub fn set_promiscuous(&self, state: bool) -> Result<(), EspError> {
```
ivmarkov, line 1035: 1. Put some documentation above the method. 
2. Do not take &self but &mut self.

ivmarkov (changes requested):

indexds:
This should be fine I think. Want me to make a wrapper for wifi as well?

ivmarkov:
> This should be fine I think. Want me to make a wrapper for wifi as well?

If the Wifi driver actually does have such a mode and you could expose it that would be great.

PR #534, opened 2024-12-24
Exposes core_id in the configuration of an HttpServer

Angelo13C:


src/http/server.rs
```diff
@@ -79,6 +79,7 @@ pub struct Configuration {
     pub http_port: u16,
     pub ctrl_port: u16,
     pub https_port: u16,
+    pub core_id: i32,
```
ivmarkov, line 82: `i32` is not ideal. Use [this](https://github.com/esp-rs/esp-idf-hal/blob/master/src/cpu.rs#L13) instead.

ivmarkov (changes requested):

ivmarkov:
Also if you could also update the Changelog as the new PR template suggests? In section Breaking, as you are adding a new field to an existing struct which is not marked with non_exhaustive.

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Change default eth key (#502)
 - ESP IDF Partitions API (#511)
 - Expose src_addr and dst_addr in espnow recv cb (#525)
+- Add `core` field to `http::server::Configuration` to control which CPU core runs the HTTP server task
```
indexds, line 20: You're in the wrong place, 0.50.0 is already released, add a new ## [Unreleased] tag above

src/http/server.rs
```diff
@@ -58,6 +58,8 @@ use embedded_svc::http::headers::content_type;
 use embedded_svc::http::*;
 use embedded_svc::io::{ErrorType, Read, Write};
 
+use esp_idf_hal::cpu::Core;
```
indexds, line 61: you'll want to do this instead since it's been re-exported in the crate
```rust
use crate::hal::cpu::Core; 
```

src/http/server.rs
```diff
@@ -141,7 +145,7 @@ impl From<&Configuration> for Newtype<httpd_config_t> {
             ))]
             task_caps: (MALLOC_CAP_INTERNAL | MALLOC_CAP_8BIT),
             stack_size: conf.stack_size,
-            core_id: i32::MAX,
+            core_id: conf.core.map(|core| core.into()).unwrap_or(i32::MAX),
```
indexds, line 148: This I'm not sure about because I don't know much about all this, but this should be a core still, no? Shouldn't you do something like this?

```rust
conf.core.map(|core| core.into()).unwrap_or(Core::Core0)
```
I'm not sure but it feels like i32::MAX would always crash.

ivmarkov:
This PR looks OK now, but merging it has to wait a bit, because adding a new field to the conf struct is backwards-incompatible (strictly speaking).

Fortunately, it is only a matter of a few days. Turns out, we need to release a new non-patch `esp-idf-svc` any way really soon, because of the new `embassy-time-driver` API which is also not backwards-compatible.

src/http/server.rs
indexds: Right, I checked, this is because of this definition

```c
#define tskNO_AFFINITY      ( ( BaseType_t ) 0x7FFFFFFF )
```

So probably add a comment that says i32::MAX is to say no core affinity?

ivmarkov:
It seems we are heading for another minor (= backwards-incompatible) release of `esp-idf-svc` so I would merge that.

Thanks!

PR #545, opened 2025-01-02
Fix crash when enabling napt

indexds:

### Submission Checklist 📝

#### Description
something something tcpip context else i crash.

I added some documentation as well. Really a paste from the wiki. There might be a way to do this cleaner but I could not for the life of me figure it out. The enable function takes two arguments so I can't really pass them directly I have to do this whole spaghetti thing. Let me know if you find an easier way i'll modify it.

#### Testing
It's not crashing.

### DIFF
--- a/src/netif.rs
+++ b/src/netif.rs
@@ -599,18 +599,16 @@ impl EspNetif {
         Ok(())
     }
 
+    #[cfg(esp_idf_lwip_ipv4_napt)]
     unsafe extern "C" fn napt_wrapper(ctx: *mut ffi::c_void) {
-        #[cfg(esp_idf_lwip_ipv4_napt)]
-        {
-            let ctx = unsafe { *(ctx as *mut (u8, i32)) };
+        let ctx = unsafe { *(ctx as *mut (u8, i32)) };
 
-            ip_napt_enable_no(ctx.0, ctx.1);
-        }
+        ip_napt_enable_no(ctx.0, ctx.1);
     }
 
     /// Enables or disables NAPT on this netif.
     ///
-    /// Enable operation can be performed only on one interface at a time.
+    /// Enable operation can be performed only on one interface at a time. 
     /// NAPT cannot be enabled on multiple interfaces according to this implementation.
     #[cfg(esp_idf_lwip_ipv4_napt)]
     pub fn enable_napt(&mut self, enable: bool) -> Result<(), EspError> {
--- a/src/netif.rs
+++ b/src/netif.rs
@@ -608,7 +608,7 @@ impl EspNetif {
 
     /// Enables or disables NAPT on this netif.
     ///
-    /// Enable operation can be performed only on one interface at a time. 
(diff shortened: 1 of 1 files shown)


ivmarkov:
Can you paste a link to the documentation where you have found that this is the correct workaround?
I'm pretty sure the `tcpip` functions are deprecated since quite some time. Also, the `enable_napt` thing was working up until 4.4.8 or 5.0...

indexds:
> Can you paste a link to the documentation where you have found that this is the correct workaround? I'm pretty sure the `tcpip` functions are deprecated since quite some time. Also, the `enable_napt` thing was working up until 4.4.8 or 5.0...

I can't find anything on the doc except this, but it doesn't speak about deprecation for this function at all. I could replicate the crash if you want the logs, it fails an assert with "function must be executed inside a tcpip context".

https://docs.espressif.com/projects/esp-idf/en/stable/esp32c5/api-reference/network/esp_netif_programming.html#_CPPv414esp_netif_nextP11esp_netif_t

ivmarkov:
... and you pasted a link to another function?

indexds:
I know, it's the documentation of that function i'm reading, 

>This API doesn't lock the list, nor the TCPIP context, as this it's usually required to get atomic access between iteration steps rather that within a single iteration. Therefore it is recommended to iterate over the interfaces inside [esp_netif_tcpip_exec()](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c5/api-reference/network/esp_netif_programming.html#group__ESP__NETIF__TCPIP__EXEC_1gade1220a20ab22fa894faa0bf49cfa98e)

>This API is deprecated. Please use [esp_netif_next_unsafe()](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c5/api-reference/network/esp_netif_programming.html#group__ESP__NETIF__LIST_1ga8ec709887656ebf0fbce1cb774d36ee4) directly if all the system interfaces are under your control and you can safely iterate over them. Otherwise, iterate over interfaces using [esp_netif_tcpip_exec()](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c5/api-reference/network/esp_netif_programming.html#group__ESP__NETIF__TCPIP__EXEC_1gade1220a20ab22fa894faa0bf49cfa98e), or use [esp_netif_find_if()](https://docs.espressif.com/projects/esp-idf/en/stable/esp32c5/api-reference/network/esp_netif_programming.html#group__ESP__NETIF__LIST_1ga4b0a98efc632c09c221a639f2bf0ba55) to search in the list of netifs with defined predicate.

It specifically says to use the tcpip exec function, so I don't think that function is deprecated, but that's the only piece of documentation I could find on that function, the tcpip_exec function itself is not marked as deprecated.

indexds:
The crash I had was something like this:

```
assert failed: tcp_alloc /IDF/components/lwip/dontrememberwhere (Required to lock TCPIP core functionality!)
```

PR #559, opened 2025-02-02
Add example of OTA update using http client

EmrysMyrddin:

### Submission Checklist 📝

#### Description

The documentation around OTA is very limited, and you have to make a lot of research on your own.

This PR introduces the example I would have been happy to find when implementing OTA updates on my project :-)

#### Testing
I've copy pasted the code from my working project. Not sure if there are unit tests for examples in this repo ?

#### Comments
I haven't added a changelog entry, not sure the addition of an example is actually a change :-P

### DIFF
--- a/examples/ota_http_client.rs
+++ b/examples/ota_http_client.rs
@@ -30,9 +30,11 @@ use esp_idf_svc::{
 
 use log::{error, info};
 
+const VERSION: &str = "1.0.0"; // You can pull this from an environment variable at build time using env! macro.
+const OTA_FIRMWARE_URI: &str = "http://your.domain/path/to/firmware";
+
 const SSID: &str = env!("WIFI_SSID");
 const PASSWORD: &str = env!("WIFI_PASS");
-const OTA_FIRMWARE_URI: &str = env!("OTA_FIRMWARE_URI");
 
 mod http_status {
     pub const OK: u16 = 200;
@@ -89,7 +91,10 @@ fn check_for_updates(client: &mut HttpClient<EspHttpConnection>) -> anyhow::Resu
         .request(
             Method::Get,
             OTA_FIRMWARE_URI,
-            &[("Accept", "application/octet-stream")],
+            &[
+                ("Accept", "application/octet-stream"),
+                ("X-Esp32-Version", VERSION),
+            ],
         )
         .context("failed to create update request")?;
     let response = request.submit().context("failed to send update request")?;
--- a/examples/ota_http_client.rs
+++ b/examples/ota_http_client.rs
@@ -15,12 +15,14 @@
 //! After a successful OTA update, you will need to reset the `otadata` partition. Otherwise, the ESP
 //! will continue to boot on the second partition, while `cargo run` is flashing the first one by default.
 //! To reset the `otadata` partition, add `--erase-parts otadata` to the runner command in `.cargo/config.toml`.
+#![allow(unexpected_cfgs)]
 
-use anyhow::Context;
+use anyhow::{anyhow, Context};
 use embedded_svc::http::client::Client as HttpClient;
-use esp_idf_svc::http::client::{EspHttpConnection, Method};
+use esp_idf_svc::http::client::{EspHttpConnection, Method, Response};
 use esp_idf_svc::io;
 use esp_idf_svc::log::EspLogger;
+use esp_idf_svc::ota::{EspFirmwareInfoLoad, EspOtaUpdate, FirmwareInfo};
 use esp_idf_svc::wifi::{BlockingWifi, ClientConfiguration, Configuration, EspWifi};
 use esp_idf_svc::{eventloop::EspSystemEventLoop, nvs::EspDefaultNvsPartition};
 use esp_idf_svc::{
@@ -37,7 +39,7 @@ const OTA_FIRMWARE_URI: &str = "http://your.domain/path/to/firmware";
 const SSID: &str = env!("WIFI_SSID");
 const PASSWORD: &str = env!("WIFI_PASS");
 
-// Add package metadata from Cargo.toml. This will be use to determine running firmware version.
+// Add package metadata from Cargo.toml. that will be used to create the App Image and used for OTA.
 esp_app_desc!();
 
 mod http_status {
@@ -98,12 +100,17 @@ fn check_firmware_is_valid() -> anyhow::Result<()> {
     Ok(())
 }
 
-fn check_for_updates(client: &mut HttpClient<EspHttpConnection>) -> anyhow::Result<()> {
-    let current_version = std::str::from_utf8(&esp_app_desc.version)
-        .context("failed to parse version from package metadata")?;
+pub fn check_for_updates(client: &mut HttpClient<EspHttpConnection>) -> anyhow::Result<()> {
+    let mut ota = EspOta::new().context("failed to obtain OTA instance")?;
+
+    let current_version = get_running_version(&ota)?;
+    info!("Current version: {current_version}");
+
+    info!("Checking for updates...");
+
     let headers = [
         ("Accept", "application/octet-stream"),
-        ("X-Esp32-Version", current_version),
+        ("X-Esp32-Version", &current_version),
     ];
     let request = client
         .request(Method::Get, OTA_FIRMWARE_URI, &headers)
@@ -114,13 +121,9 @@ fn check_for_updates(client: &mut HttpClient<EspHttpConnection>) -> anyhow::Resu
         info!("Already up to date");
     } else if response.status() == http_status::OK {
         info!("An update is available, updating...");
-        let mut ota = EspOta::new().context("failed to obtain OTA instance")?;
-
         let mut update = ota.initiate_update().context("failed to initiate update")?;
 
-        match io::utils::copy(response, &mut update, &mut [0; 1024])
-            .context("failed to download update")
-        {
+        match download_update(response, &mut update).context("failed to download update") {
             Ok(_) => {
                 info!("Update done. Restarting...");
                 update.complete().context("failed to complete update")?;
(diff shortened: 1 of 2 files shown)


examples/ota_http_client.rs
```diff
@@ -0,0 +1,147 @@
...
+use esp_idf_svc::wifi::{BlockingWifi, ClientConfiguration, Configuration, EspWifi};
+use esp_idf_svc::{eventloop::EspSystemEventLoop, nvs::EspDefaultNvsPartition};
+use esp_idf_svc::{
+    hal::peripherals::Peripherals,
+    ota::{EspOta, SlotState},
+};
+
+use log::{error, info};
+
+const VERSION: &str = "1.0.0"; // You can pull this from an environment variable at build time using env! macro.
```
Vollbrecht, line 33: We already provide a helper macro that embeds cargo version and build information into a read only part of your application via the esp_app_desc macro defined in esp-idf-sys [here](https://github.com/esp-rs/esp-idf-sys/blob/master/src/app_desc.rs#L2). That way the version you defined in your Cargo.toml can be used and is integrated automatically

Vollbrecht:
Thanks for the effort. A little more documentation can always help!

examples/ota_http_client.rs
```diff
@@ -0,0 +1,147 @@
...
+    let mut client = HttpClient::wrap(EspHttpConnection::new(&Default::default())?);
+
+    check_for_updates(&mut client)?;
+
+    check_valid_state();
+
+    Ok(())
+}
+
+fn check_valid_state() -> anyhow::Result<()> {
```
ivmarkov, line 70: Perhaps document that this functionality needs `CONFIG_BOOTLOADER_APP_ROLLBACK_ENABLE=y` which is not the case by default?

examples/ota_http_client.rs
```diff
@@ -0,0 +1,147 @@
...
+        sys_loop,
+    )?;
+
+    connect_wifi(&mut wifi)?;
+
+    let mut client = HttpClient::wrap(EspHttpConnection::new(&Default::default())?);
+
+    check_for_updates(&mut client)?;
+
+    check_valid_state();
```
ivmarkov, line 65: This has to be done post app-reboot, once the newly-updated partition was booted. Perhaps document this?

examples/ota_http_client.rs
```diff
@@ -0,0 +1,147 @@
...
+
+    if response.status() == http_status::NOT_MODIFIED {
+        info!("Already up to date");
+    } else if response.status() == http_status::OK {
+        info!("An update is available, updating...");
+        let mut ota = EspOta::new().context("failed to obtain OTA instance")?;
+
+        let mut update = ota.initiate_update().context("failed to initiate update")?;
+
+        match io::utils::copy(response, &mut update, &mut [0; 1024])
```
ivmarkov, line 110: Would be good to demo the usage of `EspFirmwareInfoLoad` here.

ivmarkov (changes requested):
Thanks, the example looks quite good already! I have two suggestions - one to more heavily document the bootloader-based rollback mechanism and the other - to demo `EspFirmweareInfoLoad` whereas the second thing I believe is important.

examples/ota_http_client.rs
EmrysMyrddin: Ooooh very cool ! I'm fixing this :-)

examples/ota_http_client.rs
EmrysMyrddin: In the body of the function, the check happens only if the currently running slot is not yet marked as valid.

I found it easier like this, to avoid having complex OTA checks in the middle of the code ^^' In my project for example, I call this function right after I've successfully opened my websocket to the server.

examples/ota_http_client.rs
EmrysMyrddin: Not sure what you mean here ? You want to demo the usage to do what ?

examples/ota_http_client.rs
EmrysMyrddin: How this have to be set ? Is there a file somewhere for the sep-idf related configuration?

EmrysMyrddin:
By browsing the ESP-IDF documentation, it appears the even have a native function that is handling the update process over HTTP automatically. You just provide an URI and it does all the process for you.

It has been not been exposed on purpose ? If not, I would be happy to make a PR for it :-)

examples/ota_http_client.rs
ivmarkov: `sdkconfig.defaults` in the root of your binary crate. You should absolutely have this file there anyway.
Also, since this is a bootloader setting, you _must_ flash the custom bootloader that is anyway built as part of building your app crate, or else the check_valid_state fn would do essentially nothing, as the standard bootloader flashed by `espflash` is unlikely to have this setting enabled (it is not enabled by default).

examples/ota_http_client.rs
ivmarkov: I think you are missing the point as to what the check is supposed to do.
Read this: https://docs.espressif.com/projects/esp-idf/en/v4.2/esp32/api-reference/system/ota.html#app-rollback

My point is, **unless you reboot** you won't be running the _new_ firmware. And then, unless you are running the _new_ firmware, what exactly are you checking? `check_valid_state` is supposed to be invoked once the _new_ firmware is running for - say - a couple of minutes and did not crash in the meantime... It is completely pointless to tell the bootloader "hey, the new firmware is fine!" given that you are actually not yet running the new firmware in the first place.

examples/ota_http_client.rs
ivmarkov: `EspFirmwareInfoLoad` allows you - while loading the new firmware - to extract from it on-the-fly exactly this `esp_app_desc!` information @Vollbrecht told you about. Further on, once you extract this information (while in the midst of flashing the new firmware) you are supposed to check that e.g. (a) the image you are flashing is a newer version than what you are having (b) it is in general correct (i.e. has this information in the first place).

This is of course optional, and you can completely skip it if you trust your HTTP server that it would always serve you a correct image and also if you do this do this with a sort of an "out of band" delivery of the firmware version using an HTTP header as you did.

But it probably still makes sense to at least hint the users that they can check the image itself using `EspFirmwareInfoLoad`.

ivmarkov:
> By browsing the ESP-IDF documentation, it appears the even have a native function that is handling the update process over HTTP automatically. You just provide an URI and it does all the process for you.
> 
> It has been not been exposed on purpose ? If not, I would be happy to make a PR for it :-)

Sure, but this function is binding the `EspOta` api to the ESP IDF HTTP client. Not all folks might be using the ESP IDF HTTP client in the first place (I'm myself using `edge-http` instead). Also, not all users might choose HTTP as the channel for delivering the OTA image.

In other words, the API you want to expose is too opinionated. I don't mind you working on exposing it, but I don't think the existing `EspOta` API can be superseded in any way by it, as the exposed API is generic and agnostic as to from where and how you pull the new OTA image (HTTP or something else), while the function you have in mind is not.

examples/ota_http_client.rs
ivmarkov: ... in a comment or so.

examples/ota_http_client.rs
EmrysMyrddin: The `check_for_updates` function is rebooting the board once the update is done :-) 

So if the program is reaching `check_valid_state()` call, it means we are already up to date. So we should verify if the running slot is valid or in pending verification state, which is what `check_valid_state` is doing.

I agree in a normal project, you would probably mark the check later, once you are sure you are able to run normally.

I've added some comments to try to make it more clear.

EmrysMyrddin:
>In other words, the API you want to expose is too opinionated. I don't mind you working on exposing it, but I don't think the existing EspOta API can be superseded in any way by it, as the exposed API is generic and agnostic as to from where and how you pull the new OTA image (HTTP or something else), while the function you have in mind is not.

Yes I totally agree ! It's more a util function that is ready to use for those how don't want manually implement it :-)
I will see if it is easy to expose, otherwise the code for making it manually is not that complex anyway.

examples/ota_http_client.rs
EmrysMyrddin: From what I've understood of this section: https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/ota.html#app-rollback
the default behavior of the bootloader is to select a not invalid slot. So it can be either Valid or Pending Validation. Which means that if the slot is explicitly marked as invalid, it will not boot.

I will test this to be sure about it.

examples/ota_http_client.rs
EmrysMyrddin: Oh ok, I will try to see how it works in my project to add it to the example.
Even if the server does check the version, double checking is never a bad idea :-P 
And more importantly, I should definitly at least verify some checksum to validate we have corretly downloaded the file. My server sends a header with the md5 of the file.

I'm not sure to understand how to extract this `EspFirmwareInfoLoad` from my update partition. Or you mean I should not use `io::utils::copy` but do the reading manually, at least until `EspFirmwareInfoLoad::fetch` returns true ?

EmrysMyrddin:
I need some help for using `esp_app_desc.version` :-) From what I understood, it a `CStr`, but I'm a bit lost about how to obtain a `&str` from it. Is there a way to do it without relying on `unsafe` keyword ?

examples/ota_http_client.rs
ivmarkov: > I'm not sure to understand how to extract this `EspFirmwareInfoLoad` from my update partition. Or you mean I should not use `io::utils::copy` but do the reading manually, at least until `EspFirmwareInfoLoad::fetch` returns true ?

Yes. That one. You should read in chunks and "feed" `EspFirmwareInfoLoad` with the data read so far.

Since the `esp_app_desc_t` struct is at the beginning of the firmware image (unless you are using the c6, where there is currently a bug with that), you need a buffer of a max 1K to accumulate the initial image bytes. Actually less than 1K, but reading from HTTP with a smaller buffer is not very efficient.

ivmarkov:
> I need some help for using `esp_app_desc.version` :-) From what I understood, it a `CStr`, but I'm a bit lost about how to obtain a `&str` from it. Is there a way to do it without relying on `unsafe` keyword ?

What are you trying to do? The `esp_app_desc!` macro does not take any parameters and sets your version to the one you have for your package in `Cargo.toml`. Furthermore, the `EspFirmwareInfoLoad::fetch` method populates directly a `FirmwareInfo` struct where everything is with `&str` or` heapless::String` already. With that said, just browse the `CStr` API.

EmrysMyrddin:
I must misunderstand what `esp_app_desc!` is doing ^^'
It's not supposed to allow me to get the firmware version ? as pointed out by @Vollbrecht ?

ivmarkov:
> I must misunderstand what `esp_app_desc!` is doing ^^' It's not supposed to allow me to get the firmware version ? as pointed out by @Vollbrecht ?

It is _defining_ the firmware version.
It is just putting a pre-populated `esp_app_desc_t` structure in a well-defined place into your firmware app image, which you can then read using `EspFirmwareInfoLoad`; and which is also read by the `EspOta` methods, which are returning it to you in the `Slot` structure. Read this https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/app_image_format.html#application-description and overall the whole page.

EmrysMyrddin:
Hooo that's makes sense ! That's why the version was not correct in the running slot when I checked ! Thank you for you help on all of this ^^'

EmrysMyrddin:
BTW, when  i add `esp_app_desc!`, it generate tons of warning about `unexpected `cfg` condition name`. Is it expected ? There isn't much documentation around this macro, so I'm not sure i miss some setup.

examples/ota_http_client.rs
EmrysMyrddin: Ok, I get it. Sorry to bother you again, but how do you create a reference for a `FirmwareInfo` ? I have to manually create the struct ? I don't see any function that creates one.

ivmarkov: Yes, manually with some default values. For now.

examples/ota_http_client.rs
EmrysMyrddin: Let's add information about that too :-)
I didn't noticed, since my app is already too large to fit 1 FR + 2 OTA partitions in my 4MB flash ^^'

EmrysMyrddin:
Ok, I think I've resolved all your comments ?

BTW, I'm not sure why the `--bootloader` option is not set by default in the `.cargo/config.yml` by default in the template ? I struggled to find out why my bootloader was not updated while changing settings in `sdkconfig.defaults` ^^'

ivmarkov:
Sorry for the delay, could not get to it today. Will take a look tmr.

EmrysMyrddin:
No problem, there is no rush :-) It's just documentation !

ivmarkov:
Thanks again!

PR #560, opened 2025-02-14
Implement MQTT outbox limits and get_outbox_size()

mike1703:

### Submission Checklist 📝
- [X] I have used `cargo fmt` command to ensure that all changed code is formatted correctly.
- [X] I have used `cargo clippy` command to ensure that all changed code passes latest Clippy nightly lints.
- [X] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-idf-svc/blob/main/esp-idf-svc/CHANGELOG.md) in the **_proper_** section.

#### Description
This change implements get_outbox_size() on the EspMqttClient to retrieve the amount of bytes currently stored in the MQTT outbox.
Additionally implement a configurable option in MqttClientConfiguration to limit the outbox size to the given amount of bytes so there is no ever growing outbox size that might fill up the entire RAM.

#### Testing
I tested the limit with this configuration and sent a a lot of bytes
```
        &MqttClientConfiguration {
            client_id: Some(client_id),
            username,
            password,
            outbox_limit: Some(5000),
            ..Default::default()
        },
```
and tested it with
```
    if client.get_outbox_size() > 5100 {
        error!("too much in the outbox, ignore this message");
    } 
```
The message is shown if no limit is configured and does not show up once a limit of 5000 is set.

```
            match client.enqueue(
                &topic,
                esp_idf_svc::mqtt::client::QoS::AtLeastOnce,
                false,
                value.as_bytes(),
            ) {
                Ok(message_id) => debug!("enqueued {topic:?} = {value} ({message_id})"),
                Err(error) => error!("cannot enqueue mqtt message: {error}"),
            }
```

This returns `cannot enqueue mqtt message: ERROR` if the outbox is full.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - OTA: New method - `EspFirmwareInfoLoad::fetch_native` - returning the full native ESP-IDF image descriptor structures
 - Added `use_serde` feature, which enables the `use_serde` feature of `embedded-svc` crate, allowing to deserialize configuration structs.
+- Implement MQTT outbox limit and get_outbox_size()
 
 ## [0.51.0] - 2025-01-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Breaking
+- Implement MQTT outbox limit and get_outbox_size()
+
 ### Fixed
 - Fix wrong BT configuration version on the c6 (issue #556)
 
 ### Added
 - OTA: New method - `EspFirmwareInfoLoad::fetch_native` - returning the full native ESP-IDF image descriptor structures
 - Added `use_serde` feature, which enables the `use_serde` feature of `embedded-svc` crate, allowing to deserialize configuration structs.
-- Implement MQTT outbox limit and get_outbox_size()
 
 ## [0.51.0] - 2025-01-15
 
--- a/src/mqtt/client.rs
+++ b/src/mqtt/client.rs
@@ -67,7 +67,7 @@ pub struct MqttClientConfiguration<'a> {
     pub task_stack: usize,
     pub buffer_size: usize,
     pub out_buffer_size: usize,
-    pub outbox_limit: Option<u64>,
+    pub outbox_limit: Option<usize>,
 
     pub username: Option<&'a str>,
     pub password: Option<&'a str>,
@@ -319,7 +319,7 @@ impl<'a> TryFrom<&'a MqttClientConfiguration<'a>>
         }
 
         if let Some(outbox_limit) = conf.outbox_limit {
-            c_conf.outbox.limit = outbox_limit;
+            c_conf.outbox.limit = outbox_limit as _;
         }
 
         #[cfg(all(esp_idf_esp_tls_psk_verification, feature = "alloc"))]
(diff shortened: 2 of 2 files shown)


Vollbrecht (commented):
Looks reasonable.The only question is if we should model the no_limit as 0 and not an Option as most other fields here, but i guess that's a nit. ( Assuming no limit is actually 0 and not something like u64::max :D )

Technically its a breaking change since we add a pub field to the struct but well its so minimal its fine from my site.

ivmarkov:
> Looks reasonable.The only question is if we should model the no_limit as 0 and not an Option as most other fields here, but i guess that's a nit. ( Assuming no limit is actually 0 and not something like u64::max :D )
> 
> Technically its a breaking change since we add a pub field to the struct but well its so minimal its fine from my site.

I actually prefer `Option`.

Also, I might very soon introduce 2 breaking changes, so breaking the API in a minor way probably not a big deal.

mike1703:
Regarding Option vs 0: I thought about going this route but then tried to mimic the style of e.g. buf.out_size.

I added a Default::default for the outbox_config_t type as I was missing this.

Unfortunately I did not find any public documentation on the behavior of the default 0 but looked at sources:
https://github.com/espressif/esp-mqtt/commit/372ab7b37451e7f13bf8433cb92ebd81b3e3d45b introduces this feature (and is still there https://github.com/espressif/esp-mqtt/blob/cea7e959eee68fc974fcae92572a1c37eca72d06/mqtt_client.c#L2119)

ivmarkov:
OK - thanks a lot! 
I'll review in details later today/tmr.

In general I need this too as well - in fact - ASAP - so this is arriving just in time. :)

mike1703:
btw, if you want me to move the change from the added to the breaking section, I can totally do this. I was just not if I should based on your comments

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - OTA: New method - `EspFirmwareInfoLoad::fetch_native` - returning the full native ESP-IDF image descriptor structures
 - Added `use_serde` feature, which enables the `use_serde` feature of `embedded-svc` crate, allowing to deserialize configuration structs.
+- Implement MQTT outbox limit and get_outbox_size()
```
ivmarkov, line 16: This should be moved to the "Breaking" section, because our MQTT struct is _not_ (yet) marked with `#[non_exhaustive]` and thus adding a new member to it is a breaking change.

src/mqtt/client.rs
```diff
@@ -67,6 +67,7 @@ pub struct MqttClientConfiguration<'a> {
     pub task_stack: usize,
     pub buffer_size: usize,
     pub out_buffer_size: usize,
+    pub outbox_limit: Option<u64>,
```
ivmarkov, line 70: While the `size` member of the MQTT C `outbox_` struct is indeed `uint64_t` (for whatever reason? it is also marked as atomic btw) I don't think we need to surface this oddity in our API. `outbox_limit` should be `Option<usize>` instead.

src/mqtt/client.rs
```diff
@@ -625,6 +631,10 @@ impl<'a> EspMqttClient<'a> {
         Self::check(unsafe { esp_mqtt_client_set_uri(self.raw_client, uri.as_ptr()) })
     }
 
+    pub fn get_outbox_size(&self) -> i32 {
```
ivmarkov, line 634: `i32` here is a weird choice. ESP-IDF seems to cast the internal `uint64_t` to `i32` (weird?? why?). Regardless, we should instead return a `usize`.

ivmarkov (changes requested):

src/mqtt/client.rs
mike1703: Yes, I'm not sure if this was done by accident or on purpose but internally it is by default 0 and if available the uint64_t.
I gave it another try and capped the value at 0 and return it as usize. wdyt?

CHANGELOG.md
mike1703: done

src/mqtt/client.rs
mike1703: agree

PR #563, opened 2025-03-06
Compile MountedLittlefs only if experimental is enabled

wzhd:
### Submission Checklist 📝
This a minor patch to fix broken compilation under some conditions.

#### Description
`crate::fs` is only compiled when the `experimental` feature is enabled, but the user may have started using the `joltwallet/littlefs` component through its raw bindings, without relying on `experimental`.
But `esp_idf_comp_joltwallet__littlefs_enabled` would be true, causing `MountedLittlefs` to be compiled even when `crate::fs` isn't. `crate::fs::littlefs::PartitionRawData` can't be unresolved and breaks the build.

#### Testing
I'm using a local repo of esp-idf-svc as a path dependency.

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -155,13 +155,21 @@ pub mod vfs {
     }
 
     /// Represents a mounted Littlefs filesystem.
-    #[cfg(all(feature = "alloc", esp_idf_comp_joltwallet__littlefs_enabled))]
+    #[cfg(all(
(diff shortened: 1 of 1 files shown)


ivmarkov:
Thanks and sorry for getting to it with a delay! I think we do not need to mark this in the Changelog - merging.

PR #567, opened 2025-03-13
Fixed inconsistent mutability in NVS

nandee95:
### Submission Checklist 📝
- [-] I have updated existing examples or added new ones (if applicable). 
  - No examples affected

### Pull Request Details 📖
Some functions in the NVS interface were left mutable which forced the use of a second mutex from rust. To avoid the need for double locking i removed mutability from these functions. This solution is viable because the C interface is already protected by mutexes and locking from rust should not be necessary. (The majority of the setters are non-mutable currently)
The C lock is visible here: https://github.com/espressif/esp-idf/blob/master/components/nvs_flash/src/nvs_api.cpp#L421

#### Testing
- I double checked that all the affected functions are indeed protected on the C side.
- The functions are working without a mutex from rust.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Fix wrong BT configuration version on the c6 (issue #556)
+- Fix inconsistent mutability in NVS (#567)
 
 ### Added
 - OTA: New method - `EspFirmwareInfoLoad::fetch_native` - returning the full native ESP-IDF image descriptor structures
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -330,7 +330,7 @@ impl<T: NvsPartitionId> EspNvs<T> {
         self.len(name).map(|v| v.is_some())
     }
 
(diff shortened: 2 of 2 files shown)


ivmarkov:
I'm fine with this.

In general, `esp-idf-hal` and `esp-idf-svc` suffer from the problem of being "too pessimistic" and putting `&mut` instead of `&self` at a bit too many places where the actual driver/service in ESP-IDF is actually thread-safe. As it is sometimes difficult to "figure it out" and ESP-IDF docu is not always very explicit about the thread-safety of a particular component.

So I agree, we have to fix it everywhere in the log run so why not start with the NVS functionality?

If you update the Changelog (not sure, does this qualify as a breaking change?), I would be happy to merge.

nandee95:
Switching from `&mut self` to just `&self` is more permissive hence its not a braking change. All the previous code should work and compile as It did before.

Updated the changelog.

ivmarkov:
> Switching from `&mut self` to just `&self` is more permissive hence its not a braking change. All the previous code should work and compile as It did before.

Sounds reasonable, yet I had my fair share of backwards-incompatible gotchas so I lean on the paranoid side.

> 
> Updated the changelog.

Thanks, merged!

PR #574, opened 2025-04-06
refactor: add must_use to event loop handles

domodwyer:
This helps catch accidental misuse of the event loop subscription API, wherein a subscription is created and then immediately removed due to the returned handle not being retained.

It causes a compiler warning to be emitted that looks like this:

```
warning: unused `EspSubscription` that must be used
  --> src/wifi.rs:56:9
   |
56 | /         event_loop
57 | |             .subscribe::<WifiEvent, _>({
...  |
90 | |             })
91 | |             .expect("failed to subscribe to event bus");
   | |_______________________________________________________^
   |
   = note: Event subscription is unregistered when handle is dropped
   = note: `#[warn(unused_must_use)]` on by default
help: use `let _ = ...` to ignore the resulting value
   |
56 |         let _ = event_loop
   |         +++++++
```

### Submission Checklist 📝
- [N/A] I have updated existing examples or added new ones (if applicable).

### DIFF
--- a/src/eventloop.rs
+++ b/src/eventloop.rs
@@ -331,6 +331,7 @@ where
     }
 }
 
+#[must_use = "Event subscription is unregistered when handle is dropped"]
 pub struct EspSubscription<'a, T>
 where
     T: EspEventLoopType,
@@ -486,6 +487,7 @@ where
     }
 }
 
+#[must_use = "Event loop is deleted when handle is dropped"]
 #[derive(Debug)]
 struct EventLoopHandle<T>(T)
 where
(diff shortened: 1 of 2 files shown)


src/eventloop.rs
```diff
@@ -486,6 +487,7 @@ where
     }
 }
 
+#[must_use = "Event loop is deleted when handle is dropped"]
```
ivmarkov, line 490: I wonder why you need that one given that it is an internal type? Just for the esp-idf-svc code itself, or am I missing something?

domodwyer: I was just about to open the PR with the `EspSubscription` annoation and I thought "I'll quickly check if there's any other Drop impls" and so stuck one on this - didn't even notice it wasn't pub.

I've pushed 717c0fb to remove this one :+1:

PR #576, opened 2025-04-09
feat(bt): Add embassy trouble / bt_hci support

Vollbrecht:
esp-idf-svc now provides an implementation of the bt_hci::Transport trait, enabling integration with the embassy-trouble Bluetooth stack.

This allows users to construct a ExternalController using the BtDriver and have it work seamlessly with embassy-trouble.

Three basic examples where added. For more checkout the examples in embassy-trouble/host

### Submission Checklist 📝

### DIFF
--- a/examples/bt_trouble_gatt_central.rs
+++ b/examples/bt_trouble_gatt_central.rs
@@ -32,7 +32,10 @@ fn main() -> anyhow::Result<()> {
     #[cfg(esp32s2)]
     panic!("ESP32-S2 does not have a BLE radio");
 
-    #[cfg(any(not(feature = "critical-section"), not(feature = "trouble")))]
+    #[cfg(all(
+        not(esp32s2),
+        any(not(feature = "critical-section"), not(feature = "trouble"))
+    ))]
     panic!("Use `--features trouble,critical-section` when building this example");
 }
 
--- a/examples/bt_trouble_gatt_peripheral.rs
+++ b/examples/bt_trouble_gatt_peripheral.rs
@@ -31,7 +31,10 @@ fn main() -> anyhow::Result<()> {
     #[cfg(esp32s2)]
     panic!("ESP32-S2 does not have a BLE radio");
 
-    #[cfg(any(not(feature = "critical-section"), not(feature = "trouble")))]
+    #[cfg(all(
+        not(esp32s2),
+        any(not(feature = "critical-section"), not(feature = "trouble"))
+    ))]
     panic!("Use `--features trouble,critical-section` when building this example");
 }
 
--- a/examples/bt_trouble_scan.rs
+++ b/examples/bt_trouble_scan.rs
@@ -31,7 +31,10 @@ fn main() -> anyhow::Result<()> {
     #[cfg(esp32s2)]
     panic!("ESP32-S2 does not have a BLE radio");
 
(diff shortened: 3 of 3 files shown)


Vollbrecht:
Have to investigate why linking against the c6 failes. The vhci api in its form i am using is available in that form through all esp-idf version and targets that have ble radio. A bit busy today, will look deeper at the weekend.

Vollbrecht:
Hmm, to my understanding in the esp32c6/h2 the underlying ble api is quite different. Even if they officially [expose](https://github.com/espressif/esp-idf/blob/master/components/bt/include/esp32c6/include/esp_bt.h#L401) the bt hci api, it seams on older ESP_IDF versions not possible to:
- run in CONTROLER_ONLY mode and using the api on c6/h2, e.g i can link fine when enabling bluedroid alongside, though that defeats the purpose a bit and introduces a lof of bloat. 
- the api itself is in kind of a wired as the same functions are now a "synchronize" api according to this [issue](https://github.com/espressif/esp-idf/issues/14655) .
- There are still recent changes to h2/c6 hci api in esp-idf that may even only allow to work with them correctly now on newer esp-idf versions looking [here](https://github.com/espressif/esp-idf/commit/ac2b67d6c5011494031ab7c9e3100366db050f6b)  and [here](https://github.com/espressif/esp-idf/commit/e154fa0af2d9cee814062d804b2d72f783300899).


That all makes it currently much harder to support c6/h2. So either this PR stays blocked till i find how to fix the above or i simply exlude h2/c6 for now...

ivmarkov:
> Hmm, to my understanding in the esp32c6/h2 the underlying ble api is quite different. Even if they officially [expose](https://github.com/espressif/esp-idf/blob/master/components/bt/include/esp32c6/include/esp_bt.h#L401) the bt hci api, it seams on older ESP_IDF versions not possible to:
> 
> * run in CONTROLER_ONLY mode and using the api on c6/h2, e.g i can link fine when enabling bluedroid alongside, though that defeats the purpose a bit and introduces a lof of bloat.
> * the api itself is in kind of a wired as the same functions are now a "synchronize" api according to this [issue](https://github.com/espressif/esp-idf/issues/14655) .
> * There are still recent changes to h2/c6 hci api in esp-idf that may even only allow to work with them correctly now on newer esp-idf versions looking [here](https://github.com/espressif/esp-idf/commit/ac2b67d6c5011494031ab7c9e3100366db050f6b)  and [here](https://github.com/espressif/esp-idf/commit/e154fa0af2d9cee814062d804b2d72f783300899).
> 
> That all makes it currently much harder to support c6/h2. So either this PR stays blocked till i find how to fix the above or i simply exlude h2/c6 for now...

I wonder how `esp-hal` is capable of providing a bt-hci controller impl on every espressif chip using the espressif bt blobs, and somehow this is not possible when using these same blobs with esp idf?

Vollbrecht:
They are not using this api, that i am using, but using a special low level api for the c6/h2. If you have a look [here](https://github.com/esp-rs/esp-hal/tree/main/esp-wifi/src/ble) the btdm.rs is basically what we are also using here as an api and that is directly supported by esp-idf. For the h2/c6 they than use the npl.rs api, that i don't have any infos about.

ivmarkov:
This is now unmergable. I think if somebody wants to address #676, this, and also the newer #677 can be used for inspiration, but it should be a fresh start.

PR #579, opened 2025-04-23
Add keep alive enable configuration to the EspHttpConnection.

gustavowd:

### Submission Checklist 📝
- [ x ] I have used `cargo fmt` command to ensure that all changed code is formatted correctly.
- [ x ] I have used `cargo clippy` command to ensure that all changed code passes latest Clippy nightly lints.
- [ x ] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-idf-svc/blob/main/esp-idf-svc/CHANGELOG.md) in the **_proper_** section.

### Pull Request Details 📖
Simple add an option the enable the keep alive configuration of the http client.

#### Description
Due to not enable the keep alive, sometimes the connection timed out, bringing several issues to the upper layer functions.

#### Testing
I tested the new client with REST API requests.

### DIFF
--- a/src/http/client.rs
+++ b/src/http/client.rs
@@ -86,6 +86,7 @@ pub struct Configuration {
     pub use_global_ca_store: bool,
     pub crt_bundle_attach: Option<unsafe extern "C" fn(conf: *mut core::ffi::c_void) -> esp_err_t>,
     pub raw_request_body: bool,
+    pub keep_alive_enable: Option<bool>
 }
 
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
@@ -136,6 +137,10 @@ impl EspHttpConnection {
 
         if let Some(timeout) = configuration.timeout {
             native_config.timeout_ms = timeout.as_millis() as _;
+        }
+
+        if let Some(keep_alive_enable) = configuration.keep_alive_enable {
+            native_config.keep_alive_enable = keep_alive_enable as _;
         }
 
         if let (Some(cert), Some(private_key)) =
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - OTA: Allow specifying image size to speed up erase
 - Bluetooth: New methods `EspBleGap::start_scanning` and `EspBleGap::stop_scanning`
 - New example, `bt_ble_gap_scanner` to demonstrate usage of added ble scanning methods
+- Added keep_alive_enable in the Configuration structure that is used in the EspHttpConnection function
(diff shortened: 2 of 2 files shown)


src/http/client.rs
```diff
@@ -86,6 +86,7 @@ pub struct Configuration {
     pub use_global_ca_store: bool,
     pub crt_bundle_attach: Option<unsafe extern "C" fn(conf: *mut core::ffi::c_void) -> esp_err_t>,
     pub raw_request_body: bool,
+    pub keep_alive_enable: Option<bool>,
```
ivmarkov, line 89: What does `None` mean for `keep_alive_enabled`? If it means same as `Some(false)`, then why are we hiding this flag behind an `Option`?

ivmarkov (changes requested):

src/http/client.rs
gustavowd: > What does `None` mean for `keep_alive_enabled`? If it means same as `Some(false)`, then why are we hiding this flag behind an `Option`?

In fact, `None` would cause the http client library to use the default option for `keep_alive_enabled`, which is currently false. However, if the library changes in the future and starts specifying this option as true by default, `None` would maintain this condition.

In any case, I have no objection to removing the Option and specifying just a boolean.

ivmarkov: I think it is simpler for the users to just use `false`.

Moreover, there is no such thing as a predefined default in esp-idf that might change. The C struct is initially just all zeroes, which for that particular parameter corresponds to `false`.

gustavowd: @ivmarkov 

Actually, there is a default configuration for keep alive if you set in the config keep_alive_enable to true.

```
static bool init_common_tcp_transport(esp_http_client_handle_t client, const esp_http_client_config_t *config, esp_transport_handle_t transport)
{
    if (config->keep_alive_enable == true) {
        client->keep_alive_cfg.keep_alive_enable = true;
        client->keep_alive_cfg.keep_alive_idle = (config->keep_alive_idle == 0) ? DEFAULT_KEEP_ALIVE_IDLE : config->keep_alive_idle;
        client->keep_alive_cfg.keep_alive_interval = (config->keep_alive_interval == 0) ? DEFAULT_KEEP_ALIVE_INTERVAL : config->keep_alive_interval;
        client->keep_alive_cfg.keep_alive_count =  (config->keep_alive_count == 0) ? DEFAULT_KEEP_ALIVE_COUNT : config->keep_alive_count;
        esp_transport_tcp_set_keep_alive(transport, &client->keep_alive_cfg);
    }
```

Maybe should be interesting to have the keep_alive_enabled just as a boolean and the other configurations as Option<>.
What do you think about this?

PR #588, opened 2025-07-25
fix(nvs): resolve contains() method returning false for existing string values

### LINKED ISSUE #585: ESP-IDF-SVC NVS contains() Bug Report - String Values Not Detected
# ESP-IDF-SVC NVS contains() Bug Report - String Values Not Detected

## Bug description

The `contains()` method in `EspNvs` returns `false` for string values that actually exist in NVS partitions, even though the same strings can be successfully retrieved using `get_str()`. This creates inconsistent behavior where data exists in flash but appears missing to the application.

**Root Cause**: The `contains()` method internally calls `len()`, which only checks for `u64` and `blob` data types via `nvs_get_u64()` and `nvs_get_blob()`. It never calls `nvs_get_str()` to check for string values stored with `nvs_set_str()`.

**Would you like to work on a fix?** Yes, we have identified the exact issue and can provide a complete solution.

## To Reproduce

1. Create an NVS partition in your ESP32 project (default or custom)
2. Initialize an `EspNvs` or `EspCustomNvs` instance with a namespace
3. Store a string value using `set_str(key, "some_string")`
4. Call `contains(key)` - it returns `false`
5. Call `get_str(key, buffer)` - it successfully retrieves the string
6. Reboot the device and repeat steps 4-5 - same inconsistent behavior

**Minimal reproduction code**:

```rust
use esp_idf_svc::nvs::{EspDefaultNvsPartition, EspNvs};
// or for custom partitions: use esp_idf_svc::nvs::{EspCustomNvsPartition, EspCustomNvs};

fn reproduce_bug() -> Result<(), Box<dyn std::error::Error>> {
    // Using default partition (bug affects all partition types)
    let partition = EspDefaultNvsPartition::take()?;
    let mut nvs = EspNvs::new(partition, "test_ns", true)?;
    
    // Alternative: Custom partition
    // let partition = EspCustomNvsPartition::take("custom_partition")?;
    // let mut nvs = EspCustomNvs::new(partition, "test_ns", true)?;
    
    // Set a string value
    nvs.set_str("test_key", "test_value")?;
    
    // BUG: contains() returns false for string values
    let contains_result = nvs.contains("test_key")?;
    println!("contains(): {}", contains_result); // Prints: false
    
    // But get_str() works correctly
    let mut buffer = [0u8; 64];
    let get_result = nvs.get_str("test_key", &mut buffer)?;
    println!("get_str(): {:?}", get_result); // Prints: Some("test_value")
    
    // Inconsistent behavior: data exists but contains() can't find it
    assert!(contains_result != get_result.is_some()); // This assertion passes, proving the bug
    
    Ok(())
}
```

## Expected behavior

`contains(key)` should return `true` for any existing NVS entry regardless of data type (u8, u16, u32, u64, i8, i16, i32, i64, string, blob). The method should be consistent with other NVS operations like `get_str()`, `get_u32()`, etc.

**Expected behavior**:

- `nvs.set_str("key", "value")` - stores string successfully ✅
- `nvs.contains("key")` - returns `true` ✅ (currently returns `false` ❌)
- `nvs.get_str("key", buffer)` - retrieves string successfully ✅

## Root Cause Analysis

The bug is in `esp-idf-svc/src/nvs.rs`:

```rust
pub fn contains(&self, name: &str) -> Result<bool, EspError> {
    self.len(name).map(|v| v.is_some())  // ← calls len()
}

fn len(&self, name: &str) -> Result<Option<usize>, EspError> {
    let c_key = to_cstring_arg(name)?;
    let mut value: u_int64_t = 0;

    // check for u64 value
    match unsafe { nvs_get_u64(self.1, c_key.as_ptr(), &mut value as *mut _) } {
        ESP_ERR_NVS_NOT_FOUND => {
            // check for blob value, by getting blob length
            let mut len = 0;
            match unsafe {
                nvs_get_blob(self.1, c_key.as_ptr(), ptr::null_mut(), &mut len as *mut _)
            } {
                ESP_ERR_NVS_NOT_FOUND => Ok(None),  // ← STOPS HERE, never checks strings!
                // ... only handles blob case
            }
        }
        // ... only handles u64 case
    }
}
```

**Problem**: The `len()` method only checks `nvs_get_u64()` and `nvs_get_blob()`. It never calls `nvs_get_str()` for string values.

### Option 1: Fix the `len()` method to check string values

```rust
fn len(&self, name: &str) -> Result<Option<usize>, EspError> {
    let c_key = to_cstring_arg(name)?;
    
    // Check for u64 value
    let mut u64_value: u_int64_t = 0;
    match unsafe { nvs_get_u64(self.1, c_key.as_ptr(), &mut u64_value as *mut _) } {
        ESP_ERR_NVS_NOT_FOUND => {
            // Check for blob value
            let mut blob_len = 0;
            match unsafe { nvs_get_blob(self.1, c_key.as_ptr(), ptr::null_mut(), &mut blob_len as *mut _) } {
                ESP_ERR_NVS_NOT_FOUND => {
                    // ← ADD THIS: Check for string value
                    let mut str_len = 0;
                    match unsafe { nvs_get_str(self.1, c_key.as_ptr(), ptr::null_mut(), &mut str_len as *mut _) } {
                        ESP_ERR_NVS_NOT_FOUND => Ok(None),
                        err => {
                            esp!(err)?;
                            Ok(Some(str_len))
                        }
                    }
                }
                err => {
                    esp!(err)?;
                    Ok(Some(blob_len))
                }
            }
        }
        err => {
            esp!(err)?;
            Ok(Some(8)) // u64 is always 8 bytes
        }
    }
}
```

### Option 2: Use `nvs_find_key()` in `contains()` method (Recommended)

```rust
pub fn contains(&self, name: &str) -> Result<bool, EspError> {
    let c_key = to_cstring_arg(name)?;
    let mut entry_type: nvs_type_t = nvs_type_t_NVS_TYPE_ANY;
    
    let result = unsafe { 
        nvs_find_key(self.1, c_key.as_ptr(), &mut entry_type as *mut _) 
    };
    
    match result {
        ESP_OK => Ok(true),
        ESP_ERR_NVS_NOT_FOUND => Ok(false),
        err => {
            esp!(err)?;
            Ok(false) // This line should never be reached due to esp!() macro
        }
    }
}
```

### Option 3: Enhanced `find_key()` method that returns data type (Additional Feature)

The `nvs_find_key` function can also return the data type of the found entry. Consider adding this as a new method:

```rust
pub fn find_key(&self, name: &str) -> Result<Option<nvs_type_t>, EspError> {
    let c_key = to_cstring_arg(name)?;
    let mut entry_type: nvs_type_t = nvs_type_t_NVS_TYPE_ANY;
    
    let result = unsafe { 
        nvs_find_key(self.1, c_key.as_ptr(), &mut entry_type as *mut _) 
    };
    
    match result {
        ESP_OK => Ok(Some(entry_type)),
        ESP_ERR_NVS_NOT_FOUND => Ok(None),
        err => {
            esp!(err)?;
            Ok(None) // This line should never be reached due to esp!() macro
        }
    }
}
```

This would allow applications to:

```rust
// Check existence and get type in one call
if let Some(data_type) = nvs.find_key("my_key")? {
    match data_type {
        nvs_type_t_NVS_TYPE_STR => { /* handle string */ },
        nvs_type_t_NVS_TYPE_U32 => { /* handle u32 */ },
        nvs_type_t_NVS_TYPE_BLOB => { /* handle blob */ },
        _ => { /* handle other types */ },
    }
}
```

**Options 2 and 3 are recommended** because:

- `nvs_find_key()` is designed specifically to check for key existence regardless of data type
- It's more efficient than trying multiple `nvs_get_*()` calls
- It's more robust and handles all current and future NVS data types
- It matches the semantic expectation of a `contains()` method
- **Uses the exact pattern from ESP-IDF commit [8740888](https://github.com/espressif/esp-idf/commit/8740888967e054d2033b7388db17b2b25e056ac5)** that introduced `nvs_find_key()` on Dec 8, 2023
- **Option 3 additionally provides data type information**, which can be useful for applications

**Note**: The `nvs_find_key` function was added to ESP-IDF specifically for this use case - checking key existence regardless of data type. The implementation above follows the exact error handling pattern used in the official ESP-IDF test suite.

## Environment

- **Crate (esp-idf-svc) version**: 0.51.0
- **ESP-IDF branch or tag**: release/v5.2
- **Target device (MCU)**: esp32
- **OS**: Windows 11

## Impact

This bug affects anyone using string values in NVS partitions with esp-idf-svc, making it a critical issue for production applications. It causes:

1. **Logic errors** where applications think data doesn't exist when it actually does
2. **Unnecessary re-initialization** of configuration values
3. **Data inconsistency** between different NVS operation results
4. **Debugging confusion** where data appears in partition dumps but `contains()` returns false

## Additional Notes

- The bug affects **all NVS partitions** (default and custom) and **string values**
- `get_str()`, `set_str()`, and other string operations work correctly
- Only the `contains()` method is affected
- Binary partition dumps show the string entries exist in flash
- This suggests the bug is in the esp-idf-svc wrapper, not the underlying ESP-IDF NVS implementation

danielmeza:
Fixes issue #585 where EspNvs::contains() incorrectly returned false for string values that actually existed in NVS partitions. The root cause was that contains() only checked for u64 and blob data types, never checking for string values stored with nvs_set_str().

Changes:
- Replace broken len()-based contains() with nvs_get_type() implementation
- Add contains_key_of_type() method for type-specific key checking
- Add find_key_type() method to retrieve key data types
- Add NvsDataType enum with From<nvs_type_t> conversion
- Introduce EspKeyValueStorage wrapper for consistent serialized storage behavior

The new contains() implementation uses nvs_get_type() to detect any key type, ensuring consistent behavior across all NVS data types (u8, u16, u32, u64, i8, i16, i32, i64, str, blob).

Fixes #585

### Submission Checklist 📝
- [ x I have used `cargo fmt` command to ensure that all changed code is formatted correctly.

#### Description
Please provide a clear and concise description of your changes, including the motivation behind these changes. The context is crucial for the reviewers.

#### Testing
Describe how you tested your changes.

### DIFF
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -34,37 +34,37 @@ pub struct NvsDefault(());
 
 /// A specialized key-value storage wrapper around `EspNvs` that provides a simplified interface
 /// for storing and retrieving arbitrary data as byte arrays.
-/// 
+///
 /// This struct was introduced to solve issue [#585](https://github.com/esp-rs/esp-idf-svc/issues/585)
 /// where the `contains()` method in `EspNvs` incorrectly returned `false` for string values
 /// that actually existed in NVS partitions. The root cause was that `EspNvs` implements
 /// two different storage strategies:
-/// 
+///
 /// 1. **Native ESP-IDF NVS API**: Direct access to ESP-IDF's native types (u8, u16, u32, u64, i8, i16, i32, i64, str, blob)
 /// 2. **Serialized storage**: Everything stored as either u64 (≤7 bytes) or blob (>7 bytes) for compatibility with serde
-/// 
+///
 /// `EspKeyValueStorage` focuses on the second approach, providing a clean interface for:
 /// - Storing any data that can be represented as `&[u8]`
 /// - Automatic optimization: values ≤7 bytes stored as u64, larger values as blobs
 /// - Consistent `contains()` method that works correctly with this storage strategy
 /// - Full compatibility with Rust serde implementations (postcard, json, etc.)
-/// 
+///
 /// ## Usage
-/// 
+///
 /// ```rust,no_run
 /// use esp_idf_svc::nvs::{EspDefaultNvsPartition, EspKeyValueStorage};
-/// 
+///
 /// # fn main() -> Result<(), Box<dyn std::error::Error>> {
 /// let partition = EspDefaultNvsPartition::take()?;
 /// let storage = EspKeyValueStorage::new(partition, "my_namespace", true)?;
-/// 
+///
 /// // Store data as bytes
 /// let data = b"hello world";
 /// storage.set_raw("my_key", data)?;
-/// 
+///
 /// // Check if key exists (this works correctly, unlike the original EspNvs bug)
 /// assert!(storage.contains("my_key")?);
-/// 
+///
 /// // Retrieve data
 /// let mut buffer = [0u8; 64];
 /// if let Some(retrieved) = storage.get_raw("my_key", &mut buffer)? {
@@ -73,9 +73,9 @@ pub struct NvsDefault(());
 /// # Ok(())
 /// # }
 /// ```
-/// 
+///
 /// ## Performance Characteristics
-/// 
+///
 /// - **Small values (≤7 bytes)**: Stored as u64 for efficiency
 /// - **Large values (>7 bytes)**: Stored as ESP-IDF blobs
 /// - **Memory efficient**: No unnecessary allocations for small values
@@ -411,7 +411,7 @@ impl<T: NvsPartitionId> EspNvs<T> {
         Ok(Self(partition, handle))
     }
 
-    pub fn contains(&self, name:&str) -> Result<bool, EspError> {
+    pub fn contains(&self, name: &str) -> Result<bool, EspError> {
         let c_key = to_cstring_arg(name)?;
         let mut entry_type: nvs_type_t = nvs_type_t_NVS_TYPE_ANY;
 
@@ -426,7 +426,11 @@ impl<T: NvsPartitionId> EspNvs<T> {
         }
     }
 
-    pub fn contains_key_of_type(&self, name:&str, data_type: NvsDataType) -> Result<bool, EspError> {
+    pub fn contains_key_of_type(
+        &self,
+        name: &str,
+        data_type: NvsDataType,
+    ) -> Result<bool, EspError> {
         let c_key = to_cstring_arg(name)?;
         let mut entry_type: nvs_type_t = nvs_type_t_NVS_TYPE_ANY;
 
@@ -822,7 +826,6 @@ impl<T: NvsPartitionId> RawStorage for EspNvs<T> {
         EspNvs::set_raw(self, name, buf)
     }
 }
-
 
 impl<T: NvsPartitionId> EspKeyValueStorage<T> {
     pub fn new(
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -32,6 +32,91 @@ pub trait NvsPartitionId {
 
 pub struct NvsDefault(());
 
+/// A specialized key-value storage wrapper around `EspNvs` that provides a simplified interface
+/// for storing and retrieving arbitrary data as byte arrays.
+/// 
+/// This struct was introduced to solve issue [#585](https://github.com/esp-rs/esp-idf-svc/issues/585)
+/// where the `contains()` method in `EspNvs` incorrectly returned `false` for string values
+/// that actually existed in NVS partitions. The root cause was that `EspNvs` implements
+/// two different storage strategies:
+/// 
+/// 1. **Native ESP-IDF NVS API**: Direct access to ESP-IDF's native types (u8, u16, u32, u64, i8, i16, i32, i64, str, blob)
+/// 2. **Serialized storage**: Everything stored as either u64 (≤7 bytes) or blob (>7 bytes) for compatibility with serde
+/// 
+/// `EspKeyValueStorage` focuses on the second approach, providing a clean interface for:
+/// - Storing any data that can be represented as `&[u8]`
+/// - Automatic optimization: values ≤7 bytes stored as u64, larger values as blobs
+/// - Consistent `contains()` method that works correctly with this storage strategy
+/// - Full compatibility with Rust serde implementations (postcard, json, etc.)
+/// 
+/// ## Usage
+/// 
+/// ```rust,no_run
+/// use esp_idf_svc::nvs::{EspDefaultNvsPartition, EspKeyValueStorage};
+/// 
+/// # fn main() -> Result<(), Box<dyn std::error::Error>> {
+/// let partition = EspDefaultNvsPartition::take()?;
+/// let storage = EspKeyValueStorage::new(partition, "my_namespace", true)?;
+/// 
+/// // Store data as bytes
+/// let data = b"hello world";
+/// storage.set_raw("my_key", data)?;
+/// 
+/// // Check if key exists (this works correctly, unlike the original EspNvs bug)
+/// assert!(storage.contains("my_key")?);
+/// 
+/// // Retrieve data
+/// let mut buffer = [0u8; 64];
+/// if let Some(retrieved) = storage.get_raw("my_key", &mut buffer)? {
+///     assert_eq!(retrieved, data);
+/// }
+/// # Ok(())
+/// # }
+/// ```
+/// 
+/// ## Performance Characteristics
+/// 
+/// - **Small values (≤7 bytes)**: Stored as u64 for efficiency
+/// - **Large values (>7 bytes)**: Stored as ESP-IDF blobs
+/// - **Memory efficient**: No unnecessary allocations for small values
+/// - **Flash efficient**: Optimized storage format reduces wear on flash memory
+
+pub struct EspKeyValueStorage<T>(EspNvs<T>);
+
+pub enum NvsDataType {
+    U8 = 1u8,
+    I8 = 17u8,
+    U16 = 2u8,
+    I16 = 18u8,
+    U32 = 4u8,
+    I32 = 20u8,
+    U64 = 8u8,
+    I64 = 24u8,
+    Str = 33u8,
+    Blob = 66u8,
+    Any = 255u8,
+}
+
+impl From<nvs_type_t> for NvsDataType {
+    fn from(nvs_type: nvs_type_t) -> Self {
+        match nvs_type {
+            nvs_type_t_NVS_TYPE_U8 => Self::U8,
+            nvs_type_t_NVS_TYPE_I8 => Self::I8,
+            nvs_type_t_NVS_TYPE_U16 => Self::U16,
+            nvs_type_t_NVS_TYPE_I16 => Self::I16,
+            nvs_type_t_NVS_TYPE_U32 => Self::U32,
+            nvs_type_t_NVS_TYPE_I32 => Self::I32,
+            nvs_type_t_NVS_TYPE_U64 => Self::U64,
+            nvs_type_t_NVS_TYPE_I64 => Self::I64,
+            nvs_type_t_NVS_TYPE_STR => Self::Str,
+            nvs_type_t_NVS_TYPE_BLOB => Self::Blob,
+            nvs_type_t_NVS_TYPE_ANY => Self::Any,
+            // In case a new type is added in the future, we halt here to avoid unnoticed issues
+            _ => todo!("Unknown NVS type, defaulting to Any"),
+        }
+    }
+}
+
 impl NvsDefault {
     fn new(reinit: bool) -> Result<Self, EspError> {
         let mut taken = DEFAULT_TAKEN.lock();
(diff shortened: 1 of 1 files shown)


src/nvs.rs
```diff
@@ -32,6 +32,91 @@ pub trait NvsPartitionId {
...
+/// - **Small values (≤7 bytes)**: Stored as u64 for efficiency
+/// - **Large values (>7 bytes)**: Stored as ESP-IDF blobs
+/// - **Memory efficient**: No unnecessary allocations for small values
+/// - **Flash efficient**: Optimized storage format reduces wear on flash memory
+pub struct EspKeyValueStorage<T: NvsPartitionId>(EspNvs<T>);
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u8)]
+pub enum NvsDataType {
+    U8 = 1,
```
ivmarkov, line 88: Perhaps instead of doing e.g. `U8 = 1` you could try `U8 = nvs_type_t_NVS_TYPE_U8 as_`?

src/nvs.rs
```diff
@@ -326,149 +411,68 @@ impl<T: NvsPartitionId> EspNvs<T> {
         Ok(Self(partition, handle))
     }
 
+    #[cfg(all(
```
ivmarkov, line 414: Instead of all of `contains`, `contains_key_type`, `find_key_type`, why not just one - `find_key` - which does what `find_key_type` currently does? Less is more?
- Using `find_key` for "contains such  key regardless of the type" would be as simple as `find_key()?.is_some()`
- Using `find_key` for "contains such key with type XYZ" would be `matches!(find_key()?, Some(NvsDataType::U8))`

... so I don't think we need the extra syntax sugar.

src/nvs.rs
```diff
@@ -810,28 +814,187 @@ impl RawHandle for EspNvs<NvsEncrypted> {
...
     }
 
     fn set_raw(&mut self, name: &str, buf: &[u8]) -> Result<bool, Self::Error> {
-        EspNvs::set_raw(self, name, buf)
+        EspKeyValueStorage::set_raw(self, name, buf)
+    }
+}
+
+impl<T: NvsPartitionId> EspKeyValueStorage<T> {
+    pub fn new(
```
ivmarkov, line 844: Perhaps just provide instead a `const fn new(nvs: EspNvs)` instead?

src/nvs.rs
```diff
@@ -810,28 +814,187 @@ impl RawHandle for EspNvs<NvsEncrypted> {
...
+        read_write: bool,
+    ) -> Result<Self, EspError> {
+        Ok(Self(EspNvs::new(partition, namespace, read_write)?))
+    }
+
+    pub fn partition(&self) -> &EspNvsPartition<T> {
+        &self.0 .0
+    }
+
+    pub fn contains(&self, name: &str) -> Result<bool, EspError> {
```
ivmarkov, line 856: Nit: we don't have to do it as it is a bit risky in introducing regressions, but since we have a fully functioning safe `EspNvs` API since quite some time (and now that we have your `find_key` method), I think `EspKeyValueStorage` can be expressed squarely in terms of safe API calls into `EspNvs` without calling the unsafe esp-idf-sys API underneath?

src/nvs.rs
```diff
@@ -810,28 +814,187 @@ impl RawHandle for EspNvs<NvsEncrypted> {
...
+        self.len(name).map(|v| v.is_some())
```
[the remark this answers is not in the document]
ivmarkov: Disregard copilot here, but the topic of expressing this `contains` in terms of safe EspNvs APIs stands.

src/nvs.rs
```diff
@@ -810,28 +814,187 @@ impl RawHandle for EspNvs<NvsEncrypted> {
...
+                ];
+
+                buf[..len as usize].copy_from_slice(&array[..len as usize]);
+
+                Ok(Some(&buf[..len as usize]))
+            }
+        }
+    }
+
+    pub fn set_raw(&self, name: &str, buf: &[u8]) -> Result<bool, EspError> {
```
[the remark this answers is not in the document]
ivmarkov: I don't think copilot is correct here. If we **know** that the underlying esp-idf-sys APIs are thread-safe, we can use `&self` for `EspKeyBalueStorage` even if the trait is more restrictive and wants `&mut self`. What we can do additionally though and that would be useful is to implement `RawStorage` not just for `EspKeyValueStorage`, but also for `&EspKeyValueStorage` which would allow users to trivially share a single `EspKeyValueStorage` instance across multiple places that do want the `RawStorage` trait.

src/nvs.rs
ivmarkov: Also since the `EspKeyValueStorage` "owns" the `EspNvs` instance, perhaps provide an `pub const fn esp_nvs(&self) -> &EspNvs {}` method.

.cargo/config.toml
```diff
@@ -1,6 +1,6 @@
 [build]
-target = "riscv32imc-esp-espidf"
-#target = "xtensa-esp32-espidf"
+# target = "riscv32imc-esp-espidf"
```
ivmarkov, line 2: Restore this to the original setup. You can always override with `cargo build ... --target ...`

src/nvs.rs
```diff
@@ -32,6 +32,91 @@ pub trait NvsPartitionId {
...
+/// # }
+/// ```
+///
+/// ## Performance Characteristics
+///
+/// - **Small values (≤7 bytes)**: Stored as u64 for efficiency
+/// - **Large values (>7 bytes)**: Stored as ESP-IDF blobs
+/// - **Memory efficient**: No unnecessary allocations for small values
+/// - **Flash efficient**: Optimized storage format reduces wear on flash memory
+pub struct EspKeyValueStorage<T: NvsPartitionId>(EspNvs<T>);
```
ivmarkov, line 83: Can you move this struct declaration right above its impl block? This is the current practice in the `esp-idf-*` crate.

ivmarkov (changes requested):
Yes, this is the direction. Some small suggestions and nits.

src/nvs.rs
```diff
@@ -841,160 +826,78 @@ impl<T: NvsPartitionId> RawStorage for EspKeyValueStorage<T> {
...
-            for v in buf.iter().rev() {
-                u64value <<= 8;
-                u64value |= *v as u_int64_t;
-            }
-
-            u64value <<= 8;
-            u64value |= buf.len() as u_int64_t;
-
-            esp!(unsafe { nvs_set_u64(self.0 .1, c_key.as_ptr(), u64value) })?;
+            let value = u64::from_le_bytes(
```
ivmarkov, line 893: That won't work unless the buf is _exactly_ 8 bytes, but it will never be, because you have an if above which says `if buf.len() < 8`.

You need to restore the original `for` loop with the shifting logic.

ivmarkov (changes requested):

src/nvs.rs
```diff
@@ -841,160 +826,78 @@ impl<T: NvsPartitionId> RawStorage for EspKeyValueStorage<T> {
...
+                    let mut u64value: u64 = 0;
+                    esp!(unsafe {
+                        nvs_get_u64(
+                            self.0 .1,
+                            to_cstring_arg(name)?.as_ptr(),
+                            &mut u64value as *mut _,
+                        )
+                    })?;
+                    buf.copy_from_slice(&u64value.to_le_bytes());
+                    Ok(Some(&buf[..8]))
```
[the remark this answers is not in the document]
danielmeza: Oh!! I miss that, restoring the previous code here

src/nvs.rs
ivmarkov: Just fix the `u64` serde by restoring the bitshift code. The rest looks quite ok.

src/nvs.rs
```diff
@@ -326,149 +411,68 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
-        } else {
-            esp!(result)?;
-            esp!(unsafe { nvs_commit(self.1) })?;
-
-            Ok(true)
+        let result = self.find_key_type(name)?;
+        match result {
+            Some(_) => Ok(true),
+            None => Ok(false),
         }
```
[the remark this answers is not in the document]
ivmarkov: No point, `contains` is not here to stay.

src/nvs.rs
danielmeza: Yes!! moved.

.cargo/config.toml
danielmeza: Restored

src/nvs.rs
```diff
@@ -810,28 +737,186 @@ impl RawHandle for EspNvs<NvsEncrypted> {
...
+            // Set the first byte as the length
+            u64value |= buf.len() as u_int64_t;
+
+            // Fill the rest of the u64 with the buffer values
+            for v in buf.iter().rev() {
+                u64value <<= 8;
+                u64value |= *v as u_int64_t;
+            }
+
+            u64value <<= 8;
```
ivmarkov, line 888: Something is still not quite right here... was it _really_ like this? As in... writing the len twice?

danielmeza: No! the original version write the len after the for loop, I will keep the original one, by  removing the len write before the loop.

src/nvs.rs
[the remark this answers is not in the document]
danielmeza: Building now! Bytheway I'm not able to build the svc repo locally tough I have installed the build dependencies with espup and rustup

ivmarkov:
Current CI fails because of a no-longer necessary `mut` in the examples. Can you fix this?

Also see my earlier comment on how the EspKeyValueStorage should work - ideally, it should not rely on `find_key`, and thus it should be available for earlier IDFs as well.

src/nvs.rs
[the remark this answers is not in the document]
danielmeza: @danielmeza We are almost done!!

Thank you for all your support and comments, I have learn a lot with this PR. 

I wonder what is the advantage on have the `impl<T: NvsPartitionId> RawStorage for &EspKeyValueStorage<T>` implementation, I have see a lot this pattern where a trait is implemented for the reference as well.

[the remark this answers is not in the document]
ivmarkov: You have probably seen something else called ["blanket impls"](https://www.judy.co.uk/blog/rust-traits-and-references/). 

These are very useful but are a bit different than what are we doing here. Here, we do something else but it boils down to the same oddity of Rust which is in fact a benefit, but it takes time to get used to: `T`, `&T`, `&mut T` - these are all **different types**. The fact that `RawStorage` is - say - implemented for `EspKeyValueStorage` does **not** mean it is implemented for `&EspKeyValueStorage` as well. Or for `&mut EspKeyValueStorage`. It might seem so, but this is because of Rust's [method resolution rules](https://doc.rust-lang.org/reference/expressions/method-call-expr.html), and sometimes by Deref impls too.

In our particular case at hand, implementing `RawStorage` for `&EspKeyValueStorage` allows you to treat immutable references to `EspKeyValueStorage` as implementation of the trait which works even though the trait wants `&mut self`, because this `&mut self` is in fact `&mut &EspKeyValueStorage`, which is perfectly fine.


With that said, I just realized that implementing the trait for `&EspKeyValueStorage` is not ideal, as it uses sometimes a **sequence** of operations on the actual EspNvs to achieve its goals. So if two concurrent threads are each using a `&EspKeyValueStorage` that might lead to inconsistencies, so unfortunately we have to remove this impl. :(

src/nvs.rs
demezaapc: I create 2 implementations, first one v5+ uses find_key and the seconds for v5- uses the safe api and simulate the contains like the old way but this time using the safe API. So I think it should be covered.

Or do we want to keep a single version of this?

demezaapc: Done!!

PR #590, opened 2025-07-30
Add mqtt v5 protocol implementation

danielmeza:
### Pull Request Details 📖
Initial implementation for the MQTT v5 implementation for the IDF v5.3 

This implements https://github.com/esp-rs/esp-idf-svc/pull/590
#### Description
This is the initial implementation for the MQTT v5 protocol for IDF. 

This implementation focus in the synchronous client, following the design drivers described at https://github.com/esp-rs/embedded-svc/pull/84, I try to reuse as much as possible but duplicating some parts that will made easy to branch specific protocol changes in the future, so I introduce a new client and put there most of the logic to archive the v5 features. 

Main changes compared with the current client implementation are
- New introduction of mqtt5 module and client. 
- Split creation for start by introducing a start method (I guess this should be promoted to embedded-svc)
- Introduce 2 new methods in the event to get message metadata and message user properties. Metadata and user properties are not reeded on all message to reduce processing in message that don't have these properties, this way the consumer can decide when to access to that info in a per message manear. 
- Add V5 properties config struct to describe the messages (Subscribe, Publish, Unsubscribe, Disconnect)PropertyConfig
- Introduce config: Option<*PropertyConfig> on existing Publish,Subscribe, Unsubscribe
- Add disconnect method to call disconnect and report the disconnect reason using the config: parameter described adobe. 
- Create transient types to hold the conversion logic between embedded-svc/esp-idf-sys types. 

#### Testing

Using an existing project, I was able to receive the metadata and user_properties info from the Receive event. 
Following has been tested ensure proper memory management: 
- Publish with config and user properties.
- Subscribe with config and user properties.
- Unsubscribe with config and user properties.
- Disconnect with config and user properties. 

Missing test: 
- Last will

## Curren issues: 
- When read user properties from recieved event heap are been allocated and not releasing properly. 

### Submission Checklist 📝


danielmeza:
@ivmarkov yes! I have moved to draf.

ivmarkov:
See #659

PR #592, opened 2025-08-05
Thread enhancements

ivmarkov:
## Thread Enhancements

This work is the outcome of the ongoing effort to enable `esp-idf-matter` to operate over the Thread protocol as well.

### Submission Checklist 📝

### Pull Request Details 📖

This PR is extending the existing Thread impl as follows:
- Full support for the SRP API (= Thread-specific mDNS)
- Ability to `start`/`stop` the Thread stack (the existing `ThreadDriver::run` method is removed). Note that stopping the Thread stack requires ESP IDF V5.5, or the not-yet released patch-levels V5.3.4, V5.4.3 and so on up to V5.1.X. Before [this change into ESP-IDF itself](https://github.com/espressif/esp-idf/commit/bd87b7cf5780e2c7d903358d33b872a164793d15), it was simply not possible to stop the Thread event loop in a clean manner, which was quite annoying
- `ThreadDriver::init / ThreadDriver::deinit` are gone. The initialization is done simply on thread driver (or `EspThread`) construction, and the deinitialization - when the driver is dropped.

With the addition of the `start` / `stop` methods, and with the removal of the `init` / `deinit` methods, the API of `ThreadDriver` / `EspThread` becomes quite similar to `WifiDriver` / `EspWifi` which I think is a big advantage, as switching one for the other is much more seamless.

#### Testing
Testing was done with `esp-idf-matter` itself.

### DIFF
--- a/src/netif.rs
+++ b/src/netif.rs
@@ -472,7 +472,7 @@ impl EspNetif {
         Ok(unsafe { esp_netif_is_netif_up(self.handle) })
     }
 
-    // TODO: Rename to `is_up_ipv4`
+    // TODO: Copy and rename to `is_up_ipv4` and deprecate the `is_up` variant in future
     pub fn is_up(&self) -> Result<bool, EspError> {
         if !self.is_netif_up()? {
             Ok(false)
--- a/src/netif.rs
+++ b/src/netif.rs
@@ -468,8 +468,13 @@ impl EspNetif {
         Ok(netif)
     }
 
+    pub fn is_netif_up(&self) -> Result<bool, EspError> {
+        Ok(unsafe { esp_netif_is_netif_up(self.handle) })
+    }
+
+    // TODO: Rename to `is_up_ipv4`
     pub fn is_up(&self) -> Result<bool, EspError> {
-        if !unsafe { esp_netif_is_netif_up(self.handle) } {
+        if !self.is_netif_up()? {
             Ok(false)
         } else {
             let mut ip_info = Default::default();
--- a/src/thread/srp.rs
+++ b/src/thread/srp.rs
@@ -1,0 +1,1072 @@
+use core::ffi::{c_void, CStr};
+use core::fmt::{self, Display};
+use core::marker::PhantomData;
+use core::net::{Ipv6Addr, SocketAddrV6};
+use core::ptr::addr_of_mut;
+
+use ::log::{debug, info, trace};
+
+use crate::sys::{
+    esp, esp_openthread_get_instance, otDnsTxtEntry, otError, otError_OT_ERROR_INVALID_ARGS,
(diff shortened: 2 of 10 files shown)


ivmarkov:
@Vollbrecht If you don't feel like reviewing this I can merge right away. It is anyway a bit specialized stuff.

Vollbrecht:
I already was looking into it, but currently am traveling. I think overall i think it looks like a solid refactor, though as i hinted i am currently not in the country so my time is limeted. 

One thing i am curious about, could you just briefly explain the gains of using the UnsafeCell<Box<ThreadInner>>. Before you needed two indirection, but this is all still a opauque pointer that you smuggle into the C callback context. 

Does the UnsafeCell allow you for cheaper init when you got it from the C side or what place purpose does it have.

src/thread/srp.rs
```diff
@@ -0,0 +1,1072 @@
...
+        host_info: &otSrpClientHostInfo,
+        _services: Option<&otSrpClientService>,
+        removed_services: Option<&otSrpClientService>,
+    ) {
+        trace!("Plat SRP changed callback");
+
+        self.cleanup(host_info, removed_services);
+
+        // TODO
+        // let state = self.state();
```
Vollbrecht, line 942: do you want to clean up this part before merging?

Vollbrecht:
So for the SRP state change we need something like a MultiWakerRegistration from embassy-sync to navigate around the problem you have with multiple async task? Is this a performance critical path or can we use some of the shelf async multi task solution for it?

src/netif.rs
```diff
@@ -468,8 +468,13 @@ impl EspNetif {
         Ok(netif)
     }
 
+    pub fn is_netif_up(&self) -> Result<bool, EspError> {
+        Ok(unsafe { esp_netif_is_netif_up(self.handle) })
+    }
+
+    // TODO: Rename to `is_up_ipv4`
```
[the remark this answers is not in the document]
ivmarkov: Yes. Untangling this would require quite a bit of changes (as we have the `NetifStatus` trait too), so I prefer to leave a TODO instead.

src/thread/srp.rs
ivmarkov: Done.

ivmarkov:
> So for the SRP state change we need something like a MultiWakerRegistration from embassy-sync to navigate around the problem you have with multiple async task? Is this a performance critical path or can we use some of the shelf async multi task solution for it?

No we don't need any of this. 

Just like Wifi, so is Thread publishing events for its various state changes on the system event bus. So if the user would like to be notified on such changes, she can subscribe to the system event bus and listen to these events. They are even mapped to Rust typesafe enums since the previous version already. Or the user can just use polling as I do, which for most cases is good enough. 

The commented out code was a carry-over from the bare-metal `openthread` impl, where obviously there is no system event bus, so there was a necessity to have other means to notify the user on state changes.

PR #593, opened 2025-08-07
Extended logging

ivmarkov:

### Submission Checklist 📝

#### Description

I've found out that our integrated logging - specifically - the fact that the Rust log level is hard-wired to the ESP-IDF log verbosity configuration - while powerful can be a bit intimidating to new users, in terms of understanding and configuring properly.

Furthermore, sometimes one might want to increase the logging verbosity of the Rust portion, _without_ increasing the log verbosity of the ESP IDF itself, because the latter often comes with a much-increased stack requirements.

Therefore, what this PR introduces are 3 separate ways to configure the ESP logger:
- `esp_idf_svc::log::init_from_esp_idf()` - this is the "integrated" logging where the logging verbosity of the Rust logs is wholly controlled by the ESP-IDF log settings (the `esp_idf_svc::log::EspLogger::initialize_default()` call is preserved and is equivalent)
- `esp_idf_svc::log::init(<level-filter>)` - a simple initialization of the Rust logging with a fixed logging level, irrespective of what the ESP-IDF native C log level verbosity is
- `esp_idf_svc::log::init_from_env()` - equivalent to the above, but the log level is taken from a `RUST_LOG` environment variable

Obviously, the user can also instantiate an `EspIdfLogger<T>` instance by herself and then initialize the `log` crate with it manually. This was anyway always possible, but so far the existing `EspLogger` was always hard-wired to the ESP-IDF log configuration.

#### Testing

With running an example with all possible options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fix #570 (c_char vs i8 mismatch on newer rustc toolchains)
 
 ### Added
+- Logging configuration enhanced with a simpler setup where Rust logs can be configured to have a verbosity which is
+  disconnected from the verbosity of the ESP-IDF native C logging (#593)
 - OTA: New method - `EspFirmwareInfoLoad::fetch_native` - returning the full native ESP-IDF image descriptor structures
 - Added `use_serde` feature, which enables the `use_serde` feature of `embedded-svc` crate, allowing to deserialize configuration structs.
 - OTA: Allow specifying image size to speed up erase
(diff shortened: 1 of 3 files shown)


Vollbrecht (commented):
Thanks for the PR!

I think i like it especially as this gives the user more options. And you made it backwards compatible. 

I think currently in all examples we are doing the EspLogger::initialize_default() dance. Should we update also those to teach the people the newer power or maybe create one example where we show it separately? I mean its not that hard but people just don't find that stuff without an example.

ivmarkov:
> Thanks for the PR!
> 
> I think i like it especially as this gives the user more options. And you made it backwards compatible.
> 
> I think currently in all examples we are doing the EspLogger::initialize_default() dance. Should we update also those to teach the people the newer power or maybe create one example where we show it separately? I mean its not that hard but people just don't find that stuff without an example.

Fair enough. I don't feel like changing all examples to use a different logging initialization (yet; maybe in future `init_from_env` is what most users would want), but I did a `logging.rs` example with explanations.

Once the esp-idf-* crates are released, perhaps we might change the `esp-idf-template` to `init_from_env` as well.

PR #596, opened 2025-08-12
fix http server timeouts

mirecta:

### Submission Checklist 📝
- [* ] I have updated existing examples or added new ones (if applicable).
- [ *] I have used `cargo fmt` command to ensure that all changed code is formatted correctly.
- [ *] I have used `cargo clippy` command to ensure that all changed code passes latest Clippy nightly lints.
- [ *] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-idf-svc/blob/main/esp-idf-svc/CHANGELOG.md) in the **_proper_** section.

#### Description
Now HTTP server accept timeouts from config not hardcoded 5 seconds

#### Testing
I set timeout for large file which receive exceed 5 seconds and it works now


src/http/server.rs
```diff
@@ -153,8 +153,8 @@ impl From<&Configuration> for Newtype<httpd_config_t> {
             max_resp_headers: conf.max_resp_headers as _,
             backlog_conn: 5,
             lru_purge_enable: conf.lru_purge_enable,
-            recv_wait_timeout: 5,
-            send_wait_timeout: 5,
+            recv_wait_timeout: conf.session_timeout.as_secs() as u16,
```
ivmarkov, line 156: I think `max_sessions` and `session_timeout` (which are currently unused) are remnants from the old HTTP server, which did aim at supporting HTTP sessions.

Perhaps it is not a very wise idea to reuse something called `session_timeout` for TCP socket receive/send timeout.

I would rather suggest that you introduce two additional `Option<Duration>` parameters, named `send_timeout` and `recv_timeout`, and you use those instead. And delete the `session_timeout` and `max_sessions` unused parameters.

ivmarkov (changes requested):

src/http/server.rs
ivmarkov: @mirecta ?

Issue #587, opened 2025-07-20
MQTT client hard crash when configuring a last will message on esp32


ApolloRoboto:
## Bug description

MQTT client hard crash when configuring a last will message on esp32 when creating the client.

Full stack trace here:

```txt
Guru Meditation Error: Core  0 panic'ed (LoadProhibited). Exception was unhandled.

Core  0 register dump:
PC      : 0x400014dc  PS      : 0x00060830  A0      : 0x80001459  A1      : 0x3ffbba30
A2      : 0x00000001  A3      : 0xfffffffd  A4      : 0x000000ff  A5      : 0x0000ff00
A6      : 0x00ff0000  A7      : 0xff000000  A8      : 0x8008b6f2  A9      : 0x3ffbba00
A10     : 0x3ffafc50  A11     : 0x3ffc870c  A12     : 0x00000019  A13     : 0x3ffc89ac
A14     : 0x00000000  A15     : 0x656e696c  SAR     : 0x00000016  EXCCAUSE: 0x0000001c
EXCVADDR: 0x00000001  LBEG    : 0x4000c2e0  LEND    : 0x4000c2f6  LCOUNT  : 0x00000000


Backtrace: 0x400014d9:0x3ffbba30 0x40001456:0x3ffbba40 0x40001445:0x3ffbba60 0x40142882:0x3ffbba80 0x40142ebc:0x3ffbbac0 0x400db13c:0x3ffbbb10 0x400d5a68:0x3ffbbce0 0x400d6fdb:0x3ffbbd00 0x40199a93:0x3ffbbec0 0x400d6dec:0x3ffbbee0 0x400df894:0x3ffbbf00 0x400df837:0x3ffbbf30 0x400ea87b:0x3ffbbf70 0x400d6ddc:0x3ffbbfb0 0x400d713b:0x3ffbbfe0 0x400ded27:0x3ffbc000 0x401a0ef3:0x3ffbc020
0x40142882 - esp_mqtt_set_config
    at /mnt/c/Users/Apollo/Repositories/rust-esp32-led-panel/.embuild/espressif/esp-idf/v5.3.2/components/mqtt/esp-mqtt/mqtt_client.c:464
0x40142ebc - esp_mqtt_client_init
    at /mnt/c/Users/Apollo/Repositories/rust-esp32-led-panel/.embuild/espressif/esp-idf/v5.3.2/components/mqtt/esp-mqtt/mqtt_client.c:867
[...]

PR #597, opened 2025-08-13
Support for last-will messages with empty payload

ivmarkov:

### Submission Checklist 📝

#### Description

ESP-IDF's MQTT client has the weird oddity that if we pass a non-null pointer to the LWT payload - yet - we pass a length of 0 - then it treats the payload as if it is an ASCIIZ (C) string, and excercises `strdup` and friends on it, which leads to out of bound reads.

The fix is to pass a NULL pointer when our LWT payload is zero-length.

#### Testing
n/a


ivmarkov:
Should fix #587

Vollbrecht (approved):
LGTM Thanks!

PR #599, opened 2025-08-16
Implement `clear` and `iter` functionality for `EspNvs`

### LINKED ISSUE #340: Expand EspNvs impl with new functions (remove all keys, iterator)
There are some interesting functions that I think are interesting to have at EspNvs level that are already defined in the bindings like:
 - nvs_erase_all https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/storage/nvs_flash.html#_CPPv413nvs_erase_all12nvs_handle_t
 - Iterator based key/value acess (nvs_entry_find, nvs_entry_next...) https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/storage/nvs_flash.html#_CPPv414nvs_entry_findPKcPKc10nvs_type_tP14nvs_iterator_t

Do you think they could be added to the EspNvs impl?

Luro02:

### Submission Checklist 📝

#### Description

This PR adds support for:

##### 1. Obtaining a handle for `EspNvs<Default>`

I guess this was forgotten? It is nice to have when calling `esp-idf-sys` functions.

##### 2. Added `EspNvs::clear`

I use a library that stores some data in the `EspNvs`, and I want to empty all stored data in some circumstances. Previously there wasn't any functionality to accomplish that, which is why I made this PR.

##### 3. Added `EspNvs::iter`

This function allows iterating over all stored keys. I don't need this function, but #340 requests this function and the `erase_all`, so I added it to close that issue with this PR

Regarding the implementation: It seems to work, but I don't know how the API should look like, couldn't find many references.

The main reference was [this example in the esp-idf docs](https://docs.espressif.com/projects/esp-idf/en/v5.5/esp32/api-reference/storage/nvs_flash.html#_CPPv424nvs_entry_find_in_handle12nvs_handle_t10nvs_type_tP14nvs_iterator_t):
```C
// Example of listing all the key-value pairs of any type under specified handle (which defines a partition and namespace)
 nvs_iterator_t it = NULL;
 esp_err_t res = nvs_entry_find_in_handle(<nvs_handle>, NVS_TYPE_ANY, &it);
 while(res == ESP_OK) {
nvs_entry_info_t info;
     nvs_entry_info(it, &info); // Can omit error check if parameters are guaranteed to be non-NULL
     printf("key '%s', type '%d' \n", info.key, info.type);
     res = nvs_entry_next(&it);
 }
 nvs_release_iterator(it);
```
The current code replicates the functionality, ignoring any errors that occur while iterating.

I do not know whether this is desirable or not.

#### Testing

Used it in a separate project, uploaded to the ESP32 and checked for any crashes.

----

This PR closes #340.

### DIFF
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -736,6 +736,13 @@ impl RawHandle for EspNvs<NvsEncrypted> {
         self.1
     }
 }
+impl RawHandle for EspNvs<NvsDefault> {
+    type Handle = nvs_handle_t;
+
+    fn handle(&self) -> Self::Handle {
+        self.1
+    }
+}
 
 /// A specialized key-value storage wrapper around `EspNvs` that provides a simplified interface
 /// for storing and retrieving arbitrary data as byte (`u8`) slices.
--- a/src/nvs.rs
+++ b/src/nvs.rs
@@ -1,4 +1,6 @@
 //! Non-Volatile Storage (NVS)
+#[cfg(esp_idf_version_at_least_5_2_0)]
+use core::marker::PhantomData;
 use core::ptr;
 
 extern crate alloc;
@@ -46,24 +48,25 @@ pub enum NvsDataType {
     I64 = nvs_type_t_NVS_TYPE_I64,
     Str = nvs_type_t_NVS_TYPE_STR,
     Blob = nvs_type_t_NVS_TYPE_BLOB,
-    Any = nvs_type_t_NVS_TYPE_ANY,
 }
 
 #[allow(non_upper_case_globals)]
-impl From<nvs_type_t> for NvsDataType {
-    fn from(nvs_type: nvs_type_t) -> Self {
+impl NvsDataType {
+    /// Converts a `nvs_type_t` to an `NvsDataType`, returning `None` if the type is not recognized.
+    #[must_use]
+    pub fn from_nvs_type(nvs_type: nvs_type_t) -> Option<Self> {
         match nvs_type {
-            nvs_type_t_NVS_TYPE_U8 => NvsDataType::U8,
-            nvs_type_t_NVS_TYPE_I8 => NvsDataType::I8,
-            nvs_type_t_NVS_TYPE_U16 => NvsDataType::U16,
-            nvs_type_t_NVS_TYPE_I16 => NvsDataType::I16,
-            nvs_type_t_NVS_TYPE_U32 => NvsDataType::U32,
-            nvs_type_t_NVS_TYPE_I32 => NvsDataType::I32,
-            nvs_type_t_NVS_TYPE_U64 => NvsDataType::U64,
-            nvs_type_t_NVS_TYPE_I64 => NvsDataType::I64,
-            nvs_type_t_NVS_TYPE_STR => NvsDataType::Str,
-            nvs_type_t_NVS_TYPE_BLOB => NvsDataType::Blob,
-            _ => NvsDataType::Any,
+            nvs_type_t_NVS_TYPE_U8 => Some(Self::U8),
+            nvs_type_t_NVS_TYPE_I8 => Some(Self::I8),
+            nvs_type_t_NVS_TYPE_U16 => Some(Self::U16),
+            nvs_type_t_NVS_TYPE_I16 => Some(Self::I16),
+            nvs_type_t_NVS_TYPE_U32 => Some(Self::U32),
+            nvs_type_t_NVS_TYPE_I32 => Some(Self::I32),
+            nvs_type_t_NVS_TYPE_U64 => Some(Self::U64),
+            nvs_type_t_NVS_TYPE_I64 => Some(Self::I64),
+            nvs_type_t_NVS_TYPE_STR => Some(Self::Str),
+            nvs_type_t_NVS_TYPE_BLOB => Some(Self::Blob),
+            _ => None,
         }
     }
 }
@@ -373,7 +376,7 @@ impl<T: NvsPartitionId> EspNvs<T> {
         let result = unsafe { nvs_find_key(self.1, c_key.as_ptr(), &mut entry_type as *mut _) };
 
         match result {
-            ESP_OK => Ok(Some(NvsDataType::from(entry_type))),
+            ESP_OK => Ok(NvsDataType::from_nvs_type(entry_type)),
             ESP_ERR_NVS_NOT_FOUND => Ok(None),
             err => {
                 esp!(err)?;
@@ -710,7 +713,7 @@ impl<T: NvsPartitionId> EspNvs<T> {
     }
 
     /// Erases all key-value pairs in the NVS namespace.
-    /// 
+    ///
     /// # Errors
     ///
     /// This function will return an error if the NVS erase operation fails, this can happen because of
@@ -723,6 +726,56 @@ impl<T: NvsPartitionId> EspNvs<T> {
         esp!(unsafe { nvs_commit(self.1) })?;
 
         Ok(())
+    }
+
+    /// Returns struct to iterate over all keys stored in this NVS namespace with the specified data type.
+    ///
+    /// A data type of `None` will return all keys regardless of their type.
+    ///
+    /// # Mutating the NVS while iterating
+    ///
+    /// Both this function and others that mutate the NVS like [`EspNvs::remove`] only require an immutable
+    /// reference, making it possible to mutate the NVS while iterating over it. For example, one could remove
+    /// keys while iterating.
+    ///
+    /// It is **not** recommended to do this, because the iterator might skip keys. It will not result in
+    /// a panic or undefinied behavior.
+    ///
+    /// # Errors
+    ///
+    /// This function will return an error if
+    /// - there is no memory available for allocation of internal structures
+    /// - for some reason the [`EspNvs::handle`] is invalid (should not happen)
+    #[cfg(esp_idf_version_at_least_5_2_0)]
+    pub fn keys(&self, data_type: Option<NvsDataType>) -> Result<EspNvsKeys<'_>, EspError> {
+        let mut raw_iter: nvs_iterator_t = core::ptr::null_mut();
+
+        match unsafe {
+            nvs_entry_find_in_handle(
+                self.1,
+                data_type
+                    .map(|ty| ty as u32)
+                    .unwrap_or(nvs_type_t_NVS_TYPE_ANY),
+                &mut raw_iter as *mut _,
+            )
+        } {
+            ESP_ERR_NVS_NOT_FOUND => {
+                return Ok(EspNvsKeys {
+                    _nvs: PhantomData,
+                    raw_iter: core::ptr::null_mut(),
+                    is_exhausted: true,
+                    key_name_buffer: [0; 16],
+                });
+            }
+            other => esp!(other)?,
+        }
+
+        Ok(EspNvsKeys {
+            _nvs: PhantomData,
+            raw_iter,
+            is_exhausted: false,
+            key_name_buffer: [0; 16],
+        })
     }
 }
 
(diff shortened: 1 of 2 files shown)


Luro02:
I will rebase the PR once all issues are resolved. I wasn't able to run clippy over the project yet, because of
```
error: failed to run custom build command for `esp-idf-sys v0.36.1 (https://github.com/esp-rs/esp-idf-sys#2749423f)`

Caused by:
  process didn't exit successfully: `C:\projects\esp-idf-svc\target\release\build\esp-idf-sys-8d4f389eaf43e703\build-script-build` (exit code: 1)
  --- stderr
  Error: Too long output directory: `\\?\C:\projects\esp-idf-svc\target\riscv32imc-esp-espidf\release\build\esp-idf-sys-8392932dfa9bd84c\out`. Shorten your project path down to no more than 10 characters (or use WSL2 and its native Linux filesystem). Note that tricks like Windows `subst` do NOT work!
warning: build failed, waiting for other jobs to finish...
```
So I am waiting for CI to show me any issues in that regard. (Yes, I know about the workarounds, but I am hoping that the CI will suffice for this PR)

src/nvs.rs
```diff
@@ -708,6 +708,36 @@ impl<T: NvsPartitionId> EspNvs<T> {
...
+    pub fn erase_all(&self) -> Result<(), EspError> {
+        esp!(unsafe { nvs_erase_all(self.1) })?;
+
+        esp!(unsafe { nvs_commit(self.1) })?;
+
+        Ok(())
+    }
+
+    /// Returns an iterator over all key-value pairs in the NVS namespace with the specified data type.
+    pub fn iter(&self, data_type: NvsDataType) -> Result<IterEspNvs<'_, T>, EspError> {
```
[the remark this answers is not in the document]
Luro02: I don't think so, I forgot that the functions that should be mutable are immutable for this type. I think I will change it to `&mut self`, which would guarantee that there are no other references while the iterator is alive?

ivmarkov:
> I will rebase the PR once all issues are resolved. I wasn't able to run clippy over the project yet, because of
> 
> ```
> error: failed to run custom build command for `esp-idf-sys v0.36.1 (https://github.com/esp-rs/esp-idf-sys#2749423f)`
> 
> Caused by:
>   process didn't exit successfully: `C:\projects\esp-idf-svc\target\release\build\esp-idf-sys-8d4f389eaf43e703\build-script-build` (exit code: 1)
>   --- stderr
>   Error: Too long output directory: `\\?\C:\projects\esp-idf-svc\target\riscv32imc-esp-espidf\release\build\esp-idf-sys-8392932dfa9bd84c\out`. Shorten your project path down to no more than 10 characters (or use WSL2 and its native Linux filesystem). Note that tricks like Windows `subst` do NOT work!
> warning: build failed, waiting for other jobs to finish...
> ```
> 
> So I am waiting for CI to show me any issues in that regard. (Yes, I know about the workarounds, but I am hoping that the CI will suffice for this PR)

(Ab)using the CI is fine, but shortening `c:\projects` to `c:\p` you need to do for all your other projects anyway. Or give up on native Windows and use WSL2.

src/nvs.rs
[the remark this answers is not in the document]
ivmarkov: No, you need to inspect the nvs source code inside ESP-IDF. It _might_ be safe and then `&self` would be preferred.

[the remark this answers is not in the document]
Luro02: So I am allowed to make assumptions about the source code? I was thinking that the safety should not rely on assumptions that might change in the future.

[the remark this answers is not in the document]
ivmarkov: Why do you think they will change in future?

src/nvs.rs
```diff
@@ -736,6 +767,72 @@ impl RawHandle for EspNvs<NvsEncrypted> {
...
 }
+impl RawHandle for EspNvs<NvsDefault> {
+    type Handle = nvs_handle_t;
+
+    fn handle(&self) -> Self::Handle {
+        self.1
+    }
+}
+
+#[cfg(esp_idf_version_at_least_5_2_5)]
```
ivmarkov, line 778: I don't think this is the minimum version where `nvs_entry_find_in_handle` appears. Did you actually check this in the ESP-IDF source code?

src/nvs.rs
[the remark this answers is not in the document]
Luro02: The iterator keeps an index to the current entry and a reference to their underlying storage type:
```c
struct nvs_opaque_iterator_t
{
    nvs_type_t type;
    uint8_t nsIndex;
    size_t entryIndex;
    nvs::Storage *storage;
    intrusive_list<nvs::Page>::iterator page;
    nvs_entry_info_t entry_info;
};
```
https://github.com/espressif/esp-idf/blob/4e036983a751e4667ade94c8f6f6bf1e7f78eff0/components/nvs_flash/src/nvs_storage.hpp#L169-L177

When an entry is fetched, it calls
https://github.com/espressif/esp-idf/blob/4e036983a751e4667ade94c8f6f6bf1e7f78eff0/components/nvs_flash/src/nvs_api.cpp#L811

which accesses `findEntryNs`
```C
    bool entryFound = handle_obj->findEntryNs(it);
    if (!entryFound) {
        free(it);
        *output_iterator = nullptr;
        return ESP_ERR_NVS_NOT_FOUND;
    }
```
The `findEntryNs` will then call [`findItem` on the pages](https://github.com/espressif/esp-idf/blob/4e036983a751e4667ade94c8f6f6bf1e7f78eff0/components/nvs_flash/src/nvs_page.cpp#L902-L911):
```C
esp_err_t Page::findItem(uint8_t nsIndex, ItemType datatype, const char* key, size_t &itemIndex, Item &item, uint8_t chunkIdx, VerOffset chunkStart)
{
    if (mState == PageState::CORRUPT || mState == PageState::INVALID || mState == PageState::UNINITIALIZED) {
        return ESP_ERR_NVS_NOT_FOUND;
    }

    size_t findBeginIndex = itemIndex;
    if (findBeginIndex >= ENTRY_COUNT) {
        return ESP_ERR_NVS_NOT_FOUND;
    }

// ...
```
The function handles the case where the index is invalid correctly, therefore I would assume that it is fine to remove entries while it is iterating over it? But it might skip some keys?

src/nvs.rs
```diff
@@ -736,6 +806,89 @@ impl RawHandle for EspNvs<NvsEncrypted> {
...
+                Some((
+                    from_cstr(&info.key[..]).try_into().unwrap(),
+                    NvsDataType::from_nvs_type(info.type_),
+                ))
+            }
+            // The nvs_entry_info only fails if any of the arguments are null.
+            // The nvs_entry_info is never null, and self.raw_iter is checked for null before the invocation.
+            //
+            // Therefore this should never happen.
+            err => unimplemented!(
```
Vollbrecht, line 869: since we think it never happens, maybe we should use unreachable here, instead of unimplemented?

Luro02:
> You have to protect all newer APIs which are not available on older ESP-IDF versions with `esp_idf_version_at_least_X_Y_Z` and with `esp_idf_version_patch_at_least_X_Y_Z`.
> 
> See https://github.com/esp-rs/esp-idf-sys/blob/2749423ff0615e4162ac0ca6cd559cae5f418506/build/common.rs#L193

Regarding this, I am having trouble compiling the code with these config features.
The esp-idf version 5.2.0 seems to be the lowest available version, which is why I would cfg everything with `#[cfg(any(esp_idf_version_patch_at_least_5_2_0, esp_idf_version_at_least_5_2_0))]`.

In my local project, it doesn't set these config variables, making it impossible to use them. To test if this issue is limited to my setup, I added the `esp_idf_version*_at_least_5_2_0` only on the import, which would make CI fail if the cfg wasn't set.

Are there any specific things to set for it to work? Should I open an issue for this?

src/nvs.rs
```diff
@@ -746,10 +746,7 @@ impl<T: NvsPartitionId> EspNvs<T> {
     /// This function will return an error if
     /// - there is no memory available for allocation of internal structures
     /// - for some reason the [`EspNvs::handle`] is invalid (should not happen)
-    #[cfg(all(
-        esp_idf_version_major = "5",
-        not(any(esp_idf_version_minor = "1", esp_idf_version_minor = "0"))
-    ))]
+    #[cfg(any(esp_idf_version_patch_at_least_5_2_0, esp_idf_version_at_least_5_2_0))]
```
ivmarkov, line 749: You don't need `esp_idf_version_patch_at_least_5_2_0`. for your particular case - where the functionality is available for ALL ESP IDFs >= 2.2.0 - `esp_idf_version_at_least_5_2_0` is sufficient.

The `esp_idf_version_patch_at_least_*` variants are useful only if Espressif decides to back-port some functionality in an earlier ESP-IDF release. Say, in ESP IDF 5.1 and it will pop-up there after patch 5. Then you need to do `any(esp_idf_version_patch_at_least_5_1_5, esp_idf_version_at_least_5_2_0)` 
which means "either from V5.1.5 till V5.1.max, OR from V5.2.2 onwards".

src/nvs.rs
```diff
@@ -814,10 +811,7 @@ impl RawHandle for EspNvs<NvsDefault> {
     }
 }
 
-#[cfg(all(
-    esp_idf_version_major = "5",
-    not(any(esp_idf_version_minor = "1", esp_idf_version_minor = "0"))
-))]
+#[cfg(any(esp_idf_version_patch_at_least_5_2_0, esp_idf_version_at_least_5_2_0))]
```
ivmarkov, line 814: Ditto here and everywhere below.

src/nvs.rs
```diff
@@ -736,6 +803,77 @@ impl RawHandle for EspNvs<NvsEncrypted> {
...
+    // The EspNvs must not be dropped while the iterator is still in use,
+    // this reference ensures that.
+    _nvs: PhantomData<&'a ()>,
+    raw_iter: nvs_iterator_t,
+    is_exhausted: bool,
+}
+
+#[cfg(any(esp_idf_version_patch_at_least_5_2_0, esp_idf_version_at_least_5_2_0))]
+impl<'a> Iterator for EspNvsKeys<'a> {
+    type Item = (heapless::String<16>, Option<NvsDataType>);
```
ivmarkov, line 825: Returning here `Option<NvsDataType>` is NOK.

The whole reason why I asked to remove the `From<nvs_type_t>` for `NvsDataType` is to make sure that our Rust `NvsDataType` does **not** contain a variant representing "any data type". The "any" data type does not make sense _exactly_ here. What does it mean a key item to have a name and have 
"any" data type? There is no such thing in ESP-IDF NVS?

You want to express the "any" semantics only when you are **creating** the iterator, and there, having the `Option<NvsDataType>` does make a total sense. Because you might want to say "iterate over all keys of type BLOB", i.e. pass `Some(NvsDataType::Blob)`. But you might _also_ want to say "iterate over all keys of ANY type". Which is modeled by passing a `None`.

But regardless what you said when creating the iterator, the iterator would always yield keys with a specific name and a specific,  **concrete** type.

ivmarkov, line 825: One more detail here that I did not mention originally:

It is not ideal to return `heapless` types from the `esp-idf-*` APIs. For one, `heapless` is not stable yet, we are using 0.8, but 0.9 is just around the corner. We don't want to bump our version when a new `heapless` version comes and we decide to use it.

We are violating our own rule in the `Wifi` and OTA APIs (and there is a bug for this) but for all more recent APIs I try to follow this and no longer expose `heapless` types.

So what do we want instead? An `&'a str`. 
But from where the `'a` would come? Ideally, it should be the lifetime of the `next` method `&mut` borrow, as in 
`fn next(&'a mut self) -> Option<Self::Item<'a>>` but this does not work. Putting aside the trickier reasons why, the other reason is that the associated type `Iterator::Item` is not lifetimed, because GATs came to Rust when `Iterator` already did exist.

What else can we do? Let's agree on the following: I'll merge the PR with `heapless::String` and then I'll do a small surgery to change the String into an `&'a str`. This means we won't be able anymore to claim that `EspNvsKeys` is an `Iterator` (as it will have an incompatible signature), but that I think is the better tradeoff compared to continuing to return `heapless` types from our APIs.

You would still be able to iterate over the keys with `while let Some(key) = keys.next() {}` Only the `for` loop won't work as it needs `keys` to be an `Iterator`. But once lending iterators land in Rust (whenever), we'll implement those for our use case.

ivmarkov (changes requested):
We are getting there. I think this is the last round of comments. After that, we should be able to merge.

ivmarkov:
@Luro02 Please also update the CHANGELOG.md file

src/nvs.rs
Luro02: Yes, that was my thought process too. I decided that an `Iterator` would be preferable over a standalone function, but I understand why exposing `heapless` is suboptimal. I adjusted the implementation.

src/nvs.rs
```diff
@@ -736,6 +805,80 @@ impl RawHandle for EspNvs<NvsEncrypted> {
...
+}
+
+#[cfg(esp_idf_version_at_least_5_2_0)]
+pub struct EspNvsKeys<'a> {
+    // The EspNvs must not be dropped while the iterator is still in use,
+    // this reference ensures that.
+    _nvs: PhantomData<&'a ()>,
+    raw_iter: nvs_iterator_t,
+    is_exhausted: bool,
+    key_name_buffer: [u8; 16],
```
ivmarkov, line 823: Why 16 and not 11?

src/nvs.rs
```diff
@@ -736,6 +805,80 @@ impl RawHandle for EspNvs<NvsEncrypted> {
...
+impl<'a> EspNvsKeys<'a> {
+    /// Returns the next key in the NVS namespace and its data type.
+    ///
+    /// After the last key is returned, this function will return `None` on subsequent calls.
+    pub fn next_key(&mut self) -> Option<(&str, NvsDataType)> {
+        if self.is_exhausted || self.raw_iter.is_null() {
+            return None;
+        }
+
+        let mut info: nvs_entry_info_t = Default::default();
```
ivmarkov, line 836: Micro opt for memory: you can just promote `nvs_entry_info_t` as a member to `EspNvsKeys` instead of having a separate key buffer, but oh well. I'll merge as -is and then maybe fix myself. No big deal. :)

src/nvs.rs
Luro02: They key buffer in nvs_info is 16 byte large

Luro02: See https://docs.esp-rs.org/esp-idf-sys/esp_idf_sys/struct.nvs_entry_info_t.html

Luro02:
Thank you for the kind and helpful reviews.

PR #611, opened 2025-09-18
Fix BLE support for the s3/c3

ivmarkov:
### Submission Checklist 📝

#### Description

Not sure whether the BLE support for the s3 did work in earlier ESP-IDF versions, but it definitely does not work in ESP-IDF 5.3+, as a bunch of config parameters are missing. Likely the same situation for the c3, though for that one I'm sure it did work in earlier ESP-IDF versions.

The PR adds the missing configs, and also unites the configuration of the c3 and the s3, because ESP-IDF does the same - reuses the c3 configs for the s3).

#### Testing
`esp-idf-matter`

EDIT: Also, a bunch of configs were added for the stock esp32.

### DIFF
--- a/.github/configs/sdkconfig.defaults
+++ b/.github/configs/sdkconfig.defaults
@@ -73,5 +73,8 @@ CONFIG_OPENTHREAD_ENABLED=y
 # Border Router again, mDNS
 #CONFIG_MDNS_MULTIPLE_INSTANCE=y
 
+# Border Router again, Wifi/IEEE802.15.4 software co-exist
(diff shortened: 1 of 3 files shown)


Vollbrecht (approved):
Just noticed, it seams pretty annoying that we dont get boolean types returned from bindgen for the respecitf fields in esp_bt_controller_config_t. My head currently not remembering correctly but is that because on the C side they are all  `#defined` and those have no real type info right?  

Anyway looks good to me. Thanks!

PR #627, opened 2025-12-16
BT Client: search functions not passing a valid pointer to BtUuid(esp_bt_uuid_t)

ferdy-lw:
### Submission Checklist 📝

#### Description

In my other project I was getting errors, BT_BTC `bta_to_btc_uuid` error, when using the client `gatt::client::EspGattc.search_service()` and  `gatt::client::EspGattc.get_service()` functions. This was due to using `BtUuid.raw()` returning a copy of the wrapped `esp_bt_uuid_t` which was being dropped. 

#### Testing
Updated bt_gatt_client, which _was_ working, and tested on another local project that was consistently failing.

### DIFF
--- a/src/bt/ble/gatt/client.rs
+++ b/src/bt/ble/gatt/client.rs
@@ -1286,7 +1286,7 @@ where
                 conn_id,
                 start_handle,
                 end_handle,
-                char_uuid.into_raw(),
+                *char_uuid.raw(),
                 results as *const _ as *mut _,
                 &mut count,
             )
@@ -1327,8 +1327,8 @@ where
                 conn_id,
                 start_handle,
                 end_handle,
-                char_uuid.into_raw(),
-                descr_uuid.into_raw(),
+                *char_uuid.raw(),
+                *descr_uuid.raw(),
                 results as *const _ as *mut _,
                 &mut count,
             )
@@ -1363,7 +1363,7 @@ where
                 gattc_if,
                 conn_id,
                 char_handle,
-                descr_uuid.into_raw(),
+                *descr_uuid.raw(),
                 results as *const _ as *mut _,
                 &mut count,
             )
--- a/src/bt.rs
+++ b/src/bt.rs
@@ -81,7 +81,7 @@ impl BtUuid {
         &self.0
     }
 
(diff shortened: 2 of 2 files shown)


src/bt/ble/gatt/client.rs
```diff
@@ -1122,13 +1122,13 @@ where
         &self,
         gattc_if: GattInterface,
         conn_id: ConnectionId,
-        filter_uuid: Option<BtUuid>,
+        filter_uuid: Option<&BtUuid>,
```
ferdy-lw: Should the other get functions that take a BtUuid also be changed to take a reference instead (and keep the api consistent)? i.e. char_uuid: BtUuid, ->  char_uuid: `&`BtUuid


```
 pub fn get_characteristic_by_uuid(
        &self,
        gattc_if: GattInterface,
        conn_id: ConnectionId,
        start_handle: Handle,
        end_handle: Handle,
        char_uuid: BtUuid,
        results: &mut [CharacteristicElement],
    ) -> Result<usize, GattStatus> 
```

ferdy-lw:
Using `BtUuid.raw()` does a copy on the wrapped `esp_bt_uuid_t`, when using it in a closure of the map I believe it's creating a temporary copy, and returning a reference to that copy, which is being dropped before the `esp_ble_gattc_search_service` can use the reference to internally 'copy', causing it to panic. Using `BtUuid.0` doesn't do a copy of `esp_bt_uuid_t`, it uses a reference to `BtUuid`. 

If I didn't use a closure, the copy of `BtUuid.raw()` probably wouldn't be removed before the esp function could us it. I think it's better to pass in a  `BtUuid` reference in anyway, and not do any copies at all.

```rust
 esp!(unsafe {
            esp_ble_gattc_get_service(
                gattc_if,
                conn_id,
                svc_uuid.map_or(core::ptr::null_mut(), |s| &s.0 as *const _ as *mut _),
                results as *const _ as *mut _,
                &mut count,
                offset,
            )
        })?;
```

ivmarkov:
(We can also have `into_raw(self)` or `into_inner(self)` that returns the raw value, but not sure we would ever need that.)

PR #636, opened 2026-02-07
Fix EspWebSocketClientConfig timeout fields not passed to C struct (#635)

### LINKED ISSUE #635: EspWebSocketClientConfig: reconnect_timeout_ms and network_timeout_ms are not passed to the C config
## Bug description

The `EspWebSocketClientConfig` struct exposes `reconnect_timeout_ms` and `network_timeout_ms` as public fields, but the `TryFrom` implementation that converts the config to `esp_websocket_client_config_t` never assigns them to the C struct. They  fall through to `..Default::default()`, which zeroes them out, causing the C library to silently fall back to its 10-second defaults regardless of what the user sets.

- Would you like to work on a fix? [y/n]

Yes, absolutely

## To Reproduce

```
let config = EspWebSocketClientConfig {
      reconnect_timeout_ms: Duration::from_secs(5),
      network_timeout_ms: Duration::from_secs(30),
      ..Default::default()
  };

  // Both values are silently ignored - the C library always uses its 10s defaults
  let client = EspWebSocketClient::new(url, &config, timeout, callback)?;
```

## Expected behavior

The client should use the values passed in by the user.

## Environment

  - Crate (`esp-idf-svc`) version: 0.51.0
  - ESP-IDF branch or tag: v5.4
  - Target device (MCU): esp32c6
  - OS: Linux (openSUSE Tumbleweed)

awalland:
Fixes #635

### Submission Checklist 📝

#### Description
 The `EspWebSocketClientConfig` struct exposes `reconnect_timeout_ms` and `network_timeout_ms` as public fields, but the `TryFrom` implementation that converts the config to `esp_websocket_client_config_t` never assigns them to the C struct. They fall through to `..Default::default()`, which zeroes them out, causing the C library to silently fall back to its 10-second defaults regardless of what the user sets.

This fix adds the two missing assignments in the struct initializer:

```
  reconnect_timeout_ms: conf.reconnect_timeout_ms.as_millis() as _,
  network_timeout_ms: conf.network_timeout_ms.as_millis() as _,
```

This is a no-op for existing code that uses `Default`, since `Duration::default()` is zero and the C library already treats zero as "use the default". It only changes behavior for users who explicitly set these fields which previously  had no effect.

#### Testing
I came across the bug by using the WS client in a project and I am already using the fixed version (vendored).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bluetooth support is no longer behind the `experimental` feature
 - Thread support is no longer behind the `experimental` feature
 - MQTT: Fix a crash when the LWT payload is empty (#597)
+- WS: Fix `EspWebSocketClientConfig` `reconnect_timeout_ms` and `network_timeout_ms` fields not being passed to the underlying C struct (#635)
 
 ### Added
 - Compatibility with ESP-IDF V5.4.x and V5.5.x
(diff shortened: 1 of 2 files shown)


ivmarkov:
Thanks!

PR #638, opened 2026-02-10
Add task_caps to http server `Configuration`

JacksonUtsch:

### Submission Checklist 📝

### Pull Request Details 📖
Exposes the task_caps configuration option on http server

#### Description
In http server `Configuration` exposes `task_caps`. Specifically my motivation is to be able to use `MALLOC_CAP_SPIRAM` to save lots of internal memory.

Could make an `Option` in order to prevent breaking change but this is more code correct.

#### Testing
Ran locally with changes, small change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - NVS: Removed `NvsDataType::Any` and replaced `From<nvs_type_t>` with `NvsDataType:from_nvs_type`
 - (#529) `Peripheral` and `PeripheralRef` removed and replaced with a simple pattern similar to the `esp-hal` one.
   - Check https://github.com/esp-rs/esp-idf-hal/pull/529 for details on that change
+- HTTP: added `task_caps` option into server `Configuration`
 
 ### Fixed
 - `WifiDriver::get_ap_info` not takes `&self` instead of `&mut self`. Convenience method `EspWifi::get_ap_info` that delegates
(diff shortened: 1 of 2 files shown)


JacksonUtsch:
Any feedback?

ivmarkov:
> Thank you for the response @ivmarkov.
> 
> The amount of RAM saved is equivalent to the stack size. It seems the stack size limits the possible payload size for the HTTP server. In my case I'd like to support quite large payloads 10s of KBs. For me, this is very important to move off of the internal heap as I am already at its limit (- ~2KBs).

I suggest increasing the stack size but still keeping it in the internal RAM for now, so as not to open the can of worms below.

As for the stack size limiting the possible payload size - no, that's incorrect. This is only true if **you** allocate - say - a large `[u8; 50000]` on-stack in your callback handler and use this as your buffer for getting the payload, but this is a bit too naive.

An alternative approach would be to just use a `Vec` to accumulate the payload. The Vec's content is allocated on the Rust heap = internal SRAM.

An even better approach is to read the HTTP request payload (I assume this is the POST payload you are referring to) in a **streaming** fashion. If I'm not mistaken, this was allowed by the ESP IDF HTTP server and my bindings (granted, haven't used this guy since a long time).

> 
> The documentation regarding SPIRAM safety does not seem exhaustive I have learned. But so far in my usage I have not seen any issues moving this stack to the SPIRAM. As this would be an explicit use case declaring `MALLOC_CAP_SPIRAM` I think it is fair to allow users to deal with any consequences and that could be noted in the added parameter.
> 
> I have not seen any issues with it yet, I just forward the payloads to a channel essentially and handle in a separate task.
> 
> Using the esp32s3

Well up to you - you have been warned :p The issue is very real, even if it does not result in an immediate crash. Which makes it so much worse.

I'll merge this because it is anyway backwards compatible.

JacksonUtsch:
Thanks! I'll advise if I come up on any issues using this.

PR #657, opened 2026-04-17
fix(ws): do not panic in EspWebSocketClient::drop on destroy failure

### LINKED ISSUE #653: EspWebSocketClient::Drop still panics: esp_websocket_client_destroy unwrap
PR [#648 ](https://github.com/fuk-yu/esp32-zapper-control/issues/648)fixed the esp_websocket_client_close unwrap in Drop, but esp_websocket_client_destroy still calls .unwrap():

```
  impl Drop for EspWebSocketClient<'_> {
      fn drop(&mut self) {
          if let Err(e) = esp!(unsafe { esp_websocket_client_close(self.handle, self.timeout) }) {
              log::warn!("WebSocket close failed during drop: {e:?}");
          }
          esp!(unsafe { esp_websocket_client_destroy(self.handle) }).unwrap(); // can still panic
      }
  }
```

  Why this is a problem

  1. Drop must not panic. On ESP-IDF targets using panic = "abort", any panic in Drop aborts the entire application — there is no unwinding to catch it. A
  Drop impl should be infallible.
  2. The new_raw construction order triggers this. The client struct is constructed before esp_websocket_client_start() is called. When start() fails (e.g.,
  server unreachable), the ? operator returns Err, dropping the already-constructed client. Drop then runs on a client whose transport was never established.
  3. esp_websocket_client_destroy can return errors. While it typically returns ESP_OK, the ESP-IDF documentation does not guarantee this. Future ESP-IDF
  versions or edge cases (e.g., double-free, corrupted handle after a failed start) could cause it to fail.

  Suggested fix

  Apply the same pattern as close — or simpler, just ignore errors from both cleanup calls:

```
  impl Drop for EspWebSocketClient<'_> {
      fn drop(&mut self) {
          let _ = esp!(unsafe { esp_websocket_client_close(self.handle, self.timeout) });
          let _ = esp!(unsafe { esp_websocket_client_destroy(self.handle) });
      }
  }
```

  Reproduction

  1. Configure EspWebSocketClient with disable_auto_reconnect: true pointing at a non-existent server
  2. Call EspWebSocketClient::new() — the TCP connection fails asynchronously
  3. The client is dropped (either from new_raw error path or after detecting the failure)
  4. Result: abort() was called — application crashes

  Environment

  - ESP32-P4, esp-idf-svc 0.52.1, ESP-IDF v5.4.3 / v5.5.4
  - panic = "abort" (default for ESP-IDF targets)

SAY-5:
### Problem

#648 fixed the `unwrap()` on `esp_websocket_client_close` inside
`EspWebSocketClient::drop`, but left the matching call to
`esp_websocket_client_destroy` as:

```rust
impl Drop for EspWebSocketClient<'_> {
    fn drop(&mut self) {
        if let Err(e) = esp!(unsafe { esp_websocket_client_close(self.handle, self.timeout) }) {
            log::warn!("WebSocket close failed during drop: {e:?}");
        }

        esp!(unsafe { esp_websocket_client_destroy(self.handle) }).unwrap();  // can still panic
    }
}
```

A `Drop` impl must not panic. On ESP-IDF targets built with
`panic = "abort"` any panic in `Drop` aborts the whole application
without unwinding, so this is worse than a missed error.

The failure is reachable in practice: `new_raw` constructs the client
before calling `esp_websocket_client_start()`, so a failed start returns
`Err` via `?` and drops a client whose transport was never established,
which is exactly the scenario where `destroy` can fail. Reported in
#653.

### Fix

Apply the same treatment already used for `close`: log the error and
swallow it. Same warning style so log noise stays consistent.

```diff
-     esp!(unsafe { esp_websocket_client_destroy(self.handle) }).unwrap();
+     if let Err(e) = esp!(unsafe { esp_websocket_client_destroy(self.handle) }) {
+         log::warn!("WebSocket destroy failed during drop: {e:?}");
+     }
```

Fixes #653

### DIFF
--- a/src/ws/client.rs
+++ b/src/ws/client.rs
@@ -624,7 +624,13 @@ impl Drop for EspWebSocketClient<'_> {
             log::warn!("WebSocket close failed during drop: {e:?}");
         }
 
-        esp!(unsafe { esp_websocket_client_destroy(self.handle) }).unwrap();
+        // A Drop impl must not panic, especially on panic=abort targets where
+        // panicking here aborts the whole application. This matches the
+        // matching esp_websocket_client_close treatment above and extends the
+        // fix from #648 to the destroy call. See #653.
+        if let Err(e) = esp!(unsafe { esp_websocket_client_destroy(self.handle) }) {
+            log::warn!("WebSocket destroy failed during drop: {e:?}");
+        }
 
         // timeout and callback dropped automatically
     }


SAY-5:
Is this still worth taking? It only stops the panic when esp_websocket_client_destroy fails inside drop, which currently aborts instead of just logging.

ivmarkov:
> Is this still worth taking? It only stops the panic when esp_websocket_client_destroy fails inside drop, which currently aborts instead of just logging.

It is, but a variation of it.

Would you look at this comment here: https://github.com/esp-rs/esp-idf-svc/issues/653#issuecomment-4149714505

It would be great if we are able to fix the root cause (the ctor fn ending up in an inconsistent state) rather than the symptom (the dtor panicking as a result).

ivmarkov:
> > Is this still worth taking? It only stops the panic when esp_websocket_client_destroy fails inside drop, which currently aborts instead of just logging.
> 
> It is, but a variation of it.
> 
> Would you look at this comment here: [#653 (comment)](https://github.com/esp-rs/esp-idf-svc/issues/653#issuecomment-4149714505)
> 
> It would be great if we are able to fix the root cause (the ctor fn ending up in an inconsistent state) rather than the symptom (the dtor panicking as a result).

What I asked for is now in #680, thanks for your work!

PR #665, opened 2026-06-25
Thread callback fixes

jakewins:
### Submission Checklist 📝
- [NA] I have updated existing examples or added new ones (if applicable).
- [X] I have used `cargo fmt` command to ensure that all changed code is formatted correctly.
- [X] I have used `cargo clippy` command to ensure that all changed code passes latest Clippy nightly lints.
- [X] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-idf-svc/blob/main/esp-idf-svc/CHANGELOG.md) in the **_proper_** section.

#### Description

Trampolines here expect a pointer to ThreadDriverInner but we were giving them pointers to callbacks, causing crashes. This sets the callbacks on ThreadDriverInner and passes that pointer instead

#### Testing

When provisioning my device with Home Assistant and it's iOS app, it enables this "active scan" thing during provisioning that no other hub I've tried does, and that in turn crashes for me without this patch. 

The ipv6 callbacks have the same problem, with this patch I'm able to set callbacks to debug low-level traffic.


ivmarkov:
Some of the targets in the CI failed because recent nightly invented a few more clippy warnings that happen to fail on the current codebase - i.e. not related to this PR.

Thanks for the fixes - merging.

PR #667, opened 2026-07-13
Non-blocking TLS server handshake helpers, gating fix, and example

sentinelt:
### Submission Checklist 📝
- [x] My changes were added to the [`CHANGELOG.md`](https://github.com/esp-rs/esp-idf-svc/blob/master/CHANGELOG.md) in the **_proper_** section.

#### Description

Three related changes for ESP-TLS server support:

1. **Gating fix (IDF ≥ 5.3)**  
   ESP-IDF removed `CONFIG_ESP_TLS_SERVER` starting with v5.3, so the Rust `esp_idf_esp_tls_server` cfg is never set on current IDF and all server-side TLS APIs were silently compiled out. Gating is now:
   - `esp_idf_esp_tls_server` on IDF ≤ 5.2 (optional Kconfig, default off)
   - `esp_idf_version_at_least_5_3_0` + `esp_idf_esp_tls_using_mbedtls` on IDF ≥ 5.3

2. **Non-blocking server handshake helpers**  
   Callers can progress handshakes on a non-blocking socket without stalling their event loop:
   - `ServerHandshakeStatus` (`Complete` / `WantRead` / `WantWrite`)
   - `EspTls::negotiate_server_init` — `esp_tls_server_session_init`
   - `EspTls::negotiate_server_continue` — `esp_tls_server_session_continue_async`
   - `ServerConfig::tls_handshake_timeout_ms` — overall handshake timeout (`0` = ESP-TLS default of 10 s), applied to both the blocking `negotiate_server` path and the async helpers

   The blocking `negotiate_server` API is unchanged aside from honoring the new timeout field and docs pointing at the async path.

   **Breaking:** `ServerConfig` gains `tls_handshake_timeout_ms`; struct literals must set it (or use `Default`).

3. **Example**  
   `examples/tls_server.rs` — multi-connection non-blocking TLS echo server driven by `select()` over init/continue readiness.

#### Testing
- `cargo fmt --all -- --check`
- Built and flashed `examples/tls_server` for **ESP32-S3** (`xtensa-esp32s3-espidf`, ESP-IDF v5.5.3)
- Device joined Wi-Fi, listened on port 8443
- Host `openssl s_client -connect <esp-ip>:8443` completed the handshake; payload was echoed (`hello-from-host`)
- Device logs showed `Accepted` → `handshake complete` → `peer closed`


ivmarkov:
@sentinelt Can you sketch here as to how folks are supposed to use the new APIs?
They seem a tad too low level so I'm struggling to immediately figure out how one could use them.

sentinelt:
> I'm struggling to immediately figure out how one could use them.

The existing code has used `negotiate_server` which blocks up to a specified timeout - not suitable for production code, because not well behaving clients can stall the server. This commit introduces `negotiate_server_init` and `negotiate_server_continue` which allow to perform TLS negotiation without blocking.

> Can you sketch here as to how folks are supposed to use the new APIs? 

There was no example for TLS based server, so I added one which uses `negotiate_server_init` and `negotiate_server_continue`.

>  They seem a tad too low level

This is indeed low level. Maybe in the future there should be some abstraction designed for TLS servers in esp-idf-svc, WDYT?


Also while writing the example I found the issue with CONFIG_ESP_TLS_SERVER being removed in newer ESP-IDF releases - I added the fix for that as a separate commit.

src/tls.rs
```diff
@@ -389,7 +389,25 @@ mod esptls {
...
     }
 
-    #[cfg(esp_idf_esp_tls_server)]
+    /// Progress of a non-blocking server-side TLS handshake.
+    ///
+    /// Available on ESP-IDF ≥ 5.5.1 with mbedTLS (`esp_tls_server_session_init` /
+    /// `esp_tls_server_session_continue_async`).
+    #[cfg(all(esp_idf_version_at_least_5_5_1, esp_idf_esp_tls_using_mbedtls,))]
+    #[derive(Debug, Clone, Copy, PartialEq, Eq)]
+    pub enum ServerHandshakeStatus {
```
ivmarkov, line 398: This introduces asymmetry with `EspTls::negotiate` where `WantRead` and `WantWrite` are returned as `EspError`s.

Consistency rules out everything else so I suggest we retire this enum.

ivmarkov (commented):
* See the request to retire `ServerHandshakeStatus`
* I also have concerns with the API being too low level. No prob with that BUT we need a higher-level **true async** API. I'll elaborate on that in a separate comment

ivmarkov:
@sentinelt On the API being too low level.

The TL;DR is that nobody in Rust is using anymore a low-level polling with a `select` explicit loop as per your example (C style).
Rust does have `async` and we should be tapping into that because with `async`, the state machines are generated by the compiler, and as a result your code looks "sequential", just as if you were using regular blocking IO.

**What should be done:**

The problem is that you (or your LLM) have not touched `EspAsyncTls`, and that guy *IS* the "high-level" true async API.

What you need to do:
* Implement a `pub async fn EspAsyncTls::negotiate_server` API, similar in spirit to the existing `pub async fn EspAsyncTls::negotiate`. 
* This new API should internally use the same new pair `esp_tls_server_session_init`/`esp_tls_server_session_continue_async` of ESP-IDF functions that your "low level" `EspTls::negotiate_init`/`EspTls::negotiate_continue` uses, except that the "high level" API should call into `EspAsyncTls::wait` just like `negotiate` does.

Once you have the above, you can re-work your example to demonstrate the async approach similar to the `tls_async` example.

I also suggest you rename then your example to `tls_server_async`. Why? Because naming it `tls_server` hints that it is using regular blocking IO, which is not the case.

ivmarkov:
See #674

PR #672, opened 2026-07-17
Start of NimBLE wrapper

jakewins:
### Submission Checklist 📝

(Have not added CHANGELOG entry yet, happy to do that if we can get the PR to a point where there's agreement this PR is good enough to be merged)

#### Description

An initial stab of a *partial* safe wrapper for the light-weight NimBLE stack, as discussed with @ivmarkov in the rs-matter Matrix chat. This adds enough for a working GATT Server; I've confirmed the example runs and works on an ESP32-C5. 

_Disclosure_: 

(1) I'm very much not an expert in BLE or NimBLE; my use case here is for the provisioning portion of Matter. This is prep-work for upstreaming a vendored patch I'm carrying for https://github.com/sysgrok/esp-idf-matter to use NimBLE there instead of Bluedroid, in order to make rs-matter work on the ESP32-C5. 

(2) I've used Claude code while writing this changeset, though the design and structure of the patch is my own and the code has been *thoroughly* manually reviewed in several iterations and, as noted above, has been tested to work on real hardware. 

Super open to feedback here in any way to make the patch palatable for the maintainer team. I've tried to strike a balance between matching the underlying API structure of NimBLE (which is quite different form Bluedroids), while aligning with the existing shape of the repo APIs.

#### Commit message

ESP-IDF ships two bluetooth stacks, one based on Bluedroid and one based on NimBLE. Bluedroid provides a complete "real" Bluetooth stack, while NimBLE provides only BLE but does it with lower resource use.

This adds an *incomplete* and *early* safe wrapper for the NimBLE stack, just enough to implement a bare-bones BLE GATT server.

The overall structure tries to find a good balance between providing a safe API on top of normal NimBLE routines, without completely diverging from the structure already established by the Bluedroid wrapper.

The two APIs are very different, so the new wrapper is built side-by-side, giving this repo `bt.rs` for full bluetooth via Bluedroid and `ble.rs` for light-weight BLE via NimBLE.

Structure:

```
src/
  ble.rs # top-level entrypoint
  ble/gap.rs # GAP stuff, wraps ble_gap_XXX and ble_svc_gap_XXX routines
  ble/gatt.rs # GATT stuff shared between clients (not implemented) and servers
  ble/gatt/gatts.rs # wraps ble_gatts_XXX
  ble/mbus.rs # wrappers for NimBLEs buffer pool
```

For GATT servers, NimBLE takes a tree structure of attributes and callback pointers. The proposed API here mirrors that approach, expecting users to allocate and provide a reference to an equivalent Rust structure - note there's quite a bit of dragons there as this ends up being a self-referential structure we give to the NimBLE stack.. I think the way I've structured it now is safe from moves while the ble stack is running but could definintely use some critical eyes.

#### Testing

I built the example, flashed it to an ESP32-C5 and then verified that my laptop's BLE stack could discover and interact with it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ remote_component = { name = "espressif/lan87xx", version = "1.*" }
 ### Added
 - Compatibility with ESP-IDF V6.0, and some pre-release 6.0.x.
 - Added support for the Generic Ethernet PHY driver: particularly useful on ESP-IDF 6.0+ as it is built-in.
+- Added early support for the NimBLE low-resource-use BLE stack, currently only GAP and GATT Server support. See examples/ble_gatt_server.rs
 
 ## [0.52.1] - 2026-03-10
 
--- a/examples/ble_gatt_server.rs
+++ b/examples/ble_gatt_server.rs
@@ -94,7 +94,10 @@ mod example {
         // on the host task. We stash the indicate handle so the loop below can push to
         // it; matching on the UUID is how we tell our characteristics apart.
         setup.on_gatts_register(|event| {
-            if let BleGattRegister::Characteristic { uuid, val_handle, .. } = event {
+            if let BleGattRegister::Characteristic {
+                uuid, val_handle, ..
+            } = event
+            {
                 if uuid == BleUuid::uuid128(IND_CHARACTERISTIC_UUID) {
                     IND_VAL_HANDLE.store(val_handle, Ordering::Relaxed);
                 }
@@ -170,7 +173,7 @@ mod example {
         }
     }
 
-    /// Configure and start a connectable legacy advertisement 
+    /// Configure and start a connectable legacy advertisement
     /// n.b. NimBLE exposes a mutually-exclusive "extended" advertisement API as well
     /// if you set the right build flags
     fn start_advertising() -> Result<(), BleError> {
@@ -187,8 +190,8 @@ mod example {
         gap::adv_set_fields(&fields)?;
 
         let params = BleAdvParams {
-            conn_mode: 2, // BLE_GAP_CONN_MODE_UND
-            disc_mode: 2, // BLE_GAP_DISC_MODE_GEN
+            conn_mode: 2,   // BLE_GAP_CONN_MODE_UND
+            disc_mode: 2,   // BLE_GAP_DISC_MODE_GEN
             itvl_min: 0x30, // 30 ms, in 0.625 ms units
             itvl_max: 0x60, // 60 ms
             ..Default::default()
--- a/examples/ble_gatt_server.rs
+++ b/examples/ble_gatt_server.rs
@@ -1,0 +1,192 @@
+//! Example of a BLE GATT server using the ESP IDF NimBLE bindings.
+//!
+//! Requires a NimBLE-enabled build.
+
+#![allow(unknown_lints)]
+#![allow(unexpected_cfgs)]
+
+#[cfg(all(not(any(esp32s2, esp32p4)), esp_idf_bt_nimble_enabled))]
+fn main() -> anyhow::Result<()> {
+    example::main()
+}
+
+#[cfg(not(all(not(any(esp32s2, esp32p4)), esp_idf_bt_nimble_enabled)))]
+fn main() -> anyhow::Result<()> {
+    panic!("This example requires a NimBLE-enabled build (CONFIG_BT_NIMBLE_ENABLED=y) on a chip with a BLE radio");
+}
+
+#[cfg(all(not(any(esp32s2, esp32p4)), esp_idf_bt_nimble_enabled))]
+mod example {
+    use core::sync::atomic::{AtomicU16, Ordering};
+    use std::sync::Mutex;
+
+    use esp_idf_svc::ble::gap::{self, BleAdvFields, BleExtAdvParams, BleGapEvent};
+    use esp_idf_svc::ble::gatt::gatts::{
+        self, BleGattAccess, BleGattCharacteristic, BleGattService, BleGattServices, ConnectionId,
+        GattsSetup,
+    };
+    use esp_idf_svc::ble::gatt::BleGattCharFlag;
+    use esp_idf_svc::ble::{ensure_addr, BleError, BleSetup, BleUuid};
+    use esp_idf_svc::hal::delay::FreeRtos;
+    use esp_idf_svc::hal::peripherals::Peripherals;
+    use esp_idf_svc::log::EspLogger;
+
+    use enumset::enum_set;
+    use log::{info, warn};
+
+    const DEVICE_NAME: &str = "esp-nimble";
+    const ADV_INSTANCE: u8 = 0;
+
+    // Our service UUID
+    pub const SERVICE_UUID: u128 = 0xad91b201734740479e173bed82d75f9d;
+
+    /// Our "recv" characteristic - i.e. where clients can send data.
+    pub const RECV_CHARACTERISTIC_UUID: u128 = 0xb6fccb5087be44f3ae22f85485ea42c4;
+    /// Our "indicate" characteristic - i.e. where clients can receive data if they subscribe to it
+    pub const IND_CHARACTERISTIC_UUID: u128 = 0x503de214868246c4828fd59144da41be;
+
+    // Server state. NimBLE fills the value handles during registration.
+    static SUBSCRIBERS: Mutex<Vec<ConnectionId>> = Mutex::new(Vec::new());
+    static RECV_VAL_HANDLE: AtomicU16 = AtomicU16::new(0);
+    static IND_VAL_HANDLE: AtomicU16 = AtomicU16::new(0);
+
+    pub fn main() -> anyhow::Result<()> {
+        esp_idf_svc::sys::link_patches();
+        EspLogger::initialize_default();
+
+        let peripherals = Peripherals::take()?;
+
+        let services = BleGattServices::new(vec![BleGattService::new(
+            true,
+            BleUuid::uuid128(SERVICE_UUID),
+            vec![
+                // "recv": clients write here; we just log what arrives.
+                BleGattCharacteristic::new(
+                    BleUuid::uuid128(RECV_CHARACTERISTIC_UUID),
+                    enum_set!(BleGattCharFlag::Write),
+                    &RECV_VAL_HANDLE,
+                    |access| {
+                        if let BleGattAccess::Write { data, .. } = access {
+                            let mut buf = [0u8; 200];
+                            match data.read(&mut buf) {
+                                Ok(n) => info!("recv {n} bytes: {:?}", &buf[..n]),
+                                Err(e) => warn!("recv read failed: {e}"),
+                            }
+                        }
+                        0
+                    },
+                ),
+                // "indicate": clients subscribe and get the counter pushed from the loop
+                // below. NimBLE adds the CCCD (0x2902) automatically for this flag, so
+                // there is no descriptor to declare and no read/write to service here.
+                BleGattCharacteristic::new(
+                    BleUuid::uuid128(IND_CHARACTERISTIC_UUID),
+                    enum_set!(BleGattCharFlag::Indicate),
+                    &IND_VAL_HANDLE,
+                    |_access| 0,
+                ),
+            ],
+        )]);
+
+        let mut setup = BleSetup::new(peripherals.modem)?;
+
+        GattsSetup::new(&mut setup).add_services(&services)?;
+
+        // We wait until the stack is "in sync" before we can start using it. Note this
+        // closure needs to handle being called multiple times in case the stack resets.
+        setup.on_sync(|| match start_advertising() {
+            Ok(()) => info!("advertising as {DEVICE_NAME:?}"),
+            Err(e) => warn!("failed to start advertising: {e}"),
+        });
+
+        setup.on_gap_event(|event| {
+            match event {
+                BleGapEvent::Connect {
+                    conn_handle,
+                    status,
+                } => info!("connected (handle {conn_handle}): {status:?}"),
+                BleGapEvent::Disconnect {
+                    conn_handle,
+                    reason,
+                } => {
+                    info!("disconnected ({reason}); restarting advertising");
+                    SUBSCRIBERS.lock().unwrap().retain(|&c| c != conn_handle);
+                    if let Err(e) = start_advertising() {
+                        warn!("failed to restart advertising: {e}");
+                    }
+                }
+                BleGapEvent::Subscribe {
+                    conn_handle,
+                    attr_handle,
+                    cur_indicate,
+                    ..
+                } => {
+                    if attr_handle == IND_VAL_HANDLE.load(Ordering::Relaxed) {
+                        let mut subs = SUBSCRIBERS.lock().unwrap();
+                        subs.retain(|&c| c != conn_handle);
+                        if cur_indicate {
+                            let _ = subs.push(conn_handle);
+                        }
+                    }
+                }
+                _ => {}
+            }
+
+            0
+        });
+
+        let _driver = setup.start()?;
+        info!("NimBLE host started");
+
+        let mut counter: u16 = 0;
+        loop {
+            FreeRtos::delay_ms(1000);
+
+            // The handle stays 0 until NimBLE has registered the characteristic.
+            let ind_handle = IND_VAL_HANDLE.load(Ordering::Relaxed);
+            if ind_handle == 0 {
+                continue;
+            }
+
+            counter = counter.wrapping_add(1);
+
+            // Copy the subscriber list out so the lock isn't held across `indicate`.
+            let subs = SUBSCRIBERS.lock().unwrap().clone();
+            for conn in subs {
+                if let Err(e) = gatts::indicate(conn, ind_handle, &counter.to_le_bytes()) {
+                    warn!("indicate to {conn} failed: {e}");
+                }
+            }
+        }
+    }
+
+    /// Configure and start a connectable legacy advertisement.
+    fn start_advertising() -> Result<(), BleError> {
+        ensure_addr(false)?;
+        gap::svc_set_device_name(DEVICE_NAME)?;
+
+        let params = BleExtAdvParams {
+            connectable: true,
+            scannable: true,
+            legacy_pdu: true,
+            itvl_min: 0x30,   // 30 ms, in 0.625 ms units
+            itvl_max: 0x60,   // 60 ms
+            own_addr_type: 0, // BLE_OWN_ADDR_PUBLIC
+            primary_phy: 1,   // BLE_HCI_LE_PHY_1M
+            secondary_phy: 1, // BLE_HCI_LE_PHY_1M
+            tx_power: 127,    // no preference
+            ..Default::default()
+        };
+
+        gap::ext_adv_configure(ADV_INSTANCE, &params)?;
+
+        let fields = BleAdvFields {
+            flags: 0x06, // LE General Discoverable, BR/EDR unsupported
+            name: Some(DEVICE_NAME),
+            ..Default::default()
+        };
+        gap::ext_adv_set_fields(ADV_INSTANCE, &fields)?;
+
+        gap::ext_adv_start(ADV_INSTANCE)
+    }
+}
--- a/examples/ble_gatt_server.rs
+++ b/examples/ble_gatt_server.rs
@@ -22,8 +22,8 @@ mod example {
 
     use esp_idf_svc::ble::gap::{self, BleAdvFields, BleGapEvent};
     use esp_idf_svc::ble::gatt::gatts::{
-        self, BleGattAccess, BleGattCharacteristic, BleGattService, BleGattServices, ConnectionId,
-        GattsSetup,
+        self, BleGattAccess, BleGattCharacteristic, BleGattRegister, BleGattService,
+        BleGattServices, ConnectionId, GattsSetup,
     };
     use esp_idf_svc::ble::gatt::BleGattCharFlag;
     use esp_idf_svc::ble::{ensure_addr, BleError, BleSetup, BleUuid};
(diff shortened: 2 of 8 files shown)


ivmarkov:
@jakewins Nice!

I did a manual (human) review already.

... but rather than putting my comments in-line (times change!) I informed Opus about my code review feedback and asked it to do a fair assessment of my feedback and come up with its **own** one.

Please find its feedback below. Perhaps, task your Opus to look at it and discuss with it, as I did?

(BTW: some of its feedback is based on the `nimble` branch in esp-idf-matter which features support for NimBLE with the raw/unsafe APIs. Basically, it checked whether it can port what we currently have in the `nimble` branch to the safe APIs you suggest in this PR.)

ivmarkov:
The review feedback from Opus
========

## Review — NimBLE safe wrappers

Thanks for this — it's a genuinely useful addition, and the instinct to model it on the existing
Bluedroid module is the right one: `BleCallback`/`BleSingleton` mirror `BtSingleton`/`BtDriver`
closely (`subscribe` / `subscribe_nonstatic` / `unsubscribe` / `call`, `AtomicBool` take/release,
`Drop`-based teardown, `esp!()`), which is exactly how a second BLE host should look next to the
first.

I reviewed it against three lenses — **lean/low-allocation**, **consistent with the rest of
esp-idf-svc**, and **soundness** — and then stress-tested it by porting a real workload onto it
(a Matter BTP GATT peripheral: 805 lines, 5 `unsafe extern "C"` callbacks, 14 `unsafe` blocks,
17 raw `sys` calls). That port is where the two blockers below came from; everything else is
either a soundness nit or an API-shape suggestion.

Also worth saying up front, because it's the load-bearing part of the design: **the
self-referential `BleGattServices` construction is sound.** The `arg` / `characteristics`
pointers target `Vec`/`Box` *heap* buffers that survive moves of the struct handle, and
`add_services(&'ble ...)` ties the services' lifetime to the driver through the borrow checker so
they can't be dropped or moved while NimBLE holds the pointers. The "dragons" comment is accurate
and the invariant it describes actually holds.

---

# 🔴 Blockers

## B1. Advertising is extended-adv only — won't link in a default NimBLE build

`src/ble/gap.rs` exposes only `ble_gap_ext_adv_*`. Those symbols are compiled out unless
`CONFIG_BT_NIMBLE_EXT_ADV=y`, which is **not** the default and not what a Thread/Matter device
uses. Checked against a stock NimBLE build:

```
# CONFIG_BT_NIMBLE_EXT_ADV is not set
ble_gap_ext_adv_configure   -- not bound --
ble_gap_ext_adv_start       -- not bound --
ble_gap_ext_adv_set_data    -- not bound --
ble_gap_adv_start           FOUND      <- legacy
ble_gap_adv_set_data        FOUND      <- legacy
```

So `gap.rs` fails to link for anyone who hasn't opted into ext-adv, and enabling ext-adv costs
RAM (per-instance advertising state) for no benefit when all you need is a legacy connectable
advertisement.

**Ask:** add the legacy path (`ble_gap_adv_set_data` / `ble_gap_adv_set_fields` /
`ble_gap_adv_start` / `ble_gap_adv_stop`) as the baseline, and `#[cfg(esp_idf_bt_nimble_ext_adv)]`
the extended API on top.

## B2. No way to configure `ble_hs_cfg.sm_*` (SMP/security)

Surfaced by the port. A Matter peripheral must disable BLE-level pairing outright, because Matter
carries its own crypto over BTP:

```rust
(*cfg).set_sm_bonding(0);
(*cfg).set_sm_mitm(0);
(*cfg).set_sm_sc(0);
```

Nothing in the PR exposes this, and it is **pre-start** configuration, so there's no way to reach
it after `start()` either. As it stands a Matter peripheral cannot be expressed without dropping
back to raw `sys`.

**Ask:** something like

```rust
pub struct BleSecurity {
    pub bonding: bool,
    pub mitm: bool,
    pub secure_connections: bool,
    pub io_cap: BleIoCap,
}
impl Default for BleSecurity { /* all off, NoInputNoOutput */ }

/// Must be called before the host task is started.
pub fn set_security(&self, sec: &BleSecurity) -> Result<(), EspError>;
```

---

# 🟡 Soundness

## S1. `val_handle: &AtomicU16` is a one-sided atomic

`BleGattCharacteristic::val_handle` is an `&AtomicU16`, and `BleGattServices::new` hands NimBLE
`AtomicU16::as_ptr()`. NimBLE then writes that handle with a **plain, non-atomic** store, on the
host task, during `ble_gatts_start` at sync — concurrently with the app's `.load()` (see the
example's polling loop). That's a mixed atomic/non-atomic access to one location: formally a data
race, even though it's benign in practice (naturally-aligned `u16`, hardware-atomic on RV32 and
Xtensa, and the `handle == 0` sentinel covers "not yet written").

The `AtomicU16` does earn its keep on the read side — it stops the compiler treating the static as
constant-zero — but the atomicity is one-sided, so the type advertises a guarantee the writer
doesn't honour. For a crate whose value proposition is *safe* wrappers, that's worth closing.

**Suggestion:** leave `ble_gatt_chr_def.val_handle` null and capture handles from NimBLE's own
registration callback (`ble_hs_cfg.gatts_register_cb`), which runs on the host task and is *your*
Rust code — store with `Release`, read with `Acquire`. Then every write is Rust-side and ordered:

```rust
pub struct ValHandle(/* AtomicU16 */);
impl ValHandle {
    pub const fn new() -> Self;
    /// `None` until the GATT DB has been registered.
    pub fn get(&self) -> Option<Handle>;
}
```

The `Option` also removes the magic `0` sentinel the example currently has to test for.

## S2. mbuf ownership on error paths

`gatts::indicate` and `gap::ext_adv_set_data` allocate via `mbuf_from_slice` and hand the mbuf to
`ble_gatts_indicate_custom` / `ble_gap_ext_adv_set_data`. Today those free the mbuf on every path
(`os_mbuf_free_chain`), so this is correct — but it's an unwritten contract that would leak if it
ever changed. Worth stating explicitly in a comment next to each call, since the code has no
cleanup of its own.

## S3. `unwrap()` in `Drop`

`esp!(nimble_port_deinit()).unwrap()` in `BleDriver::drop` aborts if teardown fails, and can
double-panic during unwind. `BtDriver` does the same thing, so it's not a regression — flagging
for consistency of judgement rather than as a request.

## S4. `id_copy_addr` swallows its return code

`let _ = unsafe { ble_hs_id_copy_addr(...) }` returns an all-zero `BleAddr` if no address of that
type is configured, indistinguishable from a real one. Should be
`fn id_addr(kind) -> Result<BleAddr, BleError>`.

---

# 🟢 API shape

## A1. The `*Setup` idiom is bespoke — but don't remove it naively

`BleSetup` → `GattsSetup` → `start(self) -> BleDriver` is a consume-`self` two-phase handoff that
doesn't appear anywhere else in the crate. (The `.start()` on `WifiDriver` / `EthDriver` /
`ThreadDriver` is a *reversible* `&mut self` runtime control, not a builder transition.)

That said, it's not arbitrary — it solves two real problems at once: NimBLE requires services to
be registered *before* the host task starts (unlike Bluedroid's dynamic `register_app`), and
keeping `GattsSetup` separate from `BleSetup` means a **GATT Central** simply never touches it.
Any replacement has to preserve both.

In particular, do **not** fold services into the driver constructor
(`BleDriver::new(modem, &services)`): that conflates the host with the peripheral role and leaves
a Central with nothing sensible to pass.

**Suggested shape** — role-agnostic driver, server as a separate handle, `new()` + `start()`
matching `WifiDriver`:

```rust
let mut driver = BleDriver::new(modem)?;   // nimble_port_init + trampolines; task NOT started
Gatts::new(&driver).add_services(&svcs)?;  // PERIPHERAL ONLY - NimBLE's pre-start window
driver.set_security(&BleSecurity::default())?;
driver.on_sync(|| ...);                    // latched: subscribing after sync fires immediately
driver.start()?;                           // host task up
// Central: identical, minus the Gatts line
```

This drops `BleSetup`/`GattsSetup`, keeps the driver role-agnostic, leaves services in NimBLE's
natural pre-start window (no `ble_gatts_reset` dance), and reuses the crate's own `new()`→`start()`
shape.

**Honest cost:** "services before start" moves from a compile-time guarantee (the Setup type-state)
to a runtime `BLE_HS_EBUSY` from `add_services`. That's consistent with how `WifiDriver` reports
out-of-order `start`/`stop`, so I think it's the right trade — but it *is* a reduction in static
safety, and if you'd rather keep the guarantee, keeping Setup is a defensible call. The one thing
I'd push on either way is A2.

## A2. Free functions on global state → methods on handles

`gatts::indicate`, `gap::ext_adv_*` and `gap::svc_set_device_name` are free functions that mutate
global NimBLE state, so they're callable with no live host and fail at runtime rather than compile
time. The crate's idiom is a handle borrowing the driver — `EspBleGap` / `EspGatts` with
`T: Borrow<BtDriver<'d, M>>` plus `subscribe` / `subscribe_nonstatic`:

```rust
pub struct Gap  <'d, T: Borrow<BleDriver<'d>>> { .. }
pub struct Gatts<'d, T: Borrow<BleDriver<'d>>> { .. }

gap.adv_start(&params)?;                 // instead of gap::ext_adv_start(inst)
gatts.indicate(conn, handle, data)?;     // instead of gatts::indicate(..)
```

That generic also lets a handle take `&driver`, `driver`, or `Arc<driver>`, same as today's
Bluedroid handles. It composes with A1 as one refactor, and leaves room for a `Gattc` handle later
without touching `BleDriver`.

## A3. Two error types

`EspError` for setup (`BleSetup::new`, `start`) and `BleError` for operations is a defensible
split (ESP-IDF-level vs NimBLE `rc`), but it's surprising on first read. A `From<BleError> for
EspError` or a short doc note on the boundary would help.

---

# 🔵 Footprint

Both of these are observations rather than change requests — the criterion I care about is
*on-the-fly* allocation, and the PR is fine there.

**F1.** The whole GATT tree is heap: `Vec<BleGattService>`, `Vec<Box<[ble_gatt_chr_def]>>`,
`Box<[ble_gatt_svc_def]>`, plus a `Box<dyn FnMut>` per characteristic. It's all setup-time, so
nothing allocates per packet. But it does foreclose a `static`/`const` table option, which is the
leaner path on the RAM-constrained targets this is most useful on (an H2 running Thread + Matter +
BLE has roughly 22 KB of free internal heap). Worth leaving room for a no-alloc constructor later.

**F2.** `BleCallback` = `Arc<UnsafeCell<Box<dyn FnMut>>>` with an `Arc` clone per dispatch — two
allocations per callback and a refcount bump per event. It mirrors `BtSingleton` exactly, so this
is a consistency *win*; GAP events are rare enough that it doesn't matter. Noted only for
completeness.

---

# ✅ Validation: porting a real workload

To check the API carries something non-trivial, I mapped a Matter BTP GATT peripheral onto it
(3 characteristics, indications, connect/disconnect/subscribe/MTU/notify-tx handling, advertising
restart on disconnect):

- **All 17 raw `sys` calls** it makes have a home in this API — no escape hatch needed, apart from
  B1 and B2.
- **~380 lines of unsafe FFI collapse to ~90 lines of safe closures**; all 5 `unsafe extern "C"`
  callbacks and essentially all 14 `unsafe` blocks disappear, since `BleGapEvent` / `BleGattAccess`
  already decode the unions and `Mbuf` already wraps the buffers.
- A hand-rolled `static AtomicPtr` "current context" pointer disappears entirely, because closures
  capture `&ctx` directly.

One ergonomic note, no API change needed: `on_sync` starts advertising and the GAP handler
*restarts* it on disconnect, so two closures both need the `Gap` handle — and a closure can't own
the handle it's registered on. The fix is the same event-context-struct pattern the Bluedroid
backend already uses (a small struct holding `&gap` / `&ctx`, with closures capturing `&ev` via the
`_nonstatic` variants). Worth a doc example, since it's the non-obvious part of composing this API.

---

# Checklist

**Must fix**
- [ ] **B1** — add legacy advertising; `cfg`-gate ext-adv behind `esp_idf_bt_nimble_ext_adv`
- [ ] **B2** — expose `ble_hs_cfg.sm_*` (`set_security`) as pre-start config

**Should fix**
- [ ] **S1** — `val_handle` via `gatts_register_cb` + `ValHandle::get() -> Option<Handle>`
- [ ] **A2** — free functions → methods on `Gap` / `Gatts` handles borrowing the driver
- [ ] **S4** — `id_addr` returns `Result`

**Consider**
- [ ] **A1** — `BleDriver::new()` + `start()` in place of `BleSetup`/`GattsSetup` (see the trade-off)
- [ ] **S2** — document mbuf ownership at the `indicate` / `set_data` call sites
- [ ] **A3** — `From<BleError>` or a doc note on the two error types
- [ ] **F1** — leave room for a non-allocating service-table constructor
- [ ] doc example for the shared-handle/event-context pattern

jakewins:
I addressed the blockers and several others, some remain, with the following rationale:

### A1:  BleDriver::new() + start() in place of BleSetup/GattsSetup (see the trade-off)

Thought through this one, no super strong feelings here but personal preference would be to keep the type-safe setup, I think it's helpful to clients to make it clear what functions must be called before the stack is started for them to have any effect - but happy to change to keep just BleDriver and make calling setup stuff after start() be runtime errors instead if you prefer.

### A2:  free functions → methods on Gap / Gatts handles borrowing the driver

This was actually the original design I tried and, in theory, I much prefer it. The trouble is there's this chicken-and-egg problem where those methods need to be accessible from within the on_sync callback *and* from code downstream of `BleSetup#start()` (or `BleDriver#start()` if we did that refactoring). 

The actual hurdle for most of those methods isn't that the driver is *started* it's that it's *currently in sync*. However, modelling that in the type system, like by passing handles or a guard or something into the on_sync callback has been.. super tricky, I've not found a design that felt like it's complexity paid for itself. 

Ultimately, the free functions match how NimBLE does it, and you simply get a runtime error if you call them at the wrong state of the NimBLE stack.. I think that's ok, at least as a starting point, though I agree I would prefer some sort of design that makes the temporal requirements encoded in the type system, I think that can be left as follow-on improvement work. 



---

I haven't tested the latest changes, will sync back here when I do (though it'll be tomorrow, I'm in the US over the summer and it's getting late for my jetlagged brain)

jakewins:
@ivmarkov alright tested it now this morning and confirmed:

- Example runs
- GAP works, I can find it via BLE from my laptop
- GATT reads work
- GATT writes work
- Subscribe/indicate works

So I think it passes the smoke test - let me know if you want additional changes to the changeset. Added a changelog entry as well!

Ah I forgot one thing in the prior comment as well: I would *love* to do F1 (alloc-free GATT API), but I tried a stab at it and made quite a mess. It seems with this early level of support perhaps that's better done as a follow on improvement to the API rather than another lark in this changeset.

ivmarkov:
> ### A1: BleDriver::new() + start() in place of BleSetup/GattsSetup (see the trade-off)
...snip...
> ### A2: free functions → methods on Gap / Gatts handles borrowing the driver
> This was actually the original design I tried and, in theory, I much prefer it. The trouble is there's this chicken-and-egg problem where those methods need to be accessible from within the on_sync callback _and_ from code downstream of `BleSetup#start()` (or `BleDriver#start()` if we did that refactoring).

I think (a) the chicken and egg problem is actually fixable and (b) it is however only fixable if A1 is implemented **as well and together with A2**. *If you do both A1 and A2 you are basically ending up with the Bluedroid APIs design.*

===

With that said, I agree - let's merge as-is, and potentially address these after that.

===

I also asked Fable to confirm/deny the above - just in case. Leaving here its reply in case we decide to vibe-code A1+A2 later, in a follow up PR:

Re A2 — I don't think the chicken-and-egg is inherent; I think it's an artifact of the
`start(self) -> BleDriver` handoff, and the Bluedroid module is the existence proof.

**The same problem is already solved twice in this crate**, once per lifetime flavor:

1. `examples/bt_gatt_server.rs` (Bluedroid): `Arc<EspBleGap>` / `Arc<EspGatts>` are stored in an
   `ExampleServer`, and the closure passed to `server.gap.subscribe(..)` captures a clone of the
   server — i.e. the callback registered *on* `gap` calls methods *on* `gap` when events arrive.
   The example even notes you can use plain references instead of `Arc`s if you're willing to
   manage lifetimes.

2. The Matter BTP peripheral over Bluedroid does exactly that borrowed variant: an event-context
   struct `GattEventContext { gap: &gap, gatts: &gatts, .. }`, with the closure registered via
   `gap.subscribe_nonstatic(|e| event_ctx.on_gap_event(e))` — a closure registered on `gap`
   capturing `&gap` (through the context). No cycle, because `subscribe_nonstatic` takes `&self`,
   so all borrows are shared.

**Why it genuinely didn't work when you tried it here:** in the current design, `start(self)`
*consumes* `BleSetup`. So a `Gap` handle borrowing the setup is invalidated by the move, and a
`Gap` handle borrowing `BleDriver` can't exist yet at the time `on_sync` is registered — the
driver is only born from `start()`. Callbacks are registered pre-start, handles could only exist
post-start, and the object they'd borrow dies in between. That *is* a chicken-and-egg — but it's
created by the consume-self typestate, not by NimBLE. Bluedroid never hits it because
`BtDriver::new()` returns the one long-lived object up front and nothing moves afterwards.

Which means A2 and the `*Setup` question are really one issue: with a driver that is created once
and never moved, the standard patterns apply directly —

```rust
let driver = BleDriver::new(modem)?;              // host init, task not started
let gatts = Gatts::new(&driver)?;
gatts.add_services(&services)?;                    // pre-start window, as today

let gap = Gap::new(&driver)?;
let ev = EventCtx { gap: &gap, ctx: &my_state };   // or Arc everything for 'static

unsafe {
    gap.subscribe_nonstatic(|e| ev.on_gap_event(e));
    driver.on_sync_nonstatic(|| ev.start_advertising());
}

driver.start()?;                                   // &self + AtomicBool (your `started` flag),
                                                   // so the shared borrows above stay valid
```

(The one design consequence: `start` wants `&self` with an interior started-flag rather than
`&mut self`/`self`, since the handles hold shared borrows of the driver across it. `BtDriver`
already works entirely through `&self`, so that's idiomatic here.)

On "free functions match how NimBLE does it" — true, but `esp_ble_gatts_*` are also free C
functions on global state, and this crate wrapped them in `EspGatts` anyway. Matching the C shape
is exactly what esp-idf-svc doesn't do elsewhere.

On sync-state vs started-state — agreed, and I'd keep "synced" entirely out of the type system;
it stays a runtime `BLE_HS_*` error in every design. But that's orthogonal to what handles buy:
they encode *existence and lifetime* (you can't call into a stack that was never created or
already dropped, and unsubscription is tied to `Drop`), which the free functions don't encode at
all. We don't have to type the hard invariant to get the cheap one.

So I'd suggest doing the two together: driver-created-once + `start(&self)`, and the ops as
methods on `Gap`/`Gatts` handles — at which point the callback access problem disappears the same
way it already has in the Bluedroid module.

ivmarkov:
> @ivmarkov alright tested it now this morning and confirmed:
> 
> * Example runs
> * GAP works, I can find it via BLE from my laptop
> * GATT reads work
> * GATT writes work
> * Subscribe/indicate works
> 
> So I think it passes the smoke test - let me know if you want additional changes to the changeset. Added a changelog entry as well!
> 
> Ah I forgot one thing in the prior comment as well: I would _love_ to do F1 (alloc-free GATT API), but I tried a stab at it and made quite a mess. It seems with this early level of support perhaps that's better done as a follow on improvement to the API rather than another lark in this changeset.

I think (but this is just intuition) that F1 would work _only_ if the char RW callback we record everywhere in these structs that describe the gatt server and that we pass down to NimBLE **is always the same, regardless of the concrete char being read / written**. Which is maybe not the end of the world given that you _do_ get the char handle as a parameter in this callback, so you _can_ figure out which characteristic you are supposed to operate. Weird therefore that NimBLE even wants a __separate__ callback per characteristic - Bluedroid takes a single callback, which is in fact the events' callback, simply with extra events for char reads / writes.

Anyway, F1 is also definitely a follow-up work!

ivmarkov:
Fable again, on the sync topic, as I asked it to think harder about specifically your sync concerns:

====

On the sync-state point specifically — I think you've identified the right invariant and I'd go
further than "tricky": encoding synced-ness in the type system is unsound *in principle*, not
just complex. `ble_hs_sync_state` is non-monotonic — the reset path zeroes it
(`ble_hs.c`, reset → `sync_state = 0` → `reset_cb` → re-sync → `sync_cb` again) — so any
type-level `Synced` witness is a promise the controller can revoke while you hold it. The only
window where sync is actually stable is inside a host-task callback (the host task is
single-threaded, so a reset can't be processed mid-callback), which is presumably why
guard-into-`on_sync` designs keep almost working.

And there's a structural reason the guard approach can't be rescued: the operations must be
callable from *outside* the callbacks. A concrete consumer — the Matter BTP peripheral — calls
`indicate()` from its own async pump task and only *initiates* advertising from `on_sync`; a
sync-witness scoped to host-task callbacks cannot express that at all. So agreed: sync stays a
runtime condition, full stop.

But I'd separate the two invariants, because they have different answers:

- **"currently synced"** — dynamic, hardware-revocable, needed from arbitrary tasks → runtime
  error. This is also exactly how the crate treats Wifi "connected" / Eth "link up"; nobody
  types those either.
- **"the stack exists and is alive"** — static, program-controlled → this is what handles
  encode (plus drop-tied unsubscription), and what free functions give up.

Sync being untypeable doesn't take the second one off the table — the Bluedroid module ships it
today.

Concretely for sync, I'd do four small things and stop there:

1. Methods just surface `BleError(BLE_HS_ENOTSYNCED)` — NimBLE already gates the ops, so this is
   free.
2. Latch `on_sync`: subscribing after the host has already synced invokes the callback
   immediately. That removes the one genuine race in this area.
3. `fn synced(&self) -> bool` wrapping `ble_hs_synced()` (already in the bindings) as an
   advisory check.
4. Name `BLE_HS_ENOTSYNCED` in `BleError::name()` and document the idiom: drive advertising
   start/restart from `on_sync` + the disconnect event; everything else may transiently fail
   with `ENOTSYNCED` across a controller reset, and callers should treat that as retryable.

PR #677, opened 2026-07-31
Add Trouble BLE Support over ESP-IDF VHCI

dickermoshe:

### Submission Checklist 📝

#### Description

This crate already includes support for ESP-IDF’s Bluedroid and NimBLE BLE stacks. However, the [Trouble](https://github.com/embassy-rs/trouble) crate makes it much easier to create Bluetooth Low Energy applications. 

This PR introduces Trouble support by implementing the `bt-hci` traits. This is also how the `esp-radio` crate enables Bluetooth for devices not using `esp-idf`

In addition to the changes included here, there was also an issue with nightly Rust not working with some upstream changes in the Rust compiler. To be honest, I'm not sure what it is. To have CI pass, I've capped the nightly release to `nightly-2026-07-29`, but I'm sure this is not how you want to fix it. If you could let me know how you want to fix it, I'll do whatever you want.

#### Testing

I added a small Trouble beacon example to exercise the controller and VHCI transport together.


.github/configs/sdkconfig.defaults.trouble
```diff
@@ -0,0 +1,15 @@
...
+# Required on chips that let the application select the controller HCI
+# interface. Other chips use VHCI for controller-only mode automatically.
+CONFIG_BT_LE_HCI_INTERFACE_USE_RAM=y
+
+# Trouble reads into a 259-byte HCI buffer. An ACL payload of 251 bytes occupies
+# 256 bytes with the ACL header and H4 packet indicator.
+CONFIG_BT_LE_ACL_BUF_SIZE=251
+
+# The Trouble runner and host resources exceed ESP-IDF's default main-task stack.
+CONFIG_ESP_MAIN_TASK_STACK_SIZE=32768
```
ivmarkov, line 15: `HostResources` should not be allocated on-stack, but staticlaly in bss or on heap. It is another topic that they regressed it recently, where such an allocation was not possible anymore due to the `Controller` re-generification. Fortunately, fixed in `main`: https://github.com/embassy-rs/trouble/issues/645

.github/workflows/publish-dry-run.yml
```diff
@@ -3,7 +3,9 @@ name: PublishDryRun
 on: workflow_dispatch
 
 env:
-  rust_toolchain: nightly
+  # Temporary workaround for rust-lang/rust#158168. Remove after the fix reaches nightly:
```
ivmarkov, line 6: Revert, this script does not run every day.

.github/workflows/publish.yml
```diff
@@ -3,7 +3,9 @@ name: Publish
 on: workflow_dispatch
 
 env:
-  rust_toolchain: nightly
+  # Temporary workaround for rust-lang/rust#158168. Remove after the fix reaches nightly:
+  # https://github.com/rust-lang/rust/pull/160170
+  rust_toolchain: nightly-2026-07-29
```
ivmarkov, line 8: Ditto.

examples/trouble_beacon.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    use esp_idf_svc::bt_controller::{Ble, EspBtController, EspVhciTransport};
+    use esp_idf_svc::hal::peripherals::Peripherals;
+    use esp_idf_svc::hal::task::block_on;
+    use esp_idf_svc::nvs::EspDefaultNvsPartition;
+
+    type Controller = ExternalController<EspVhciTransport<'static>, 10>;
+
+    pub fn main() -> Result<()> {
+        esp_idf_svc::sys::link_patches();
```
ivmarkov, line 52: This is no longer necessary, as it is called automatically by the startup code in `esp-idf-sys`. The other two explicit `link` calls should also not be necessary, unless you hit an actual linking issue.

src/bt_controller/trouble.rs
```diff
@@ -0,0 +1,335 @@
+#[cfg(not(esp_idf_bt_controller_only))]
```
ivmarkov, line 1: While the primary driver for implementing the `bt-hci` traits on top of `esp-idf-svc` is to run `trouble` on top of it:
- Naming the bti-hci controller module `trouble` is incorrect. In theory there could be other BLE hosts that could use it
- Hiding this code behind a feature named `trouble` is also not ideal for the same reasons. Furthermore, hiding this code behind a feature in the first place seems unnecessary?

I suggest:
- Remove the `trouble` feature altogether. Whether the bt-hci support is compiled-in or compiled-out should **only** depend on the `esp_idf_*` flags
- Merge `bt_controller/trouble.rs` into `bt_controller.rs`. What is this separation supposed to model anyway?

src/bt_controller/trouble.rs
```diff
@@ -0,0 +1,335 @@
...
+    not(esp_idf_bt_le_hci_interface_use_ram)
+))]
+compile_error!("the `trouble` feature requires CONFIG_BT_LE_HCI_INTERFACE_USE_RAM=y");
+
+use core::cell::RefCell;
+use core::fmt;
+use core::sync::atomic::{AtomicBool, Ordering};
+
+use alloc::collections::VecDeque;
+use alloc::vec::Vec;
```
ivmarkov, line 15: This module (or ts parent `bt_controller` one after you merge this back into `bt_controller` should have `cfg(feature = "alloc")`.

src/bt_controller/trouble.rs
```diff
@@ -0,0 +1,335 @@
...
+    NullPacket,
+}
+
+struct RxState {
+    queue: VecDeque<RxMessage>,
+    error: Option<TransportError>,
+}
+
+static ACTIVE: AtomicBool = AtomicBool::new(false);
+static RX_STATE: BlockingMutex<IsrRawMutex, RefCell<RxState>> =
```
ivmarkov, line 45: `esp-idf-svc` has its own blocking mutex, and its own signals. Ideally you should use those here, or else this code can only be enabled when the `embassy` feature is enabled.

While I find it unlikely that somebody would use `trouble` on top of `esp-idf-svc` without enabling the `embassy` support, it is a bit cleaner if you don't require `embassy` here.

src/bt_controller/trouble.rs
```diff
@@ -0,0 +1,335 @@
...
+        Self::InvalidPacket(value)
+    }
+}
+
+/// A packet-oriented `bt-hci` transport over ESP-IDF's virtual HCI interface.
+///
+/// The transport owns the controller so callbacks cannot outlive the
+/// controller lifecycle. ESP-IDF exposes one global VHCI endpoint, hence only
+/// one value can be active at a time.
+pub struct EspVhciTransport<'d> {
```
ivmarkov, line 200: The `bt-hci` crate went from 0.3 to 0.9 in less than one year, which is quite a headache for the HALs that want to support these traits. The problem is that once `trouble` bumps its requirement to `bt-hci V0.X` - while the HAL in question is still on `bt-hci V0.(x - 1)` - there is an immediate pressure to release a new version of the HAL so that it could be used with the newer `trouble`. And also pressure for the HAL to support older `bt-hci` versions so that newer versions of the HAL could be used with older versions of `trouble`

Should `bt-hci` breaking version be released less often, the pain would've been much smaller, but alas, this is not the case, or at least not yet.

There was a discussion on this in the Embassy Matrix room ~ 1 month ago.

For esp-based HALs, there is an escape hatch though. Rather than depending on `bt-hci` directly, you could depend on this: https://github.com/embassy-rs/bt-hci/tree/main/bt-hci-driver (warning - the name might change to `bt-hci-transport` - look at the discussion in the Embassy Matrix room). Depending on _just_ the transport crate means esp-idf-svc would simultaneously support mutliuple bt-hci versions, and most importantly - would be forward-compatible with future bt-hci releases.

src/bt_controller.rs
```diff
@@ -0,0 +1,137 @@
...
+    fn mode() -> esp_bt_mode_t {
+        esp_bt_mode_t_ESP_BT_MODE_BTDM
+    }
+}
+
+/// An initialized and enabled ESP-IDF Bluetooth controller without a host.
+///
+/// This type consumes the HAL Bluetooth modem token. Dropping it disables and
+/// deinitializes the controller.
+pub struct EspBtController<'d, M>
```
ivmarkov, line 96: How about we name this type `BtControllerDriver`, so that it matches `BtDriver` (the Bluedroid host + controller) and `BleDriver` (the NimBLE host + controller)?

src/bt_controller.rs
```diff
@@ -0,0 +1,137 @@
...
+    M: BtMode,
+{
+    fn drop(&mut self) {
+        esp!(unsafe { esp_bt_controller_disable() }).unwrap();
+
+        esp!(unsafe { esp_bt_controller_deinit() }).unwrap();
+    }
+}
+
+#[cfg(feature = "trouble")]
```
ivmarkov, line 133: As per my previous comments, merge this module here, remove the `trouble` feature.

src/lib.rs
```diff
@@ -9,6 +9,7 @@
 //! - `std`: Enable the use of std. Enabled by default.
 //! - `experimental`: Enable the use of experimental features.
 //! - `embassy-time-driver`: Implement an embassy time driver.
+//! - `trouble`: Expose a `bt-hci` transport over ESP-IDF VHCI for use with Trouble.
```
ivmarkov, line 12: Ditto.

src/lib.rs
```diff
@@ -40,6 +41,8 @@ pub mod ble;
     feature = "alloc",
 ))]
 pub mod bt;
+#[cfg(all(not(any(esp32s2, esp32p4)), esp_idf_bt_enabled, feature = "trouble",))]
```
ivmarkov, line 44: add feature alloc, remove feature trouble, etc.

Cargo.toml
```diff
@@ -26,6 +26,14 @@ embassy-time-driver = ["dep:embassy-time-driver", "embassy-time-queue-utils"]
 alloc = ["esp-idf-hal/alloc", "embedded-svc/alloc", "uncased/alloc"]
 nightly = ["embedded-svc/nightly", "esp-idf-hal/nightly"]
 experimental = ["embedded-svc/experimental", "esp-idf-hal/experimental"]
+trouble = [
```
ivmarkov, line 29: Remove

Cargo.toml
```diff
@@ -57,17 +65,22 @@ embassy-time-queue-utils = { version = "0.3", optional = true }
 embassy-futures = "0.1.2"
 embedded-storage = { version = "0.3", optional = true }
 futures-io = { version = "0.3", optional = true }
+bt-hci = { version = "0.9", optional = true, default-features = false }
```
ivmarkov, line 68: Removing the `trouble` feature would mean bt-hci becomes non-optional. Oh well. See also my previous comment on bt-hci-drvier (= bt-hci-transport) as a (much better) alternative.

Cargo.toml
```diff
@@ -57,17 +65,22 @@ embassy-time-queue-utils = { version = "0.3", optional = true }
...
+embassy-sync = { version = "0.7", optional = true }
+embedded-io = { version = "0.7", optional = true }
```
ivmarkov, line 70: not necessary to list explicitly. comes via `embedded-svc` anyway

Cargo.toml
```diff
@@ -57,17 +65,22 @@ embassy-time-queue-utils = { version = "0.3", optional = true }
 embassy-futures = "0.1.2"
 embedded-storage = { version = "0.3", optional = true }
 futures-io = { version = "0.3", optional = true }
+bt-hci = { version = "0.9", optional = true, default-features = false }
+embassy-sync = { version = "0.7", optional = true }
```
ivmarkov, line 69: As per my earlier comment, let's try not to introduce this.

ivmarkov (changes requested):
Thank you for taking the time and effort!

I think exposing the controller _might_ have some small benefits indeed (like running trouble on top of esp-idf-svc) so I'm not opposed to it in general.

We just have to be careful how to do it so that we don't reintroduce a ton of dependencies that might be a bit of a headache to support in future, as the benefit exists but is not that big.

For one, the host in `trouble` is not certified, and it is unclear when or if Espressif would certify `trouble` as _the_ BLE host to be used for their `esp-hal` baremetal portfolio.

... and Espressif would almost **never**, ever certify `trouble` on top of the community-maintained `esp-idf-*` crates.

In fact, one major reason I'm keeping the `esp-idf-*` crates alive - in the presence of `esp-hal` - is that they run on top of ESP-IDF and as such inherit the certified status of the Wifi, Thread and BT stacks (both NimBLE host for which APIs were just introduced and Bluedroid host) of ESP-IDF. None of these three are certified in `esp-hal` yet, which is a bit of a roadblock for production (as opposed to hobby) use-cases. Running `trouble-host` on top of ESP-IDF would of course not be a certified setup.

